Issue #116, opened 2024-02-09
Buffer.from(str, 'base64') fails to decode valid Base64


dforsber:
I wanted to try out LLRT, but I failed to get base64 decoding to work.

```javascript
Buffer.from(str, 'base64');
```

Dgarc359:
Hey there, I've been browsing this project as I work with AWS daily, excited to see where it goes! Took some time to investigate this myself to see if I could reproduce the issue. The only time I was able to produce an error was when decoding a Buffer into `ascii`, but this seems like it's probably intended behavior.

@dforsber [Here's](https://github.com/Dgarc359/llrt-buffer-decoding-example) an example project that will get you started with a zip file that you can deploy to AWS. Here's the output you should expect when you test the lambda:

<img width="1267" alt="image" src="https://github.com/awslabs/llrt/assets/87497958/8eec38d3-315b-431b-bf6b-83d4e8165ae4">

Please note: you will need to provide your own llrt binary for security purposes

dforsber:
```shell
Welcome to Node.js v18.18.2.
Type ".help" for more information.
> Buffer.from('eyJhbGciOiJIUzUxMiIsInR5cCI6IkpXVCIsImtpZCI6IlRhcFRva2VuIn0', 'base64').toString()
'{"alg":"HS512","typ":"JWT","kid":"TapToken"}'
```

```shell
./llrt -e 'Buffer.from("eyJhbGciOiJIUzUxMiIsInR5cCI6IkpXVCIsImtpZCI6IlRhcFRva2VuIn0", "base64").toString()'
Error: Base64Error
    at <eval> (eval_script:1:8)
```

==> It seems I need to add `=` to the end.

```shell
 ./llrt -e 'console.log(Buffer.from("eyJhbGciOiJIUzUxMiIsInR5cCI6IkpXVCIsImtpZCI6IlRhcFRva2VuIn0=", "base64").toString())'
{"alg":"HS512","typ":"JWT","kid":"TapToken"}
```

richarddavison:
Ah right, thank you. This is a padding issue. Will fix 👌

PR #120, opened 2024-02-11
Fix Buffer.from base64 when padding is missing

richarddavison:
*Issue #, if available:*
Fixes: https://github.com/awslabs/llrt/issues/116

*Description of changes:*
Changes b64 decoding to forgiving mode that ignores missing padding.

### DIFF
--- a/tests/buffer.test.ts
+++ b/tests/buffer.test.ts
@@ -87,13 +87,7 @@ describe("Buffer.from", () => {
     const input2 = "SGVsbG8sIHdvcmxkIQ";
     const buffer2 = Buffer.from(input2, "base64");
     assert.strictEqual(buffer2.toString(), "Hello, world!");
-
-    const input3 = "SGVsbG8sIHdvcmxkIQ=";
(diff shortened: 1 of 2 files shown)


tests/buffer.test.ts
```diff
@@ -83,8 +83,18 @@ describe("Buffer.from", () => {
...
+    const input2 = "SGVsbG8sIHdvcmxkIQ";
+    const buffer2 = Buffer.from(input2, "base64");
+    assert.strictEqual(buffer2.toString(), "Hello, world!");
+
+    const input3 = "SGVsbG8sIHdvcmxkIQ=";
+    const buffer3 = Buffer.from(input3, "base64");
+    assert.strictEqual(buffer3.toString(), "Hello, world!");
   });
 
+  it("should create a buffer from a string with forgiving base64 encoding", () => {});
```
shyim, line 96: should this test be empty? 🤔

richarddavison: Nope! Thanks for spotting!

PR #155, opened 2024-02-14
Update license section in README

mertalev:
This section still refers to MIT-0, while it seems the current license is now Apache-2.0.


richarddavison (approved):

richarddavison:
Oh right, thanks for noticing!

Issue #150, opened 2024-02-13
process.hrtime.bigint() returns result inconsistent with node.js runtime


floydspace:
```js
// teat.js
console.log(process.hrtime.bigint());
```
<img width="476" alt="image" src="https://github.com/awslabs/llrt/assets/5180700/fb0c7c6a-118e-4480-bb5d-93573f51e518">

https://github.com/awslabs/llrt/blob/b01b7aab8a8182b05be57cfa297235250e58817f/src/process.rs#L40

richarddavison:
Resolution should be the same, right now is micros rather than nanos. So we'll need to fix it!

PR #156, opened 2024-02-14
Fix process hr-time-res

richarddavison:
*Issue #, if available:*
Fixes: https://github.com/awslabs/llrt/issues/150

*Description of changes:*
This will return higher resolution to the `process.hrtime.bigint()`. The "time basis" from node is slightly different where node uses an offset from a previous value (that seems to be inited from uptime) where as LLRT uses an start offset from startup time of the program. This change does not effect the behavior of timers as they are not based from unix epoch and should have nanosecond precision.

### DIFF
--- a/src/console.rs
+++ b/src/console.rs
@@ -132,6 +132,7 @@ fn stringify_primitive<'js>(
             Type::Undefined => result.push_str(COLOR_BLACK),
             Type::Int | Type::Float | Type::Bool => result.push_str(COLOR_YELLOW),
             Type::Symbol => result.push_str(COLOR_GREEN),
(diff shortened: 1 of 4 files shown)


src/console.rs
```diff
@@ -144,6 +145,12 @@ fn stringify_primitive<'js>(
         } else {
             "false"
         }),
+        Type::BigInt => {
+            let mut buffer = itoa::Buffer::new();
+            let big_int = value.as_big_int().unwrap();
+            result.push_str(buffer.format(big_int.clone().to_i64()?));
+            result.push('n');
```
floydspace, line 152: 👍

PR #152, opened 2024-02-13
add documentation on using aws sam

Dgarc359:
*Issue #, if available:*
#149 

*Description of changes:*
Simple documentation outlining how AWS SAM can be used. I also made a repository with a small POC example [HERE](https://github.com/Dgarc359/llrt-aws-sam-example)

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

Issue #149, opened 2024-02-13
AWS SAM example sought


kaihendry:
Is there some AWS::Serverless-2016-10-31 template.yaml & aws-actions/setup-sam@v2 example please?

https://github.com/awslabs/llrt?tab=readme-ov-file#configure-lambda-functions-to-use-llrt isn't straightforward enough

Dgarc359:
Hey there! I've pushed a PR with some simple instructions, but I also created a sample project here:
https://github.com/Dgarc359/llrt-aws-sam-example

It doesn't go over using actions, although you should be able to work off of having the ability to bundle LLRT's binary along with your JS

richarddavison:
Fixed in https://github.com/awslabs/llrt/pull/158

PR #158, opened 2024-02-14
AWS SAM -- Readme Documentation + Example

Dgarc359:
*Issue #, if available:*
#149
*Description of changes:*
Update readme with reference to new example sam project in example directory

Apologies that #152 got closed by accident. Mistake on my part.

My formatter also removed trailing spaces from the Readme. I can push another commit to clean that up if it's necessary.

Tagging @richarddavison for visibility


richarddavison (changes requested):
Thanks for the PR. Great start! Can we make this a bit less opinionated? For example, remove all the linting, jest/tests, eslint etc and focus on a minimum working example.
Even though it's pretty cool to always pull the latest version via the Makefile it would be a bit simpler to just point to an existing zip file as a layer. That way, we can also avoid building with docker, and don't require users to have Make installed.

Dgarc359:
> Thanks for the PR. Great start! Can we make this a bit less opinionated? For example, remove all the linting, jest/tests, eslint etc and focus on a minimum working example. Even though it's pretty cool to always pull the latest version via the Makefile it would be a bit simpler to just point to an existing zip file as a layer. That way, we can also avoid building with docker, and don't require users to have Make installed.

Latest change applies the above. Worth noting that the example no longer pulls the llrt zip during build and expects the user to place the zip file in the root of the `llrt` directory. I saw a similar pattern being applied to the example in the `infrastructure` directory. Thank you for the feedback :)

example/llrt-sam/.gitignore
```diff
@@ -0,0 +1,634 @@
+
```
richarddd, line 1: Can we also make this bare minimum?

example/llrt-sam/template.yaml
```diff
@@ -0,0 +1,53 @@
...
+Globals:
+  Function:
+    Timeout: 3
+
+Resources:
+
+  LlrtLayer:
+    Type: AWS::Serverless::LayerVersion
+    Properties:
+      ContentUri: ../../llrt-lambda-x64.zip
```
richarddd, line 18: ARM is actually faster, so we should default to that

richarddd: Also remember set `CompatibleArchitectures` to `arm64`

example/llrt-sam/template.yaml
```diff
@@ -0,0 +1,53 @@
...
+      CompatibleRuntimes:
+        - provided.al2023
+  HelloWorldFunction:
+    Type: AWS::Serverless::Function # More info about Function Resource: https://github.com/awslabs/serverless-application-model/blob/master/versions/2016-10-31.md#awsserverlessfunction
+    Properties:
+      CodeUri: hello-world/
+      Handler: app.lambdaHandler
+      Runtime: provided.al2023
+      Architectures:
+        - x86_64
```
richarddd, line 28: arm64

richarddd (changes requested):
Almost there! Just a couple of minor changes and it LGTM

example/llrt-sam/template.yaml
kaihendry: Wasn't any faster when I tried arm64 https://github.com/kaihendry/llrt-aws-sam-example/actions/runs/7907126599/job/21583462311#step:7:20 vs x64 https://github.com/kaihendry/llrt-aws-sam-example/actions/runs/7907082851/job/21583326692#step:7:20 thought it probably wasn't the best test

richarddavison: That probably due to caches, Lambda uses a tiered cache that gets a bit better after a while. Also ARM is lower cost.  It's not a major difference in performance but you should experience slightly faster cold starts at least (at lower cost) :)

richarddavison (approved):

example/llrt-sam/template.yaml
```diff
@@ -0,0 +1,55 @@
...
+        - arm64
+      Layers:
+        - !Ref LlrtLayer
+    Metadata: # Manage esbuild properties
+      BuildMethod: esbuild
+      BuildProperties:
+        External:
+          - '@aws-sdk'
+          - '@smithy'
+          - uuid
```
kaihendry, line 39: Does this mean these are already present in the layer / runtime?

richarddavison: Yes! See https://github.com/awslabs/llrt?tab=readme-ov-file#using-aws-sdk-v3-with-llrt

PR #167, opened 2024-02-16
Make `process` as module

### LINKED ISSUE #161: `process` as module
there is the [Nodejs built-in `process` module](https://nodejs.org/api/process.html) available to use, and some packages (like [mongodb](https://github.com/mongodb/node-mongodb-native)) prefer using it instead of the global module.
The module implementation is pretty simple it [re-exports the global module](https://github.com/nodejs/node/blob/v21.6.2/lib/process.js)

would it be possible to do the same in `llrt`?

floydspace:
closes #161

Hey @richarddavison , please review carefully, I'm not as good in rust as in js and ts.
But I covered implementation with tests and it looks good to me,
if you see any improvements to make, pls let me know.

### DIFF
--- a/src/process.rs
+++ b/src/process.rs
@@ -3,9 +3,16 @@
 use std::{collections::HashMap, env};
 
 use rquickjs::{
-    atom::PredefinedAtom, convert::Coerced, function::Constructor, object::Property, prelude::Func,
+    atom::PredefinedAtom,
+    convert::Coerced,
+    function::Constructor,
+    module::{Declarations, Exports, ModuleDef},
+    object::Property,
+    prelude::Func,
     Array, BigInt, Ctx, Function, IntoJs, Object, Result, Value,
 };
+
+use crate::module::export_default;
 
 use crate::{STARTED, VERSION};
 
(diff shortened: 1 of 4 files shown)


floydspace:
I'm a bit concerned about repetition of code, that I have to explicitly remap all functions from global process to module process, but maybe it is the only way in Rust, I'm open to suggestions .

src/process.rs
```diff
@@ -115,3 +122,49 @@ pub fn init(ctx: &Ctx<'_>) -> Result<()> {
...
+        declare.declare("default")?;
+        Ok(())
+    }
+
+    fn evaluate<'js>(ctx: &Ctx<'js>, exports: &mut Exports<'js>) -> Result<()> {
+        let globals = ctx.globals();
+        let process: Object = globals.get("process")?;
+
+        export_default(ctx, exports, |default| {
+            default.set("env", process.get::<&str, Value>("env"))?;
```
richarddavison, line 152: We can iterate through the enumarable keys of the process object and then set them accordingly. That way we can get rid of the repeat:
 export_default(ctx, exports, |default| {
 
      for key in process.keys() {
         ...
      }
 }

tests/process.test.ts
```diff
@@ -0,0 +1,47 @@
+import _process from "process";
+
+describe("process", () => {
+  it("should have a process env", () => {
+    assert.equal(_process.env.PWD, process.env.PWD);
```
richarddavison, line 5: Will this ENV exist for all platforms? Can we check for `_` or something else?

richarddavison (changes requested):

richarddavison:
Awesome first PR! You look like a true Rustacean  🦀

src/process.rs
floydspace: I wanted to do similar like it is done in `timers`.
However `timers` has all props as `Function` type
```
let function: Function = globals.get(func_name)?;
```
But for `process` it didn't work, bc all props are different types.
could you give me some snippet how to make it generic?

tests/process.test.ts
floydspace: I can do assertion simply by `process.env` memory reference, should be sufficient.

src/process.rs
floydspace: I found better solution

src/process.rs
```diff
@@ -115,3 +120,34 @@ pub fn init(ctx: &Ctx<'_>) -> Result<()> {
...
+
+        declare.declare("default")?;
+        Ok(())
+    }
+
+    fn evaluate<'js>(ctx: &Ctx<'js>, exports: &mut Exports<'js>) -> Result<()> {
+        let globals = ctx.globals();
+        let process: Object = globals.get("process")?;
+
+        exports.export("default", process)?;
```
richarddavison, line 149: This won't allow you to access the named exports from process tho:
`import {env} from "process"`

richarddavison (changes requested):

src/process.rs
floydspace: I addressed it, spot on.

richarddavison (approved):

Issue #168, opened 2024-02-17
'Response' is not defined?


laiso:
I noticed that the Response object, which is strangely defined globally in the source code[^1], cannot be referenced during the execution of llrt. I found this while writing a simple web server.

I can see this in the following way:

```bash
llrt -e 'new Request("http://localhost:3000/") && console.log("OK")'

OK
```

```bash
llrt -e 'new Response() && console.log("OK")'

ReferenceError: 'Response' is not defined
    at <eval> (eval_script:1:5)
```

[^1]: https://github.com/awslabs/llrt/blob/f0a4983d4fe45890f037741a66ee53707d5adcb5/src/http/mod.rs#L16-L28

Dgarc359:
I think I have identified the root cause of this issue. I will post a PR with a targetted fix. My approach will ensure that when the Response object is first created, users will receive this object here:
![image](https://github.com/awslabs/llrt/assets/87497958/4e68472f-d5d4-4cc8-b107-7851c9f926f0)

PR #169, opened 2024-02-17
add constructor to response struct 

Dgarc359:
*Issue #, if available:*
#168

*Description of changes:*
Adds a simple constructor for the Response struct. I was not able to find the reference class in nodejs github. Please let me know if I need to refactor the default values or function signature. Thanks!


richarddavison:
Thanks for the PR 🥇 ! However, it's a little more complex due to we need to capture body which required some more refactoring. I already have a branch where I started addressing this. PR coming early next week!

PR #170, opened 2024-02-17
Implement support import `fs` promises from root of fs module

floydspace:
To improve compatibility with many packages using fs promises, I implemented the compatibility with the following way of import
```ts
import {promises as fs} from 'fs';
```
it does not have any breaking changes, just refactored a bit the fs module
What do you think?


richarddavison:
This is a step in the right direction!
Callbackify existing fs/promises should also be possible. This would be simplest to do from js.

floydspace:
> ... This would be simplest to do from js.

yes, the `llrt` runtime consumer can workaround it on js side, however, it will lead to making a lot of patches for the third-party libraries, or making runtime-specific implementation, like it is usually done with `bun` or `deno`.
But if we could make `llrt` as compatible with `node` as possible, the enabling `llrt` usage could be much effortless.

richarddavison:
Sorry I was not clear. I mean it should definitely be added to LLRT just that it can be implemented in JS rather than rust (callbackify)

build.mjs
```diff
@@ -36,7 +36,7 @@ const ES_BUILD_OPTIONS = {
     "uuid",
     "hex",
     "os",
-    "fs/promises",
```
richarddavison, line 39: Don't we need to specify both `fs/promises` and `fs`? as they are separate modules

richarddavison (changes requested):

build.mjs
floydspace: `fs/promises` will be excluded by default as the child path of `fs`.
Trust me, it was my contribution to `esbuild`.
https://github.com/evanw/esbuild/pull/186

richarddavison (approved):

PR #171, opened 2024-02-18
Update Step Functions service endpoint from sfn to states

kylegrabfelder:
When importing the `SFNClient` from `@aws-sdk/client-sfn`, the runtime will throw an error: `client error (Connect)`.

This was occurring because the `addAwsSdkInitTask` function in the bootstrap was trying to connect to
`https://sfn.<region>.amazonaws.com` rather than `https://states.<region>.amazonaws.com`.

This change updates the service endpoint subdomain from `sfn` to `states`.


richarddavison (approved):

richarddavison:
Thanks for the PR 👏

PR #179, opened 2024-02-20
Crypto.randomBytes returns a Buffer

### LINKED ISSUE #177: Crypto.randomBytes returns Uint8Array instead of Buffer
Say we have the following code:

```js
// index.mjs
import { randomBytes } from "crypto";
console.log(randomBytes(10).toString("base64"));
console.log(randomBytes(10));
```

It does not return the same result (aside from randomness) as Node.js

```sh
node index.mjs
# base64 string
3HJqTvKlThpveA==
<Buffer ff 68 8c 61 7d 57 32 8b 00 36>


llrt index.mjs
# just an array
199,143,52,135,209,139,167,219,9,14
Uint8Array [ 169, 68, 55, 29, 169, 125, 37, 213, 114, 125 ]
```

It seems the returned type from randomBytes is different between llrt and [node](https://nodejs.org/api/crypto.html#cryptorandombytessize-callback).

tmokmss:
*Issue #, if available:*
closes #177 

*Description of changes:*
I just stumbled upon this issue so let me try fix this 🙏 

I'm still not sure how should I validate the entire build, but at least I confirmed `crypto.randomBytes` returns a buffer locally.

### DIFF
--- a/src/crypto/mod.rs
+++ b/src/crypto/mod.rs
@@ -15,6 +15,7 @@ use rquickjs::{
 };
 
 use crate::{
+    buffer::Buffer,
     encoding::encoder::{bytes_to_b64_string, bytes_to_hex_string},
     module::export_default,
     utils::{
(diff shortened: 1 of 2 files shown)


richarddavison (approved):

richarddavison:
@tmokmss great PR! Thank you!

PR #180, opened 2024-02-20
docs: use the correct rollup plugins

Kikobeats:
The current Rollup plugins suggested are old. I updated to reference correct modules 🙂


richarddavison (approved):

richarddavison:
Fantastic @Kikobeats! Thanks for the PR!

PR #188, opened 2024-02-21
keys and forEach method for Headers

### LINKED ISSUE #186: `Headers.forEach` function is missing
In LLRT, [`Headers.forEach`](https://developer.mozilla.org/en-US/docs/Web/API/Headers/forEach) is not implemented.

```
llrt  -e "console.log(new Headers().forEach)"  
undefined

node  -e "console.log(new Headers().forEach)"
[Function: forEach]
```

I think this is a final missing piece to run [Hono](https://hono.dev/) hello world on LLRT.

tmokmss:
*Issue #, if available:*
closes #186 

*Description of changes:*

I omitted [forEach's third argument `object`](https://developer.mozilla.org/en-US/docs/Web/API/Headers/forEach) for now because i'm not sure how to efficiently pass Headers object in each loop (without cloning btreemap), and also it's rarely used anyway.

I add [`keys` method](https://developer.mozilla.org/en-US/docs/Web/API/Headers/keys) as it was also missing.

### DIFF
--- a/src/http/headers.rs
+++ b/src/http/headers.rs
@@ -83,13 +83,9 @@ impl Headers {
         self.js_iterator(ctx)
     }
 
-    pub fn for_each<'js>(&self, ctx: Ctx<'js>, callback: Function<'js>) -> Result<()> {
+    pub fn for_each<'js>(&self, callback: Function<'js>) -> Result<()> {
         for header in self.headers.iter() {
-            callback.call((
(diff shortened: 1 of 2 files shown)


richarddavison (approved):

richarddavison:
Fantastic! Thanks for the PR! LGTM

PR #193, opened 2024-02-23
add a deployment option using AWS CDK

tmokmss:
*Issue #, if available:*

*Description of changes:*

Let me add a link to the llrt cdk construct to README 🙏  I believe this will help anyone who wants to try LLRT with minimum effort.

Also, I confirmed that Hono can successfully run on the latest LLRT (even with JSX SSR). [example code](https://github.com/tmokmss/cdk-lambda-llrt/tree/main/example)

Thank you for the great work! 

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,6 +38,20 @@ That's it 🎉
 The following [example project](example/llrt-sam/) sets up a lambda
 instrumented with a layer containing the llrt runtime.
 
+### Option 4: AWS CDK
+
+You can use [`cdk-lambda-llrt` construct library](https://github.com/tmokmss/cdk-lambda-llrt) to easily deploy LLRT Lambda functions with AWS CDK.
+
+```ts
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -38,6 +38,20 @@ That's it 🎉
 The following [example project](example/llrt-sam/) sets up a lambda
 instrumented with a layer containing the llrt runtime.
 
+### Option 4: AWS CDK
+
+You can use [`cdk-lambda-llrt` construct library](https://github.com/tmokmss/cdk-lambda-llrt) to easily deploy LLRT Lambda functions with AWS CDK.
```
richarddavison, line 43: Nit: Remove "easy", we try to avoid it because it's very contextual :)

richarddavison (changes requested):

README.md
tmokmss: Thanks, fixed!

richarddavison (approved):

PR #211, opened 2024-02-27
add e2e tests support with Jest/Mochai assertions support

fredbonin:
*Description of changes:*
This pull request adds support for running integration tests within the LLRT runtime and support for Jest/Mocha assertions. End-to-end (E2E) tests from the aws-sdk-js-v3 repository were also included for Amazon S3, DynamoDB, and Amazon Cognito to validate functionality and integration.


build.mjs
```diff
@@ -12,7 +12,21 @@ const SRC_DIR = path.join("src", "js");
 const TESTS_DIR = "tests";
 const OUT_DIR = "bundle";
 const SHIMS = new Map();
-const TEST_FILES = await fs.readdir(TESTS_DIR);
+const TEST_FILES = [];
+async function readAllTestFilesRecursively(directory) {
```
richarddavison, line 16: We can refactor this to a `readDirRecursive` function that takes in a directory and a predicate filter function as argument. We can also use `readDir` option `withFileTypes` that saves the additional `fs.stat` call.

richarddavison (changes requested):

build.mjs
```diff
@@ -12,7 +12,22 @@ const SRC_DIR = path.join("src", "js");
 const TESTS_DIR = "tests";
 const OUT_DIR = "bundle";
 const SHIMS = new Map();
-const TEST_FILES = await fs.readdir(TESTS_DIR);
+const TEST_FILES = [];
+async function readAllTestFilesRecursively(directory, filePredicate) {
```
richarddavison, line 16: NIT: `readFilesRecursive` since it can be used for other files as well :)

build.mjs
```diff
@@ -12,7 +12,22 @@ const SRC_DIR = path.join("src", "js");
...
+  // Used readdir because of https://github.com/Effect-TS/effect/issues/1801
+  for (const file of (await fs.readdir(directory, { withFileTypes: true }))) {
+    const absolute = path.join(directory, file.name);
+    if (file.isDirectory()) {
```
richarddavison, line 20: We could do something like this parallelize dirwalk:
```javascript
async function getFilesRecursive(dir, predicate = () => true) {
  const dirents = await fs.readdir(dir, { withFileTypes: true });
  const files = [];
  await Promise.all(
    dirents.map(async (dirent) => {
      const absolutePath = path.resolve(dir, dirent.name);
      if (dirent.isDirectory()) {
        files.push(...(await getFilesRecursive(absolutePath, predicate)));
      } else if (predicate(dirent, absolutePath)) {
        files.push(absolutePath);
      }
    }),
  );
  return files;
}
```

richarddavison (changes requested):

richarddavison (approved):

PR #223, opened 2024-03-01
feat: add recursive option for readdir

### LINKED ISSUE #197: fs.readdir recursive doesn't work
hello,

files = await fs.readdir(someDir,{recursive:true})

doesn't work. 
only the files in someDir are returned. 
it seems the options are ignored.

it would be nice if recrsion would work.

shulandmimi:
Issue fix #197

*Description of changes:*

add recursive option for readdir

```js
fs.readdir('example', { recursive: true })
    .then(res => {
        console.log(res);
    });

```

### DIFF
--- a/fixtures/fs/readdir/readdir.js
+++ b/fixtures/fs/readdir/readdir.js
@@ -1,0 +1,5 @@
+import fs from 'fs/promises';
+
+fs.readdir('./', { recursive: true }).then(res => {
+    console.log(res);
+});
--- a/fixtures/fs/readdir/readdir.js
+++ b/fixtures/fs/readdir/readdir.js
@@ -1,5 +1,3 @@
 import fs from 'fs/promises';
 
-fs.readdir('./', { recursive: true }).then(res => {
-    console.log(res);
-});
+fs.readdir('./', { recursive: true }).then(res => {});
--- a/fixtures/fs/readdir/recursive/readdir.js
+++ b/fixtures/fs/readdir/recursive/readdir.js
@@ -1,0 +1,5 @@
+import fs from 'fs/promises';
+
+fs.readdir('./', { recursive: true }).then(res => {
+    console.log(res);
+});
--- a/fixtures/fs/readdir/recursive/readdir.js
+++ b/fixtures/fs/readdir/recursive/readdir.js
@@ -1,5 +1,3 @@
 import fs from 'fs/promises';
 
-fs.readdir('./', { recursive: true }).then(res => {
-    console.log(res);
-});
+fs.readdir('./', { recursive: true }).then(res => {});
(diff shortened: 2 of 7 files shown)


src/fs/read_dir.rs
```diff
@@ -57,25 +64,57 @@ pub async fn read_dir<'js>(
...
+            }
+        }
+    };
     let mut items = Vec::with_capacity(64);
+    let mut dirs = VecDeque::from_iter([PathBuf::from(path)]);
 
-    while let Some(child) = dir.next_entry().await? {
-        if let Some(name) = child.path().file_name() {
-            let name = name.to_string_lossy().to_string();
+    while let Some(mut dir_buf) = dirs.pop_back() {
```
richarddavison, line 100: Nice we have a non-recursive traversal 👍 
I have similar functionality in here that could be changed to return an iterator instead and reused here.
https://github.com/awslabs/llrt/blob/37eb92b062f858a42567c1457ee199d977c51153/src/utils/io.rs#L45
Also, we might speed up things if we parallelize the walk of subdirectories by spawning tasks that joining the awaits that calls next entries.

fixtures/fs/readdir/readdir.js
```diff
@@ -0,0 +1,5 @@
+import fs from 'fs/promises';
+
+fs.readdir('./', { recursive: true }).then(res => {
+    console.log(res);
```
richarddavison, line 4: Do we need the console.log here?

fixtures/fs/readdir/recursive/readdir.js
```diff
@@ -0,0 +1,5 @@
+import fs from 'fs/promises';
+
+fs.readdir('./', { recursive: true }).then(res => {
+    console.log(res);
```
richarddavison, line 4: Same as above

src/fs/read_dir.rs
```diff
@@ -57,25 +64,57 @@ pub async fn read_dir<'js>(
     path: String,
     options: Opt<Object<'js>>,
 ) -> Result<ReadDir> {
+    let mut path = path;
     let with_file_types = options
```
richarddavison, line 68: When we now have two options. We can first check if we have options and then access the variables:
```rust
if let Some(opts) = opts.0{
 //check for withFileTypes & recursive option
}
```

richarddavison (changes requested):
Thanks for the PR! Looks great, some minor adjustments! :)

src/fs/read_dir.rs
shulandmimi: Thanks for your suggestion, I added `DirectoryWalker` to replace them, you can see if they meet your expectations

richarddavison (approved):

richarddavison (approved):

PR #235, opened 2024-03-05
feat: implement readdirSync

fredbonin:
### Description of changes

Implemented readdirSync (https://nodejs.org/api/fs.html#fsreaddirsyncpath-options)

### Checklist

- [X] Ran `make fix` to format and apply Clippy auto fixes
- [N/A] Updated documentation if needed ([API.md](API.md)/[README.md](README.md)/Other)

*By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.*

### DIFF
--- a/src/fs/read_dir.rs
+++ b/src/fs/read_dir.rs
@@ -11,8 +11,8 @@ use crate::{
     utils::io::DirectoryWalker,
 };
 
-#[rquickjs::class]
 #[derive(rquickjs::class::Trace)]
+#[rquickjs::class]
 pub struct Dirent {
     #[qjs(skip_trace)]
     metadata: Metadata,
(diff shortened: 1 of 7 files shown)


src/utils/io.rs
```diff
@@ -102,4 +121,23 @@ where
...
+        let dir = std::fs::read_dir(dir)?;
+
+        for entry in dir.flatten() {
+            let name = entry.file_name().to_string_lossy().to_string();
+            if !(self.filter)(&name) {
+                continue;
+            }
+            let entry_path = entry.path();
+            let metadata = entry.metadata();
+            match metadata {
```
richarddavison, line 135: Why match here? Can we use `metadata()?;`? Or is it so that we check the rest of the dirs if we have a failture? If that's the case, it should be reflected in async version as well and we can simply do `continue` instead of pushing to stack right?

richarddavison (changes requested):

src/utils/io.rs
fredbonin: This is exactly why, I wanted to continue with the rest of the dirs but with no metadata for one.

I can change the async version so that it has the same behaviour or I can return directly like in the async version.

Any preferences ?

richarddavison: How does node handle this? For example if it can't access a subdirectory, does it throw error or return all other? I suspect the sooner.

fredbonin: After looking at it more and this doc : https://man7.org/linux/man-pages/man3/readdir.3.html, I will fail fast instead of ignoring the metadata for one directory

richarddavison:
Also add fs to [API.md](https://github.com/awslabs/llrt/pull/API.md)?

src/fs/read_dir.rs
```diff
@@ -93,28 +143,31 @@ pub async fn read_dir<'js>(
             }
         }
     };
-    let mut items = Vec::with_capacity(64);
 
-    let mut directory_walker = DirectoryWalker::new(PathBuf::from(path), |_| true);
+    let mut directory_walker: DirectoryWalker<fn(&str) -> bool> =
```
richarddavison, line 147: Do we need explicit type here?

richarddavison (changes requested):

src/fs/read_dir.rs
fredbonin: I tried to type the closure instead but the compiler won't let me (I had to put the explicit type in there for it to compile)

richarddavison (approved):

PR #239, opened 2024-03-05
chore: infra example improvements

### LINKED ISSUE #238: Update example runtimes
In the CDK example showing how to use the LLRT layer, Amazon Linux 2 (`Runtime.PROVIDED_AL2`) is being used instead of Amazon Linux 2023 (`Runtime.PROVIDED_AL2023`), recommended in the `README`:

https://github.com/awslabs/llrt/blob/c2810e180eaec66829fd92cdbe0e4cd8f0c0169f/example/infrastructure/src/index.ts#L176

The list of compatible runtimes for the layer is also confusing:
https://github.com/awslabs/llrt/blob/c2810e180eaec66829fd92cdbe0e4cd8f0c0169f/example/infrastructure/src/index.ts#L157-L163

* It should include `PROVIDED_AL2023`
* The `NODEJS_*` runtimes should be removed, as they prevent LLRT from being used
* `PROVIDED_AL2` should also be removed, as I don't see a reason why it should be used over 2023

I can open a PR to fix it, but let me know I'm missing something here 👍

nmussy:
### Issue # (if available)

Fixes #238 

### Description of changes

* Update the Amazon Linux runtime from version 2 to 2023
* Add a working example to demonstrate how to use a NodeJS runtime and the LLRT layer with `AWS_LAMBDA_EXEC_WRAPPER`
* Updated comments and other copy/paste errors induced by updating the default runtime from Node 18 to 20
* Fix `aws-sdk` external module Node 16 example

### Checklist

- [ ] ~~Created unit tests in `tests/unit` and/or in Rust for my feature if needed~~
- [x] Ran `make fix` to format and apply Clippy auto fixes

*By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.*

### DIFF
--- a/example/infrastructure/src/index.ts
+++ b/example/infrastructure/src/index.ts
@@ -230,13 +230,16 @@ const main = async () => {
     }
   );
 
-  // Node 20, provided "aws-sdk"
+  // Node 16, provided "aws-sdk"
   const v2Function = new aws_lambda_nodejs.NodejsFunction(stack, "V2", {
     functionName: "example-v2",
-    entry: "../functions/src/v2.mjs",
+    entry: "../functions/src/v2.js",
     ...props,
+    runtime: aws_lambda.Runtime.NODEJS_16_X,
     bundling: {
       ...props.bundling,
+      format: aws_lambda_nodejs.OutputFormat.CJS,
+      banner: "",
       externalModules: ["aws-sdk"],
     },
   });
--- a/example/infrastructure/src/index.ts
+++ b/example/infrastructure/src/index.ts
@@ -238,8 +238,6 @@ const main = async () => {
     runtime: aws_lambda.Runtime.NODEJS_16_X,
     bundling: {
       ...props.bundling,
-      format: aws_lambda_nodejs.OutputFormat.CJS,
-      banner: "",
       externalModules: ["aws-sdk"],
     },
   });
--- a/example/infrastructure/src/index.ts
+++ b/example/infrastructure/src/index.ts
@@ -156,29 +156,56 @@ const main = async () => {
     code: aws_lambda.Code.fromAsset("../../llrt-lambda-arm64.zip"),
     compatibleRuntimes: [
       aws_lambda.Runtime.NODEJS_16_X,
-      aws_lambda.Runtime.NODEJS_20_X,
+      aws_lambda.Runtime.NODEJS_18_X,
       aws_lambda.Runtime.NODEJS_20_X,
       aws_lambda.Runtime.NODEJS_LATEST,
-      aws_lambda.Runtime.PROVIDED_AL2,
+      aws_lambda.Runtime.PROVIDED_AL2023,
     ],
     compatibleArchitectures: [aws_lambda.Architecture.ARM_64],
   });
 
-  // LLRT hello
-  const helloLlrtFunction = new aws_lambda.Function(
+  // Amazon Linux runtime LLRT hello
+  const helloLlrtAlRuntimeFunction = new aws_lambda.Function(
     stack,
-    "HelloLlrtFunction",
+    "HelloLlrtAlRuntimeFunction",
     {
-      functionName: "example-hello-llrt",
+      functionName: "example-hello-llrt-al-runtime",
       code: aws_lambda.Code.fromAsset(sourceDirs["hello.mjs"]),
       ...props,
       environment: {},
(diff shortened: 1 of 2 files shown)


example/infrastructure/src/index.ts
```diff
@@ -202,7 +230,7 @@ const main = async () => {
     }
   );
 
-  // Node 18, provided "aws-sdk"
+  // Node 20, provided "aws-sdk"
   const v2Function = new aws_lambda_nodejs.NodejsFunction(stack, "V2", {
```
nmussy, line 234: This handler is also currently broken. Downgrading it to using the Node 16 runtime is not sufficient, as the v2 `aws-sdk` doesn't play nice with ESM. We can either update the runtime version and import statement in `v2.mjs` to something like `'aws-sdk/lib/aws.js'`, or just remove this example

richarddavison: Thanks for reporting. We should avoid importing the entire sdk 'aws-sdk/lib/aws.js', is the demo broken some other way?

nmussy: I went with CJS and `require`, it's not ideal but it's working now

example/infrastructure/src/index.ts
```diff
@@ -320,7 +351,7 @@ const main = async () => {
       code: aws_lambda.Code.fromAsset("../functions/build"),
       handler: "index.handler",
       ...props,
-      runtime: aws_lambda.Runtime.PROVIDED_AL2,
+      runtime: aws_lambda.Runtime.PROVIDED_AL2023,
```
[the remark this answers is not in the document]
nmussy: Done: #241

nmussy:
Should be good to merge

example/infrastructure/src/index.ts
richarddavison: I'll take a look! It should work with ESM as well no?

nmussy: You're right, thanks. The issue was only with `import DynamoDB from "aws-sdk/clients/dynamodb.js";`, the Lambda can keep the ESM configuration. I've reverted that change 👍

example/infrastructure/src/index.ts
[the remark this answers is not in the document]
richarddavison: Fixed by https://github.com/awslabs/llrt/pull/243

example/infrastructure/src/index.ts
```diff
@@ -156,36 +156,64 @@ const main = async () => {
...
       ...props,
       environment: {},
-      runtime: aws_lambda.Runtime.PROVIDED_AL2,
+      runtime: aws_lambda.Runtime.PROVIDED_AL2023,
+      layers: [llrtLayer],
+    }
+  );
+
+  // Node runtime LLRT hello
+  const helloLlrtNodeRuntimeFunction = new aws_lambda.Function(
```
richarddavison, line 182: I don't think we need to deploy this option. Maybe document it in README instead :)

richarddavison (changes requested):

example/infrastructure/src/index.ts
nmussy: Yeah, I mentioned it in the issue:

> It doesn't seem that AWS_LAMBDA_EXEC_WRAPPER is well documented in the AWS docs, there's only a couple of mentions of that environment variable in examples. It might be a good idea to add it this project README, as well as shed some light on the advantage of using a Node runtime instead of AL (I assume different packages/libraries attached to the image?) I'll leave that up to you.

This example is not very visible, and doesn't explain why you would choose it over AL2023

richarddavison: There is little to no value. Maybe best option here would be to add a comment around `compatibleRuntimes` explaining that EXEC_WRAPPER can be used. Or simply remove Node from `compatibleRuntimes`. What do you think?

nmussy: I'm very much removing them from the list of compatible runtimes if there is no point for it. We would list additional runtimes that require an env variable, which silently fails if it's not set (Node is used, not LLRT), and all for no meaningful gain. It doesn't feel like good DX, showing something that is just technically feasible

richarddavison: Agree!

nmussy: Done 👍

richarddavison (approved):

Issue #248, opened 2024-03-07
AbortController


richarddavison:
https://developer.mozilla.org/en-US/docs/Web/API/AbortController

No-op implementation is in:
https://github.com/awslabs/llrt/blob/36f8030a2c8194c84c9db1c37455ea3cd103ef97/src/events.rs#L378

Also see https://github.com/awslabs/llrt/issues/182
```[tasklist]
### Tasks
- [x] Add signal and abort handling to Fetch API
```

MozarellaMan:
Hi, I'd like to try working on this - just to clarify, wouldn't this include implementing `AbortSignal`? Or could it just return the no-op AbortSignal?

```rust
#[rquickjs::methods]
impl<'js> AbortController<'js> {
    #[qjs(constructor)]
    fn new() -> Self {
        Self {
            signal: AbortSignal
        }
    }

    #[qjs(get)]
    fn signal(&self) -> AbortSignal {
        self.signal
    }

    fn abort(reason: Opt<Value<'js>>) {

    }
}
```
I started with this but not sure how useful it would be without methods like `abort(reason)` on `AbortSignal`

richarddavison:
Hi @MozarellaMan! Thanks for looking into this. That's right, we need to have the actual implementation. For instance if signal is passed to fetch we should abort the request if signal is set. But it would suffice to just have the AbortController and signal classes and I'll can sort out the rest :)

MozarellaMan:
Nice! I'll see how far I get with stubbing out stuff. One other question, I'm running into an error with building, specifically with [generate_compression_dictionary](https://github.com/awslabs/llrt/blob/8bcbfe4fe893262c280bcd46bd4820f0c1267925/build.rs#L218) in `build.rs`. I'm not exactly sure what this is doing to be honest. But I get this error which I assume is from `zstd`
```
 --- stderr
  !  Warning : nb of samples too low for proper processing ! 
  !  Please provide _one file per sample_. 
  !  Alternatively, split files into fixed-size blocks representative of samples, with -B# 
  Error 14 : nb of samples too low
```
I'm building this on M1 Mac so I'm guessing it might have to do with file system stuff? Is this build step necessary for development - could I comment it out for now?

richarddavison:
[...]

PR #271, opened 2024-03-14
Add AbortController implementation

MozarellaMan:
### Issue # (if available)

https://github.com/awslabs/llrt/issues/248

### Description of changes

Adds an implementation for `AbortController` with stubbed methods for `AbortSignal` which would have the actual event aborting implementation.  This implementation attempts to follow the specification found here https://dom.spec.whatwg.org/#ref-for-dom-abortcontroller-abortcontroller%E2%91%A0

### Checklist

### DIFF
--- a/src/events.rs
+++ b/src/events.rs
@@ -445,6 +445,12 @@ impl<'js> AbortSignal<'js> {
         self.aborted
     }
 
+    #[qjs(get)]
+    pub fn reason(&self) -> Option<Value<'js>> {
+        self.reason.clone()
+    }
+
+
     #[qjs(skip)]
     pub fn set_reason(&mut self, reason: Opt<Value<'js>>) -> Result<Option<Value<'js>>> {
         if let Some(new_reason) = reason.0 {
--- a/src/events.rs
+++ b/src/events.rs
@@ -3,7 +3,6 @@
 #![allow(clippy::mutable_key_type, clippy::for_kv_map)]
 
 use std::{
-    borrow::BorrowMut,
     sync::{Arc, RwLock},
     time::Duration,
 };
@@ -12,8 +11,8 @@ use rquickjs::{
     class::{JsClass, OwnedBorrow, Trace, Tracer},
     module::{Declarations, Exports, ModuleDef},
     prelude::{Func, Opt, Rest, This},
-    CatchResultExt, Class, Ctx, Function, IntoJs, Object, Result, String as JsString, Symbol,
-    Value,
+    CatchResultExt, Class, Ctx, Exception, Function, IntoJs, Object, Result, String as JsString,
+    Symbol, Value,
 };
 
 use tracing::trace;
@@ -408,6 +407,7 @@ impl<'js> AbortController<'js> {
     }
 
     pub fn abort(
+        ctx: Ctx<'js>,
         this: This<Class<'js, Self>>,
         reason: Opt<Value<'js>>,
     ) -> Result<Class<'js, Self>> {
@@ -415,8 +415,12 @@ impl<'js> AbortController<'js> {
         if reason.0.is_some() {
             this.0.borrow_mut().signal.borrow_mut().set_reason(reason)?;
         } else {
-            // TODO store DOMException as reason
-            this.0.borrow_mut().signal.borrow_mut().set_reason(reason)?;
+            let abort_exception = Exception::from_value("AbortError".into_js(&ctx)?)?; // TODO: AbortError DOMException
+            this.0
+                .borrow_mut()
+                .signal
+                .borrow_mut()
+                .set_reason(Opt(Some(abort_exception.into_value())))?;
         }
 
         Ok(this.0)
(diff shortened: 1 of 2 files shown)


src/events.rs
```diff
@@ -372,29 +375,95 @@ impl<'js> EventEmitter<'js> {
...
+            reason: None,
+        }
+    }
+
+    #[qjs(get)]
+    pub fn aborted(&self) -> bool {
+        self.aborted
+    }
+
+    pub fn set_reason(&mut self, reason: Opt<Value<'js>>) -> Result<Option<Value<'js>>> {
```
richarddavison, line 450: Use `#[qjs(skip)]` or move to another `impl`

richarddavison: Should this also set aborted to true?

src/events.rs
```diff
@@ -372,29 +375,95 @@ impl<'js> EventEmitter<'js> {
...
+        let new_reason = mem::replace(&mut self.reason, reason.0);
```
richarddavison, line 451: Is mem::replace same as `self.reason.replace()`?

richarddavison (changes requested):
Looking great so far, thank you for the contributions! Added some minor comments

src/events.rs
MozarellaMan: I thought so, but it looks like `Option::replace()` takes `T` not `Option<T>`, so I'd have to unwrap the `reason` parameter first. Then I guess set `self.reason` to None if `reason` is None?

src/events.rs
MozarellaMan: > Should this also set aborted to true?

I wasn't sure as the spec didn't mention the abort controller setting `aborted` to true. I assume maybe whatever observes that the `reason` has changed would then set `aborted` to true. I could set it to true for now till that is implemented

MozarellaMan: Also what does `qjs(skip)` do? I couldn't find any documentation for it 👀

richarddavison: `qjs(skip)` will not expose the function to JS user land.

src/events.rs
richarddavison: there is `Option::swap` also 👀

MozarellaMan: I couldn't find an `Option::swap` 🤔 It feels like it should exist though

MozarellaMan: The test I wrote is still failing on asserting the `reason` when `abort()` is called, but it looks like `aborted` is being set to true at least. Will look into it

src/events.rs
```diff
@@ -372,29 +372,112 @@ impl<'js> EventEmitter<'js> {
...
+        AbortSignal {
+            aborted: true,
+            reason: reason.0,
+        }
+    }
+
+    // TODO: Returns an AbortSignal that will automatically abort after a specified time.
+    #[qjs(static)]
+    pub fn timeout() -> AbortSignal<'js> {
+        todo!()
```
richarddavison, line 480: Here you can simply use `ctx.spawn_exit` to spawn a task and then set abort after the specified time. First create a new abort signal and convert it to a Value: `AbortSignal{...}.into_js()`. Then clone it and return one copy while pass the other copy to the async task which then sets the abort value after said time. Something like:
```rust
 pub fn timeout<'js>(ctx: Ctx<'js>, mills: size) -> Result<Class<'js,Self>> {
        let signal = AbortSignal {
            aborted: true,
            reason: reason.0,
        }
        
        let signal_instance =  Class::instance(ctx,signal);
        let signal_instance2 = signal_instance.clone()
        
        ctx.spawn_exit(async move {
           tokio::time::sleep(mills).await;
           let borrow = signal_instance.borrow();
           borrow.aborted = true;
           borrow.reason = "Timeout" //check spec
        })?;
        Ok(signal_instance2);
    }

**Review by richarddavison — CHANGES_REQUESTED** [review]

File: src/events.rs (hunk 3 above)
Comment (MozarellaMan) [reply src/events.rs]: Thanks! That's really neat 😮 One question about the spec - it mentions putting `DOMException` objects as the `reason` in different cases. I'm not sure if I can I just use the quickjs `Exception` struct here - it looks like `DOMException` is a specific [interface type](https://webidl.spec.whatwg.org/#idl-DOMException). I could include making that in this PR if necessary

Comment (richarddavison) [reply src/events.rs]: Mark it as a todo and we'll address it later :) DOMEx feels a bit out of scope for this PR

File: src/events.rs [hunk 4]
```diff
@@ -474,10 +481,25 @@ impl<'js> AbortSignal<'js> {
...
+        };
+
+        let signal_instance = Class::instance(ctx.clone(), signal)?;
+        let signal_instance2 = signal_instance.clone();
+
+        ctx.spawn_exit(async move {
+            tokio::time::sleep(Duration::from_millis(milliseconds)).await;
+            let borrow = signal_instance.borrow_mut();
+            borrow.set_aborted(true);
+            borrow.set_reason(Opt(None)); // TODO: TimeoutError DOMException
```
richarddavison, line 498: Use an an Error class

richarddavison (changes requested):

src/events.rs
MozarellaMan: Wasn't sure the best way to test this, doesn't seem practical with `jest`

richarddavison (approved):

PR #275, opened 2024-03-16
Fix 'make release' in Windows WSL2 environment

nabetti1720:
### Description of changes

- Fix `make release` in Windows WSL2 environment

### Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -32,6 +32,8 @@ endif
 
 ifeq ($(ARCH),aarch64)
 	ARCH = arm64
+else ifeq ($(ARCH),x86_64)
+	ARCH = x64
 endif
 
 CURRENT_TARGET ?= $(TARGET_$(DETECTED_OS)_$(ARCH))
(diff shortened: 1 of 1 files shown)


nabetti1720:
Hi @calavera, Thank you for your approval.
However, I guess I failed to take into account the CI impact of this modification.

Close the Pull Request once.

nabetti1720:
This was due to the conversion of `ARCH` in `MAKEFILE`.
It seems that in the current `MAKEFILE`, there are places that expect it to be `x86_64` and other places that expect it to be `x64`.
I don't know if this is the right way to do it, but I added a process in the form of an alias so that if the `ARCH` is `x86_64`, the `x64` artifacts are created.

Locally, the following commands all worked.
```bash
$ make release
$ make release-linux
$ make test-ci
```

I have just pushed the correction.
Is it possible to run workflow again in this state?

richarddd (approved):

PR #278, opened 2024-03-17
feat: implement statSync, writeFileSync

fredbonin:
### Description of changes

Implement writeFileSync and statSync

### Checklist

- [X] Updated documentation if needed ([API.md](API.md)/[README.md](README.md)/Other)

### DIFF
--- a/src/fs/write_file.rs
+++ b/src/fs/write_file.rs
@@ -22,3 +22,11 @@ pub async fn write_file<'js>(ctx: Ctx<'js>, path: String, data: Value<'js>) -> R
 
     Ok(())
 }
+
+pub fn write_file_sync<'js>(ctx: Ctx<'js>, path: String, data: Value<'js>) -> Result<()> {
+    let bytes = get_bytes(&ctx, data)?;
+
+    std::fs::write(&path, bytes).or_throw_msg(&ctx, &format!("Can't write \"{}\"", &path))?;
+
+    Ok(())
+}
--- a/API.md
+++ b/API.md
@@ -62,6 +62,8 @@ Available globally
 [readFileSync](https://nodejs.org/api/fs.html#fsreadfilesyncpath-options)
 [rmdirSync](https://nodejs.org/api/fs.html#fsrmdirsyncpath-options)
 [rmSync](https://nodejs.org/api/fs.html#fsrmsyncpath-options)
+[statSync](https://nodejs.org/api/fs.html#fsstatsyncpath-options)
+[writeFileSync](https://nodejs.org/api/fs.html#fswritefilesyncfile-data-options)
 
 ## fs/promises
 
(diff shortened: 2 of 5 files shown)


src/fs/write_file.rs
```diff
@@ -22,3 +22,11 @@ pub async fn write_file<'js>(ctx: Ctx<'js>, path: String, data: Value<'js>) -> R
 
     Ok(())
 }
+
+pub fn write_file_sync<'js>(ctx: Ctx<'js>, path: String, data: Value<'js>) -> Result<()> {
+    let bytes = get_bytes(&ctx, data)?;
+
+    std::fs::write(&path, bytes).or_throw_msg(&ctx, &format!("Can't write \"{}\"", &path))?;
```
richarddavison, line 29: Do we need to `flush`? I saw that CI sometimes failed when not flushing

richarddavison (changes requested):

src/fs/write_file.rs
fredbonin: Seems not necessary with this function : 

```
/// Write a slice as the entire contents of a file.
///
/// This function will create a file if it does not exist,
/// and will entirely replace its contents if it does.
///
/// Depending on the platform, this function may fail if the
/// full directory path does not exist.
///
/// This is a convenience function for using [`File::create`] and [`write_all`]
/// with fewer imports.
```

and write_all

```
    /// Attempts to write an entire buffer into this writer.
    ///
    /// This method will continuously call [`write`] until there is no more data
    /// to be written or an error of non-[`ErrorKind::Interrupted`] kind is
    /// returned. This method will not return until the entire buffer has been
    /// successfully written or such an error occurs. The first error that is
    /// not of [`ErrorKind::Interrupted`] kind generated from this method will be
    /// returned.
    ///
    /// If the buffer contains no data, this will never call [`write`].
    ///
    /// # Errors
    ///
    /// This function will return the first error of
    /// non-[`ErrorKind::Interrupted`] kind that [`write`] returns.
    ///
    /// [`write`]: Write::write
    ///
    /// # Examples
    ///
    /// ```no_run
    /// use std::io::prelude::*;
    /// use std::fs::File;
    ///
    /// fn main() -> std::io::Result<()> {
    ///     let mut buffer = File::create("foo.txt")?;
    ///
    ///     buffer.write_all(b"some bytes")?;
    ///     Ok(())
    /// }
    /// ```
    #[stable(feature = "rust1", since = "1.0.0")]
    fn write_all(&mut self, mut buf: &[u8]) -> Result<()> {
```

richarddavison: I might not be needed for sync, but CI failed (sometimes) when not doing so for async (which is just a wrapper around sync call). Leets keep it like this and keep an eye out for failed tests

PR #287, opened 2024-03-21
chore: use expect(..) instead of assert(..)

fredbonin:
### Description of changes

Used expect instead of assert for all unit tests

### Checklist

- [N/A] Created unit tests in `tests/unit` and/or in Rust for my feature if needed
- [N/A] Updated documentation if needed ([API.md](API.md)/[README.md](README.md)/Other)

### DIFF
--- a/tests/unit/buffer.test.ts
+++ b/tests/unit/buffer.test.ts
@@ -3,10 +3,10 @@ describe("Buffer.alloc", () => {
     const size = 10;
     const buffer = Buffer.alloc(size);
 
-    assert.strictEqual(buffer.length, size);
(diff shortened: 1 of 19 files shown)


richarddavison (changes requested):
Fantastic! Can we now remove deps on `assert` and update docs? Readme references `assert` module

fredbonin:
> Fantastic! Can we now remove deps on `assert` and update docs? Readme references `assert` module

Definitely, but with all the changes I think it would be better to do it in a second PR. What do you think ?

richarddavison (approved):

PR #290, opened 2024-03-22
feat: implement globalThis.navigator.userAgent

nabetti1720:
### Description of changes

I implemented one of the APIs, `globalThis.navigator.userAgent`.

This is my first time coding in the rust language, so I will promote it as a draft version.
Please point out if it does not follow `llrt` or `rust` conventions.

### Local execution results

image

### Checklist

### DIFF
--- a/src/http/fetch.rs
+++ b/src/http/fetch.rs
@@ -160,7 +160,7 @@ pub(crate) fn init(ctx: &Ctx<'_>, globals: &Object) -> Result<()> {
                 let mut req = Request::builder()
                     .method(method)
                     .uri(uri)
-                    .header("user-agent", format!("llrt {}", VERSION))
+                    .header("user-agent", format!("llrt/{}", VERSION))
                     .header("accept", "*/*");
 
                 if let Some(headers) = headers {
--- a/src/http/fetch.rs
+++ b/src/http/fetch.rs
@@ -160,7 +160,7 @@ pub(crate) fn init(ctx: &Ctx<'_>, globals: &Object) -> Result<()> {
                 let mut req = Request::builder()
                     .method(method)
                     .uri(uri)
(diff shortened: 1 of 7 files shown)


src/http/fetch.rs
```diff
@@ -160,7 +160,7 @@ pub(crate) fn init(ctx: &Ctx<'_>, globals: &Object) -> Result<()> {
                 let mut req = Request::builder()
                     .method(method)
                     .uri(uri)
-                    .header("user-agent", format!("llrt {}", VERSION))
+                    .header("user-agent", format!("llrt/{}", VERSION))
```
richarddavison, line 163: nice finding!

tests/unit/navigator.test.ts
```diff
@@ -0,0 +1,9 @@
+import defaultImport from "navigator";
+import * as namedImport from "navigator";
+
+describe("navigator.userAgent", () => {
+  it("should have a navigator userAgent", () => {
+    expect(defaultImport.userAgent).toEqual(navigator.userAgent);
```
richarddavison, line 6: should we maybe do a regex check here or starts with "llrt/"?

richarddavison (changes requested):
Looks great! Minor test comment

tests/unit/navigator.test.ts
nabetti1720: Thanks for pointing that out.
According to the specification, it is not required to start with a slash, but we will follow the implementation and make sure that it starts with "llrt/".

richarddavison: Ok if that's not the case they we should revert to `llrt {version}` because this is already used for stats :)

nabetti1720: OK, I understand.
I'll try to revert to `llrt {version}`.

richarddavison (approved):

PR #306, opened 2024-03-26
feat: implement randomInt in crypto module

fredbonin:
### Description of changes

Implement randomInt with support for min and max value only so that it can be used with powertools.

### Checklist

- [X] Updated documentation if needed ([API.md](API.md)/[README.md](README.md)/Other)

### DIFF
--- a/tests/unit/crypto.test.ts
+++ b/tests/unit/crypto.test.ts
@@ -51,7 +51,7 @@ describe("random", () => {
     crypto.randomFillSync(uint8Array);
     expect(uint8Array.length).toEqual(16);
     for (const byte of uint8Array) {
-      expect(byte >= 0 && byte <= 255).toBeTruthy()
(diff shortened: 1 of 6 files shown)


tests/unit/crypto.test.ts
```diff
@@ -1,4 +1,5 @@
 import crypto from "crypto";
+import nodeCrypto from 'node:crypto';
```
richarddd, line 2: Do we need a node prefix import? I recond that If you want to test prefix imports that should be another test :)

richarddd (changes requested):

tests/unit/crypto.test.ts
fredbonin: no you are right, it is generic and tested as part of console. I will remove this.

richarddavison (approved):

Issue #182, opened 2024-02-20
`signal` property missing from `Request` object


jabrks:
Forgive me as I was unsure if `Request` was covered by `fetch` in the compatibility matrix or not, but I noticed when creating a new `Request` object with an `AbortSignal` that the same `AbortSignal` is not then accessible from the `signal` property of the newly-created object

```js
// index.js
const controller = new AbortController();
const req = new Request('http://localhost/', { signal: controller.signal });

console.log(req.signal);
```

```bash
node index.js
# AbortSignal { aborted: false }
```

```bash
./llrt index.js
# undefined
```

richarddavison:
We've been working on increasing the WHATWG/WinterCG Fetch compatibility. What's still missing is AbortController and Readable/Writable streams. The AbortController support is relatively straight forward to implement as it's a pretty light API.

`AbortSignal` is currently only stubbed in LLRT and has no effect. This is something we're working on but PR's are welcome:
https://github.com/awslabs/llrt/blob/1ece463450aaab989a4ecd612caf34606e1c86ef/src/events.rs#L378

jabrks:
I'd completely missed that `AbortController` was stubbed. Thanks for confirming that this is on the roadmap

PR #312, opened 2024-03-30
Implement Request#signal.

calavera:
### Issue # (if available)

Fixes https://github.com/awslabs/llrt/issues/182

### Description of changes

Allow to pass an AbortSignal as an option to a request. Ensure that only AbortSignal types are allowed.

### Checklist

### DIFF
--- a/src/events.rs
+++ b/src/events.rs
@@ -9,10 +9,11 @@ use std::{
 
 use rquickjs::{
     class::{JsClass, OwnedBorrow, Trace, Tracer},
+    function::OnceFn,
     module::{Declarations, Exports, ModuleDef},
     prelude::{Func, Opt, Rest, This},
-    CatchResultExt, Class, Ctx, Exception, Function, IntoJs, Object, Result, String as JsString,
-    Symbol, Undefined, Value,
+    Array, CatchResultExt, Class, Ctx, Exception, Function, IntoJs, Object, Result,
+    String as JsString, Symbol, Undefined, Value,
 };
 
 use tracing::trace;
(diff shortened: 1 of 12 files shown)


src/events.rs
```diff
@@ -428,6 +428,7 @@ impl<'js> AbortController<'js> {
 }
 
 //TODO implement static methods abort() and timeout(miliseconds)
+#[derive(Clone, Debug, Default)]
```
richarddavison, line 431: nice, missed this one

src/http/request.rs
```diff
@@ -94,11 +104,21 @@ impl<'js> Request<'js> {
             None
         };
 
+        let signal = if let Some(signal) = &self.signal {
```
richarddavison, line 107: `.map()` instead of if let

src/http/request.rs
```diff
@@ -111,13 +131,35 @@ fn assign_request<'js>(request: &mut Request<'js>, ctx: Ctx<'js>, obj: &Object<'
         request.method = method;
     }
 
+    if obj.contains_key("signal").unwrap() {
```
richarddavison, line 134: We can simplify this code to something like
```rust
if let Some(signal) = obj.get::<_,Object>("signal"){
     let signal = AbortSignal::from_js(&ctx, signal.into_value())?;
     request.signal  = Some(Class::instance(ctx.clone(), signal)?);
}
```

src/http/request.rs
```diff
@@ -94,11 +104,21 @@ impl<'js> Request<'js> {
             None
         };
 
+        let signal = if let Some(signal) = &self.signal {
+            Some(Class::<AbortSignal>::instance(
+                ctx.clone(),
+                signal.borrow().clone(),
```
richarddavison, line 110: We should not clone signal value but rather clone the signal reference no? If we abort the first request the cloned request should be aborted as well.
`signal.clone()`

richarddavison (changes requested):
Thanks! Looks good so far. We also need to check if signal is aborted when we use the request. Similarly the same signal should be passed to response and handle aborted value as well

richarddavison:
> > We also need to check if signal is aborted when we use the request. Similarly the same signal should be passed to response and handle aborted value as well
> 
> @richarddavison can you point me to where that's handled? I cannot find the right place to add those checks.

First place would be to here:
https://github.com/awslabs/llrt/blob/a3a9fa73700006c46ebe3171635ffb48a70d440c/src/http/fetch.rs#L175

And then when response is read:
https://github.com/awslabs/llrt/blob/a3a9fa73700006c46ebe3171635ffb48a70d440c/src/http/response.rs#L133

I see two solutions:
1. Store a broadcast channel and broadcast on signal so async tasks can be cancelled
2. Periodically check for cancellations between async calls (may load unnecessary data)

richarddavison (approved):

PR #313, opened 2024-03-30
feat: implement globalThis.performance API

nabetti1720:
### Description of changes

I have implemented the following several APIs.
- globalThis.performance.timeOrigin
- globalThis.performance.now()

### Checklist

### DIFF
--- a/src/performance.rs
+++ b/src/performance.rs
@@ -10,11 +10,7 @@ use crate::module::export_default;
 
 use crate::STARTED;
 
-use once_cell::sync::Lazy;
-
-use chrono::Utc;
-
-static TIME_ORIGIN: Lazy<f64> = Lazy::new(|| (Utc::now().timestamp_micros() as f64) / 1e3);
+use crate::vm::TIME_ORIGIN;
 
 fn get_time_origin() -> f64 {
     *TIME_ORIGIN
--- a/src/performance.rs
+++ b/src/performance.rs
@@ -1,0 +1,69 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+use rquickjs::{
+    module::{Declarations, Exports, ModuleDef},
+    prelude::Func,
+    Ctx, Object, Result, Value,
+};
+
+use crate::module::export_default;
+
+use crate::STARTED;
+
+use once_cell::sync::Lazy;
+
+use chrono::Utc;
+
+static TIME_ORIGIN: Lazy<f64> = Lazy::new(|| (Utc::now().timestamp_micros() as f64) / 1e3);
+
+fn get_time_origin() -> f64 {
+    *TIME_ORIGIN
+}
+
+fn now() -> f64 {
+    let started = unsafe { STARTED.assume_init() };
+    let elapsed = started.elapsed();
+
+    elapsed.as_secs_f64() + (elapsed.as_micros() as f64) / 1e3
+}
+
+pub fn init(ctx: &Ctx<'_>) -> Result<()> {
+    let globals = ctx.globals();
+
+    let performance = Object::new(ctx.clone())?;
+
+    performance.set("timeOrigin", get_time_origin())?;
+    performance.set("now", Func::from(now))?;
+
+    globals.set("performance", performance)?;
+
+    Ok(())
+}
+
+pub struct PerformanceModule;
+
+impl ModuleDef for PerformanceModule {
+    fn declare(declare: &mut Declarations) -> Result<()> {
+        declare.declare("timeOrigin")?;
+        declare.declare("now")?;
+        declare.declare("default")?;
+        Ok(())
+    }
+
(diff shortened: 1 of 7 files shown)


src/performance.rs
```diff
@@ -0,0 +1,69 @@
...
+
+use crate::module::export_default;
+
+use crate::STARTED;
+
+use once_cell::sync::Lazy;
+
+use chrono::Utc;
+
+static TIME_ORIGIN: Lazy<f64> = Lazy::new(|| (Utc::now().timestamp_micros() as f64) / 1e3);
```
richarddavison, line 17: Can we reuse the same global variable we init in https://github.com/awslabs/llrt/blob/589900a7430cd496e6a96857ed91a598d6495de1/src/main.rs#L72C16-L72C23 and use in https://github.com/awslabs/llrt/blob/589900a7430cd496e6a96857ed91a598d6495de1/src/process.rs#L45

Maybe we should move this variable to `vm.rs ` as well

richarddavison (changes requested):

richarddavison:
Thanks for the PR! Just a minor "enhancement" and we're good to go

src/performance.rs
nabetti1720: Thanks for the quick review. Let me ask two questions.

1. Are you saying that the TIME_ORIGIN should be generated at the time of vm startup, not at the time of performance object's construct? Because it is a more precise time and the variables are reusable.
2. Can I assume that "this variable" is just TIME_ORIGIN? It is my understanding that STARTD is not a moving target.

nabetti1720: Please review again so we can correct the above assuming the above is correct.

nabetti1720: Sorry. I have reconsidered the above question.

I thought that if `TIME_ORIGIN` is initialized in `vm.rs`, `STARTED` should be moved in the same way. Because if the date and time are calculated by adding the numbers in `TIME_ORIGIN` and `STARTD`, it would be more accurate to initialize them at the same time.

We also thought that moving `STARTED: MaybeUninit<Instant>` and `TIME_ORIGIN: MaybeUninit<f64>` to `vm.rs` would make it more clear that they belong in `vm`.

richarddavison: Yes! `Maybe` we should have an atomic bool with a look if false to check if we've already inited `STARTED` and TIME_ORIGIN. I wan't to make instantiating VM thread safe. Now `STARTED` and `TIME_ORIGIN` would cause program to panic if a second instance of VM is created.

On another note, can we remove `STARTED` in favor of `TIME_ORIGIN ` or vice versa? They seem to have the same use case

nabetti1720: > Yes! `Maybe` we should have an atomic bool with a look if false to check if we've already inited `STARTED` and TIME_ORIGIN. I wan't to make instantiating VM thread safe. Now `STARTED` and `TIME_ORIGIN` would cause program to panic if a second instance of VM is created.

I am not familiar with AtomicBool. Is it ok to use the following implementation?

vm.rs:
``` rust
static INITIALIZED_TIME_ORIGIN: AtomicBool = AtomicBool::new(false);

impl Vm {
    ...
    pub async fn new() -> StdResult<Self, Box<dyn std::error::Error + Send + Sync>> {
        if INITIALIZED_TIME_ORIGIN
            .compare_exchange(false, true, Ordering::Acquire, Ordering::Relaxed)
            .is_ok()
        {
            // initialization instruction
        }
        ...
```
I feel that this implementation cannot handle the case where another thread references `performance.timeOrogin` or `performance.now()` between the if statement decision and the initialization process. Am I thinking too much?

> On another note, can we remove `STARTED` in favor of `TIME_ORIGIN ` or vice versa? They seem to have the same use case

The reasons why we thought it necessary to manage `STARTED` and `TIME_ORIGIN` are as follows.

- `STARTED`, i.e., `Instant::now()`, is a monotonic clock, but does not preserve the runtime date and time.

- I understand that `TIME_ORIGIN`, i.e., `Utc::now()`, can obtain a timestamp at runtime, but it is not a monotonic clock and is affected by OS-level time changes such as `NTP`.

If `TIME_ORIGIN` is a monotonic clock or the runtime can be obtained from `STARTED`, it may be possible to unify the two variables.

richarddavison: STARED is just used for HR time so can be replaced with TIME_ORIGIN.

I think we can simplify here and store nanoseconds since unix epoch in a AtomicUsize:
```rust
static TIME_ORIGIN: AtomicBool = AtomicUsize::new(0);

impl Vm {
    ...
    pub async fn new() -> StdResult<Self, Box<dyn std::error::Error + Send + Sync>> {
        if TIME_ORIGIN.load(Ordering::Relaxed) == 0{
              let time_origin = Utc::now().timestamp_nanos_opt() as size;
              TIME_ORIGIN.store(time_origin, Ordering::Relaxed)
        }
}
```

nabetti1720: Thanks for the advice.
We have made improvements, please review.

richarddavison (approved):

Issue #320, opened 2024-04-01
URL to_string bug


gc-victor:
In the URL to_string method, there is a bug in line 115:

https://github.com/awslabs/llrt/blob/808084b8cbef50218be3a17b110e8c438490e013/src/http/url.rs#L115

Should be:

`&self.protocol, user_info, &self.hostname, port, &self.pathname, &search, &hash` 

Should be changed `&self.host`  by `&self.hostname`.

Reference:
https://developer.mozilla.org/en-US/docs/Web/API/URL/host
https://developer.mozilla.org/en-US/docs/Web/API/URL/hostname

gc-victor:
You are right!

I have checked it again and I think the best option should be:

`&self.protocol, user_info, &self.host, &self.pathname, &search, &hash`

Either option will work, but using "host", the "port" part of the function adding the ":" can be removed.

Example:

```javascript
const url = new URL("https://developer.mozilla.org:443/en-US/docs/Web/API/URL/host");
console.log(url.toString()) // https://developer.mozilla.org/en-US/docs/Web/API/URL/host
console.log(url.host) // developer.mozilla.org
console.log(url.port) // empty-string
```

References:
https://developer.mozilla.org/en-US/docs/Web/API/URL/toString
https://url.spec.whatwg.org/#url-serializing
https://url.spec.whatwg.org/#url-miscellaneous

PR #321, opened 2024-04-02
Fix URL port/port when default protocols

richarddavison:
### Issue # (if available)

Fixes https://github.com/awslabs/llrt/issues/320

### Description of changes

Port should be removed if default for the protocol

### Checklist

### DIFF
--- a/src/http/url.rs
+++ b/src/http/url.rs
@@ -13,6 +13,9 @@ use crate::utils::result::ResultExt;
 
 use super::url_search_params::URLSearchParams;
 
+static DEFAULT_PORTS: &[&str] = &["21", "80", "443"];
+static DEFAULT_PROTOCOLS: &[&str] = &["ftp", "http", "https"];
+
 #[derive(Clone)]
 #[rquickjs::class]
 pub struct URL<'js> {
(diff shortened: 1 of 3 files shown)


src/http/url.rs
```diff
@@ -13,6 +13,9 @@ use crate::utils::result::ResultExt;
 
 use super::url_search_params::URLSearchParams;
 
+static DEFAULT_PORTS: &[&str] = &["21", "80", "443"];
+static DEFAULT_PROTOCOLS: &[&str] = &["ftp", "http", "https"];
```
gc-victor, line 17: Using the url crate and parsing the URLs shouldn't be required. :thinking: 

Reference:
https://github.com/servo/rust-url/blob/73803fa780c8890e0b40cd01fbbd5362fbe4c9e0/url/tests/setters_tests.json#L764

richarddd: > Using the url crate and parsing the URLs shouldn't be required. 🤔

Can you expand? I don't follow

richarddavison: Oh I see what you mean. I still think it's needed tho, because it should be reflected when setting `port`, `protocol` or `host`. I can work around this by parsing a fake url and extract the port from that, but it seems rather inefficient. But to your point, `filtered_port` can likely be removed from the `new` method then.

gc-victor: The url crate has setters that can be used for those cases.

Reference:
https://docs.rs/url/2.5.0/url/struct.Url.html?search=set_

richarddavison: That's right, but it would require us to store the url instead of the values! However this is not a bad thing as we can refactor the URL module to instead hold a `url` value and just call methods of that instead of extracting the values from a single parse. Would you be interested in taking this on? PR Welcome 👍

gc-victor: I will work on it as soon as I have some time.

PR #326, opened 2024-04-03
Making LLRT available as a crate

### LINKED ISSUE #319: Publishing LLRT as a crate
Hey all!

I'm working on adding a JavaScript runtime scripting to my Rust app. I started with Deno because it's broken down into crates, making it easy to use only what I need. This setup works well for me.

I'm looking at AWS's LLRT now, thinking it might give my app a performance boost. But it looks like LLRT isn't packaged in a way that lets me easily drop it into my Rust project.

I'm wondering if there are plans to make LLRT more modular, like Deno, for easier use in Rust apps. Also, I'm open to contributing to this effort when I have time. Is this something you'd be interested in?

Thanks for any info or suggestions you can provide!

KaanMol:
### Issue #319 (if available)
closes #319 

### Description of changes
Currently it is pretty "ok", some naming etc is made up on the fly, like the package names or like module_builder.

Splitting AWS parts out to its own package is possibly a better option to pick up in a second PR, since this PR is getting a bit on the big side. 

#### Tasks that might be picked up with this PR:
- Moving AWS specific parts out of the core of LLRT;

### Checklist


KaanMol:
@richarddavison, could you take a look and let me know if it aligns with what you had in mind? It is more of a proof of concept. I dont like how I did the module declaration and also the bind to global part, however at the same time no other solution comes to my mind.

src/module_builder.rs
```diff
@@ -0,0 +1,129 @@
...
+            builtin_resolver: ModuleResolver::default(),
+            module_loader: ModuleLoader::default(),
+            module_names: HashSet::new(),
+            init_global: Vec::new(),
+        }
+    }
+
+    pub fn with_default() -> Self {
+        Self::new()
+            .with_module("crypto", CryptoModule)
```
richarddavison, line 71: I like this approach! Is there a way to avoid hardcoding names of modules when using the builder pattern? Maybe an enum? Then we we can may implement `Into<ModuleInfo>` for the enum and use that as arguments for withModule?

richarddavison (changes requested):
I like this approach! Now third party projects can pull in a lightweight runtime and use the modules they wan't. Unused modules would then be automatically removed at compile time. Just some minor changes regarding module builder

src/module_builder.rs
KaanMol: I thought of something like that, but how would people add their own modules without changing the internal enum?
I actually wanted to add a name field in the module definitions, but it wasn't object safe so gave up on that.

richarddavison: They would just implement` Into<ModuleInfo>` for their structs or enum or whatever wrapper they use

KaanMol:
@richarddavison I had another problem, how would we allow passing values from LLRT to Rust context? Do you maybe have an idea? Deno has a `Scope` for example

richarddavison:
> @richarddavison I had another problem, how would we allow passing values from LLRT to Rust context? Do you maybe have an idea? Deno has a `Scope` for example

Can you expand a bit what you mean pass values? I'm not familiar with `Scope` in Deno, can you send me documentation?

KaanMol:
> > @richarddavison I had another problem, how would we allow passing values from LLRT to Rust context? Do you maybe have an idea? Deno has a `Scope` for example
> 
> Can you expand a bit what you mean pass values? I'm not familiar with `Scope` in Deno, can you send me documentation?

Their documentation is not the best, they dont have their runtime really documented.
I haven't really looked at it in dept, but maybe you have an idea how one would do this.

How could one start the runtime maybe pass the JS function some params from Rust and also get the data returned in the JS function to do something with it after the execution?

richarddavison:
> > > @richarddavison I had another problem, how would we allow passing values from LLRT to Rust context? Do you maybe have an idea? Deno has a `Scope` for example
> > 
> > 
> > Can you expand a bit what you mean pass values? I'm not familiar with `Scope` in Deno, can you send me documentation?
> 
> Their documentation is not the best, they dont have their runtime really documented. I haven't really looked at it in dept, but maybe you have an idea how one would do this.
> 
> How could one start the runtime maybe pass the JS function some params from Rust and also get the data returned in the JS function to do something with it after the execution?

This is very simple, for example call JS from Rust:
```javascript
globals.anyJsObjectOnGlobalScope = (arg) => {
  return arg;
}
```
```rust
let global_js_fn : Function = ctx.globals().get("anyJsObjectOnGlobalScope")?;
let result_string: String = global_js_fn.call(("hello",))?;
```

Or call Rust from JS:
```
rustFunction("hi")
```
```rust
ctx.globals().set("rustFunction",Func::from(|value:String|{
   //returns same value
   value
})
```

KaanMol:
> > > > @richarddavison I had another problem, how would we allow passing values from LLRT to Rust context? Do you maybe have an idea? Deno has a `Scope` for example
> > > 
> > > 
> > > Can you expand a bit what you mean pass values? I'm not familiar with `Scope` in Deno, can you send me documentation?
> > 
> > 
> > Their documentation is not the best, they dont have their runtime really documented. I haven't really looked at it in dept, but maybe you have an idea how one would do this.
> > How could one start the runtime maybe pass the JS function some params from Rust and also get the data returned in the JS function to do something with it after the execution?
> 
> This is very simple, for example call JS from Rust:
> 
> ```js
> globals.anyJsObjectOnGlobalScope = (arg) => {
>   return arg;
> }
> ```
> 
> ```rust
> let global_js_fn : Function = ctx.globals().get("anyJsObjectOnGlobalScope")?;
> let result_string: String = global_js_fn.call(("hello",))?;
> ```
> 
> Or call Rust from JS:
> 
> ```
> rustFunction("hi")
> ```
> 
> ```rust
> ctx.globals().set("rustFunction",Func::from(|value:String|{
>    //returns same value
>    value
> })
> ```

Oh thats amazing!

src/module_builder.rs
KaanMol: All names have been moved to the modules themself with a into :)

KaanMol:
@richarddavison I just implemented LLRT as a test in my own project, it brought it back from around 2/3 sec with Deno to around 100ms, I also ran LLRT's benchmarks and didn't notice a change before and after these changes :)

richarddavison:
> @richarddavison I just implemented LLRT as a test in my own project, it brought it back from around 2/3 sec with Deno to around 100ms, I also ran LLRT's benchmarks and didn't notice a change before and after these changes :)

Have you thought about moving out the Lambda specific implementations we do? For instance the console contains a lot of Lambda specific code, even at its core (for example treating newlines). I suspect the best thing right now would be to feature gate it. Another option could be to pass a config and use a closure for the exposed functions so we can pass config values down the the console argument handling (for example line endings, output etc).

KaanMol:
> > @richarddavison I just implemented LLRT as a test in my own project, it brought it back from around 2/3 sec with Deno to around 100ms, I also ran LLRT's benchmarks and didn't notice a change before and after these changes :)
> 
> Have you thought about moving out the Lambda specific implementations we do? For instance the console contains a lot of Lambda specific code, even at its core (for example treating newlines). I suspect the best thing right now would be to feature gate it. Another option could be to pass a config and use a closure for the exposed functions so we can pass config values down the the console argument handling (for example line endings, output etc).

Yes, like I mentioned previously, I thought of making `console` utils and implementing the module within `LLRT` while creating a second module, but in `LLRT-lambda` or some other name. The thing I actually find complicated is that some JS gets compiled for improved start-up times, and it also includes AWS libraries. Do you know if it would be easy to split that?

richarddavison:
> Yes, like I mentioned previously, I thought of making `console` utils and implementing the module within `LLRT` while creating a second module, but in `LLRT-lambda` or some other name. The thing I actually find complicated is that some JS gets compiled for improved start-up times, and it also includes AWS libraries. Do you know if it would be easy to split that?

Yes, that can be easily removed by feature gate or in this case remove that logic from the module loader and resolver:
https://github.com/awslabs/llrt/blob/bfaa41ecd20df9de971aa0358571e30c2a46ae3f/src/vm.rs#L225
https://github.com/awslabs/llrt/blob/bfaa41ecd20df9de971aa0358571e30c2a46ae3f/src/vm.rs#L346

Right now the binary resolver does some other things as well so that can complexity probably be refactored a bit.

KaanMol:
> > Yes, like I mentioned previously, I thought of making `console` utils and implementing the module within `LLRT` while creating a second module, but in `LLRT-lambda` or some other name. The thing I actually find complicated is that some JS gets compiled for improved start-up times, and it also includes AWS libraries. Do you know if it would be easy to split that?
> 
> Yes, that can be easily removed by feature gate or in this case remove that logic from the module loader and resolver:
> 
> https://github.com/awslabs/llrt/blob/bfaa41ecd20df9de971aa0358571e30c2a46ae3f/src/vm.rs#L225
> 
> 
> https://github.com/awslabs/llrt/blob/bfaa41ecd20df9de971aa0358571e30c2a46ae3f/src/vm.rs#L346
> 
> Right now the binary resolver does some other things as well so that can complexity probably be refactored a bit.

So I wonder if it wouldn't be a better idea to start splitting this PR up in multiple parts. It already got very big. 
Right now I have split LLRT up in 2 parts, one of them being the core and one is the executable itself.

How would you feel if we set the scope for this PR only to what it is now, where LLRT is 2 parts with a module selection etc. And after merging I would start splitting the AWS parts away. At least if a big refactor happens somewhere, I would not have to somehow merge that in this newer structure every time.

KaanMol:
@richarddavison Could you review this once you have time? :)
Splitting the AWS parts is something I am going to do in its own PR

richarddavison (changes requested):
Minor nit, but LGTM!

pack
```diff
@@ -94,7 +94,7 @@ const char *data = "$(printf "\\\x%02x" $num_parts)${INPUT_SIZES_STR}${OUTPUT_SI
 EOF
 
 cp "$1" "${1}.bak"
-cp src/main.c $temp_folder/
+cp llrt_core/src/main.c $temp_folder/
```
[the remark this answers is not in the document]
KaanMol: Got it, I will also note it down to move to the lambda specific package later :)

KaanMol:
@richarddavison Done!

richarddavison (approved):

PR #329, opened 2024-04-05
feat: Implement DOMException()

### LINKED ISSUE #330: Implement DOMException()

In llrt, `DOMException()` is not yet implemented and needs to be worked on.
https://developer.mozilla.org/en-US/docs/Web/API/DOMException

nabetti1720:
### Issue # (if available)
Fixes #330  

### Description of changes

This PR provides [DOMException()](https://developer.mozilla.org/en-US/docs/Web/API/DOMException), which is not yet implemented in `llrt`.

The unit test with javascript seems to be working well, but we have not yet implemented the return of exceptions from rust to javascript.

events.rs:
```rust
let abort_exception = Exception::from_value("AbortError".into_js(&ctx)?)?; // TODO: AbortError DOMException
```
```rust
let timeout_exception = Exception::from_value("TimeoutError".into_js(&ctx)?)?; // TODO: Timeout DOMException
```

What implementation is needed to eliminate this problem?
Any reference to implementations or sample code in already existing projects would be helpful.

### Checklist

### DIFF
--- a/src/exceptions.rs
+++ b/src/exceptions.rs
@@ -1,15 +1,31 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
+use rquickjs::{function::Opt, Class, Ctx, Result};
+use std::error::Error;
+use std::fmt::{self, Display};
 
-use rquickjs::{atom::PredefinedAtom, function::Opt, Class, Ctx, Result};
-
-#[derive(Clone, Default)]
+#[derive(Debug, Default)]
 #[rquickjs::class]
 #[derive(rquickjs::class::Trace)]
 pub struct DOMException {
     message: String,
     name: String,
 }
+
+impl Display for DOMException {
+    fn fmt(&self, _f: &mut fmt::Formatter) -> fmt::Result {
+        todo!()
+        /*
+        if self.message.is_empty() {
+            return self.name.clone();
+        }
+
+        format!("{}: {}", &self.name, &self.message)
+        */
+    }
+}
+
+impl Error for DOMException {}
 
 #[rquickjs::methods]
 impl<'js> DOMException {
@@ -29,15 +45,6 @@ impl<'js> DOMException {
     #[qjs(get, rename = "name")]
     pub fn get_name(&self) -> String {
         self.name.clone()
-    }
-
-    #[qjs(rename = PredefinedAtom::ToString)]
-    pub fn to_string(&self) -> String {
-        if self.message.is_empty() {
-            return self.name.clone();
-        }
-
-        format!("{}: {}", &self.name, &self.message)
     }
 }
 
--- a/src/exceptions.rs
+++ b/src/exceptions.rs
@@ -1,57 +1,69 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
-use rquickjs::{function::Opt, Class, Ctx, Result};
-use std::error::Error;
-use std::fmt::{self, Display};
+use rquickjs::{
+    atom::PredefinedAtom,
+    function::Constructor,
+    prelude::{Func, This},
+    Ctx, IntoJs, Object, Result, Value,
+};
 
-#[derive(Debug, Default)]
-#[rquickjs::class]
-#[derive(rquickjs::class::Trace)]
 pub struct DOMException {
     message: String,
     name: String,
+    stack: Result<String>,
 }
 
-impl Display for DOMException {
-    fn fmt(&self, _f: &mut fmt::Formatter) -> fmt::Result {
-        todo!()
-        /*
-        if self.message.is_empty() {
-            return self.name.clone();
-        }
-
-        format!("{}: {}", &self.name, &self.message)
-        */
-    }
-}
-
-impl Error for DOMException {}
-
-#[rquickjs::methods]
-impl<'js> DOMException {
-    #[qjs(constructor)]
-    pub fn new(_ctx: Ctx<'js>, message: Opt<String>, name: Opt<String>) -> Result<Self> {
-        Ok(Self {
-            message: message.0.unwrap_or(String::from("")),
-            name: name.0.unwrap_or(String::from("Error")),
-        })
-    }
-
-    #[qjs(get, rename = "message")]
-    fn get_message(&self) -> String {
-        self.message.clone()
-    }
-
-    #[qjs(get, rename = "name")]
-    pub fn get_name(&self) -> String {
-        self.name.clone()
+impl<'js> IntoJs<'js> for DOMException {
+    fn into_js(self, ctx: &Ctx<'js>) -> Result<Value<'js>> {
+        //TODO: It seems that the parameters are not being pulled in properly from the JavaScript side.
+        let message = self.message;
+        let name = self.name;
+        let val_message = message.into_js(ctx)?;
+        let val_name = name.into_js(ctx)?;
+        let constructor: Constructor = ctx.globals().get(stringify!(DOMException))?;
+        constructor.construct((val_message, val_name))
     }
 }
 
-pub fn init(ctx: &Ctx<'_>) -> Result<()> {
-    let globals = ctx.globals();
+//impl<'js> DOMException {}
 
-    Class::<DOMException>::define(&globals)?;
+fn message<'js>(_this: This<Object<'js>>, _ctx: Ctx<'js>) -> Result<String> {
+    // TODO: I don't know how to get the message property out of this object.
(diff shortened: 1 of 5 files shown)


src/exceptions.rs
```diff
@@ -0,0 +1,78 @@
...
+    let globals = ctx.globals();
+
+    Class::<DOMException>::define(&globals)?;
+
+    Ok(())
+}
+
+pub struct ExceptionsModule;
+
+impl ModuleDef for ExceptionsModule {
```
richarddavison, line 62: I don't think `DomException` should be an importable module? It would be enough to have it at global scope.

src/exceptions.rs
```diff
@@ -0,0 +1,78 @@
...
+#[derive(Clone, Default)]
+#[rquickjs::class]
+#[derive(rquickjs::class::Trace)]
+pub struct DOMException {
+    message: String,
+    name: String,
+}
+
+#[rquickjs::methods]
+impl<'js> DOMException {
```
richarddavison, line 21: I think we might need to inherit from `Error` in order for stack traces to be populated. Please verify that stacks are populated if throwing.

richarddavison (changes requested):

src/exceptions.rs
nabetti1720: We have responded to this one.

src/exceptions.rs
nabetti1720: Sorry @richarddavison , this effort seems to have been too difficult for me. I have tried many times, but I just can't get the implementation to combine the `error trait` and `rquickjs` and how to return it to the Javascript side as an Exception. Can you please assist me with the implementation?

In trying to incorporate `AbortError` and `TimeoutError`, I am not even sure if the `DOMException` structure is correct in this form.

richarddavison: Hi i think it’s fine to do what you’re already doing. You can test by just running `throw new DOMException()`

nabetti1720: Hi @richarddavison.
As for what I can do, I just added the missing test cases.

src/exceptions.rs
```diff
@@ -0,0 +1,52 @@
...
+    pub fn get_name(&self) -> String {
+        self.name.clone()
+    }
+
+    #[qjs(rename = PredefinedAtom::ToString)]
+    pub fn to_string(&self) -> String {
+        let mut message = String::new();
+
+        if !self.message.is_empty() {
+            message = format!(": {}", &self.message)
```
richarddavison, line 39: I think we should return here and avoid an additional format

tests/unit/exceptions.test.ts
```diff
@@ -0,0 +1,80 @@
...
+    expect(e.name).toEqual("def");
+  });
+  it("result of the toString method should be 'def: abc'", () => {
+    expect(e.toString()).toEqual("def: abc");
+  });
+  it("throwing and catching exceptions can be done", () => {
+    try {
+      throw new DOMException("abc", "def");
+    } catch (ex) {
+      expect(ex.toString()).toEqual(e.toString());
```
richarddavison, line 77: Can we also check for stack? If stack is not present we need to change the implementation to inheritance. We'll sort it together :)

richarddavison (changes requested):
Almost there!

src/exceptions.rs
nabetti1720: Is the implementation you have in mind something like the following?

```rust
    pub fn to_string(&self) -> String {
        if self.message.is_empty() {
            return self.name.clone();
        }

        format!("{}: {}", &self.name, &self.message)
    }
 ```

Or should I avoid the "Name: Message" format?

Comment (nabetti1720) [reply src/exceptions.rs]: Another effort, inherited from error, duplicated fmt(), which must be implemented.
Therefore, we decided to move the output processing to fmt().

File: src/exceptions.rs (hunk 2 above)
Comment (nabetti1720) [reply src/exceptions.rs:31]: We decided to go back to the drawing board and inherit the `Error` pointed out. We also added a definition of `fmt()`, which is mandatory to implement.
I ported the code from the `toString()` method, but since the return value is not a `String`, calling it from JavaScript seems to return an `object`, not a `string`. This is not important at this time, so I replaced it with a `todo!`.

Is the next step to allow DOMException() to hold Stack?

File: tests/unit/exceptions.test.ts (hunk 4 above)
Comment (nabetti1720) [reply tests/unit/exceptions.test.ts:77]: We are trying to work on error inheritance in the rust implementation. We will modify it accordingly, so we will leave it as it is for now.

Comment (richarddavison) [reply tests/unit/exceptions.test.ts:77]: Sorry for being unclear. This inheritance is in JavaScript user land. For reference see how `buffer.rs` inherits from `Uint8Array` here: https://github.com/awslabs/llrt/blob/7904eb84a34c2cfd20c6a58b46f288d8999a48a1/src/buffer.rs#L195

Comment (nabetti1720) [reply tests/unit/exceptions.test.ts:77]: I am sorry. I guess I have been very wrong so far. So this means that I can create a new type on the rust side by inheriting the type that exists on the Javascript side? In other words, I can create an init() like the following and create a DOMException?

```rust
pub fn init<'js>(ctx: &Ctx<'js>) -> Result<()> {
    let ex = ctx.eval::<Object<'js>, &str>(&format!(
        "class {0} extends Error {{}}\n{0}",
        stringify!(DOMException)
    ))?;
    set_prototype(ctx, ex)
}
```
If this is the case, we will need to take some time to review the implementation with reference to the implementation of buffers.rs and its relationship to Uint8Array.

richarddavison: Yes, that exactly it! Then you simply define objects on the prototype.

nabetti1720: Hi @richarddavison ,
Referring to `Buffer.rs`, we have made `Error` in JavaScript user land inherit for `DOMException`.

I think we are getting close to the way it should be, but there are some things that are not working and I left a `TODO:`. Could you please advise?

src/exceptions.rs
richarddavison: I think actually it would be simpler to do this:
```rust
#[rquickjs::class]
#[derive(rquickjs::class::Trace)]
struct DOMException {
    message: String,
    name: String,
    stack: String,
}

#[rquickjs::methods]
impl DOMException {
    #[qjs(constructor)]
    fn new<'js>(ctx: Ctx<'js>, message: String, name: String) -> Result<Self> {
        let error_ctor: Constructor = ctx.globals().get(PredefinedAtom::Error)?;
        let new: Object = error_ctor.construct((message.clone(),))?;

        Ok(Self {
            name,
            message,
            stack: new.get::<_, String>(PredefinedAtom::Stack)?,
        })
    }

    #[qjs(get)]
    fn name(&self) -> String {
        self.name.clone()
    }

    #[qjs(get)]
    fn stack(&self) -> String {
        self.stack.clone()
    }

    #[qjs(get)]
    fn message(&self) -> String {
        self.message.clone()
    }
}
```
And then set inheritance by setting the prototype, where you define the class (in init)
```rust
Class::<DOMException>::define(&globals)?;
let dom_ex_proto = Class::<DOMException2>::prototype(ctx.clone()).unwrap();
let error_ctor: Object = globals.get(PredefinedAtom::Error)?;
let error_proto = error_ctor.get_prototype();
dom_ex_proto.set_prototype(error_proto.as_ref())?;
```

nabetti1720: Thanks for the detailed advice.
We have modified the implementation based on what has been pointed out to us.
In addition, we were able to pass all the test cases we had prepared.

richarddavison (dismissed):

src/exceptions.rs
```diff
@@ -0,0 +1,69 @@
...
+#[rquickjs::class]
+#[derive(rquickjs::class::Trace)]
+struct DOMException {
+    message: String,
+    name: String,
+    stack: String,
+}
+
+#[rquickjs::methods]
+impl DOMException {
```
richarddavison, line 18: Only toString left and we're good to go! Nice job!

richarddavison (changes requested):

src/exceptions.rs
nabetti1720: In other words, toString() is unnecessary!

richarddavison: It's not in spec because of `Error` inheritance. Since we're not really inheriting just adding to prototype chain we must manually implement it as well

nabetti1720: What do you mean by "manually implement it"?
I believe we are probably missing a test case, since all of the test cases we currently have in place are passing.
What test cases should we create to detect the lack of manual implementation?

richarddavison: I mean add it to the `impl` and create a test case for it :)

richarddavison (approved):

Issue #340, opened 2024-04-20
Redirects are not handled in fetch()


nabetti1720:
In other implementations (`node`, `bun`, etc.), when fetching the specified URI, a redirect is processed and the final response (status code 200) is returned, but in `llrt`, the redirect is not processed and status code 301 is returned.

Reproduction procedure:
```javascript
// fetch.js
const main = async () => {
  try {
    const response = await fetch('https://google.com');
    console.log(response.status);
    console.log(response.headers.get('location'));
  } catch(e) {
    console.log(e);
  }
}

main();
```

Execution results in `llrt`:
```bash
shinya@MBA2022M2 llrt-test % ./llrt fetch.js 
301
https://www.google.com/
```

Execution results in `bun`:
```bash
shinya@MBA2022M2 llrt-test % bun fetch.js 
200
null
```

richarddavison:
Thanks for the well documented report. PR is welcome, want to take shot at it?
This could be implemented as looping over the response until no redirect header is present if `redirect` option is set to `follow` on the `Request` or fetch options.
https://github.com/awslabs/llrt/blob/d05da26103acc1fdf45afe15965ef05013e5f488/src/http/fetch.rs#L81

nabetti1720:
Hi @richarddavison ,
I would love to work on this assignment if the opportunity arises :)

PR #346, opened 2024-04-26
feat: Implementation of redirect option for fetch()

nabetti1720:
### Issue # (if available)

#340 

### Description of changes

- Minimal improvements have been made to allow for redirect processing.[(reference information)](https://fetch.spec.whatwg.org/#http-redirect-fetch)

### Checklist

### DIFF
--- a/src/http/fetch.rs
+++ b/src/http/fetch.rs
@@ -63,7 +63,8 @@ pub(crate) fn init(ctx: &Ctx<'_>, globals: &Object) -> Result<()> {
             async move {
                 let options = options?;
 
-                let mut uri: Uri = options.url.parse().or_throw(&ctx)?;
+                let initial_uri: Uri = options.url.parse().or_throw(&ctx)?;
+                let mut uri: Uri = initial_uri.clone();
                 let method_string = options.method.to_string();
                 let method = options.method;
                 let abort_receiver = options.abort_receiver;
@@ -80,6 +81,7 @@ pub(crate) fn init(ctx: &Ctx<'_>, globals: &Object) -> Result<()> {
                         &options.headers,
                         &options.body,
                         &response_status,
+                        &initial_uri,
                     )?;
 
                     let res = if let Some(abort_receiver) = &abort_receiver {
@@ -129,7 +131,10 @@ fn build_request(
     headers: &Option<Headers>,
     body: &Full<Bytes>,
     prev_status: &u16,
+    initial_uri: &Uri,
 ) -> Result<Request<Full<Bytes>>> {
+    let same_origin = is_same_origin(uri, initial_uri);
+
     let change_method = should_change_method(*prev_status, method);
 
     let (method_to_use, req_body) = if change_method {
@@ -146,7 +151,11 @@ fn build_request(
 
     if let Some(headers) = headers {
         for (key, value) in headers.iter() {
-            if change_method && is_request_body_header_name(key.as_str()) {
+            let header_name = key.as_str();
+            if change_method && is_request_body_header_name(header_name) {
+                continue;
+            }
+            if !same_origin && is_cors_non_wildcard_request_header_name(header_name) {
                 continue;
             }
             req = req.header(key, value)
@@ -154,6 +163,24 @@ fn build_request(
     }
 
     req.body(req_body).or_throw(ctx)
+}
+
+fn is_same_origin(uri: &Uri, initial_uri: &Uri) -> bool {
+    is_same_scheme(uri, initial_uri)
+        && is_same_host(uri, initial_uri)
+        && is_same_port(uri, initial_uri)
+}
+
+fn is_same_scheme(uri: &Uri, initial_uri: &Uri) -> bool {
+    uri.scheme() == initial_uri.scheme()
+}
+
+fn is_same_host(uri: &Uri, initial_uri: &Uri) -> bool {
+    uri.host() == initial_uri.host()
+}
+
+fn is_same_port(uri: &Uri, initial_uri: &Uri) -> bool {
+    uri.authority().and_then(|a| a.port()) == initial_uri.authority().and_then(|a| a.port())
 }
 
 fn should_change_method(prev_status: u16, method: &Method) -> bool {
@@ -173,6 +200,10 @@ fn is_request_body_header_name(key: &str) -> bool {
         key,
         "content-encoding" | "content-language" | "content-location" | "content-type"
     )
+}
+
+fn is_cors_non_wildcard_request_header_name(key: &str) -> bool {
+    matches!(key, "authorization")
 }
 
 struct FetchOptions<'js> {
@@ -332,5 +363,77 @@ mod tests {
 
         assert!(!is_request_body_header_name("content-length"));
         assert!(!is_request_body_header_name("accept"));
+    }
+
+    #[test]
+    fn test_is_same_origin() {
+        let uri1 = Uri::from_static("https://example.com:8080/path");
+        let uri2 = Uri::from_static("https://example.com:8080/path");
+
+        assert!(is_same_origin(&uri1, &uri2));
+
+        let uri3 = Uri::from_static("http://example.com/path");
+        let uri4 = Uri::from_static("https://example.com/path");
+
+        assert!(!is_same_origin(&uri3, &uri4));
+
+        let uri5 = Uri::from_static("https://example.com:8080/path");
+        let uri6 = Uri::from_static("https://example.org:8080/path");
+
+        assert!(!is_same_origin(&uri5, &uri6));
+
+        let uri7 = Uri::from_static("https://example.com:8080/path");
+        let uri8 = Uri::from_static("https://example.com:8081/path");
+
+        assert!(!is_same_origin(&uri7, &uri8));
+    }
+
+    #[test]
+    fn test_is_same_scheme() {
+        let uri1 = Uri::from_static("https://example.com");
+        let uri2 = Uri::from_static("https://example.com");
+
+        assert!(is_same_scheme(&uri1, &uri2));
+
+        let uri3 = Uri::from_static("http://example.com");
+        let uri4 = Uri::from_static("https://example.com");
+
+        assert!(!is_same_scheme(&uri3, &uri4));
+    }
+
+    #[test]
+    fn test_is_same_host() {
+        let uri1 = Uri::from_static("https://example.com");
+        let uri2 = Uri::from_static("https://example.com");
+
+        assert!(is_same_host(&uri1, &uri2));
+
+        let uri3 = Uri::from_static("https://example.com");
+        let uri4 = Uri::from_static("https://example.org");
+
+        assert!(!is_same_host(&uri3, &uri4));
+    }
+
+    #[test]
+    fn test_is_same_port() {
+        let uri1 = Uri::from_static("https://example.com:8080");
+        let uri2 = Uri::from_static("https://example.com:8080");
+
+        assert!(is_same_port(&uri1, &uri2));
+
+        let uri3 = Uri::from_static("https://example.com:8080");
+        let uri4 = Uri::from_static("https://example.com:9090");
+
+        assert!(!is_same_port(&uri3, &uri4));
+
+        let uri5 = Uri::from_static("https://example.com");
+        let uri6 = Uri::from_static("https://example.com");
+
+        assert!(is_same_port(&uri5, &uri6));
+
+        let uri7 = Uri::from_static("https://example.com:8080");
+        let uri8 = Uri::from_static("https://example.com");
+
+        assert!(!is_same_port(&uri7, &uri8));
     }
 }
--- a/src/http/fetch.rs
+++ b/src/http/fetch.rs
@@ -130,15 +130,16 @@ fn build_request(
     body: &Full<Bytes>,
     prev_status: &u16,
 ) -> Result<Request<Full<Bytes>>> {
-    let method_clone = method.clone();
-    let change_method = (matches!(prev_status, 301 | 302) && matches!(method_clone, Method::POST))
-        || (matches!(prev_status, 303) && !matches!(method_clone, Method::GET | Method::HEAD));
+    let change_method = should_change_method(*prev_status, method);
+
+    let (method_to_use, req_body) = if change_method {
+        (Method::GET, Full::default())
+    } else {
+        (method.clone(), body.clone())
+    };
 
     let mut req = Request::builder()
-        .method(match change_method {
-            true => Method::GET,
-            _ => method_clone,
-        })
+        .method(method_to_use)
         .uri(uri.clone())
         .header("user-agent", format!("llrt {}", VERSION))
         .header("accept", "*/*");
@@ -152,11 +153,19 @@ fn build_request(
         }
     }
 
-    if change_method {
-        req.body(Full::default()).or_throw(ctx)
-    } else {
-        req.body(body.clone()).or_throw(ctx)
+    req.body(req_body).or_throw(ctx)
+}
+
+fn should_change_method(prev_status: u16, method: &Method) -> bool {
+    if matches!(prev_status, 301 | 302) {
+        return *method == Method::POST;
     }
+
+    if prev_status == 303 {
+        return !matches!(*method, Method::GET | Method::HEAD);
+    }
+
+    false
 }
 
 fn is_request_body_header_name(key: &str) -> bool {
--- a/src/http/fetch.rs
+++ b/src/http/fetch.rs
@@ -2,7 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 use bytes::Bytes;
 use http_body_util::Full;
-use hyper::{Request, Uri};
+use hyper::{header::HeaderName, Request, Uri};
 
 use rquickjs::{
     atom::PredefinedAtom,
@@ -25,6 +25,8 @@ use crate::{
 use crate::{security::HTTP_ALLOW_LIST, VERSION};
 
 use super::response::Response;
+
+const MAX_REDIRECT_COUNT: u32 = 20;
 
 pub(crate) fn init(ctx: &Ctx<'_>, globals: &Object) -> Result<()> {
     if let Some(Err(err)) = &*HTTP_ALLOW_LIST {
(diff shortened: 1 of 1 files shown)


nabetti1720:
We will continue to address the following points:
1. return network error response.
2. Replace only the url when build_request() is called multiple times.

src/http/fetch.rs
```diff
@@ -61,33 +63,48 @@ pub(crate) fn init(ctx: &Ctx<'_>, globals: &Object) -> Result<()> {
...
+                            if let Ok(location_str) = location_headers.to_str() {
+                                uri = location_str.parse().or_throw(&ctx)?;
+                                ensure_url_access(&ctx, &uri)?;
+                            }
+                        },
+                        None => break res,
+                    };
+
+                    if options.redirect == "manual" {
+                        return Err(Exception::throw_type(&ctx, "Unsupported redirect option"));
```
richarddavison, line 97: We can just ignore the redirect here and return the raw response

src/http/fetch.rs
```diff
@@ -61,33 +63,48 @@ pub(crate) fn init(ctx: &Ctx<'_>, globals: &Object) -> Result<()> {
...
+                    } else if options.redirect == "error" {
+                        // TODO: Return Response({type: “error”, status: 0}) to terminate fetch.
+                        return Err(Exception::throw_message(&ctx, "TODO: Not Implemented options.redirect: errror"));
```
richarddavison, line 100: Its fine to just retorn exception here:
`Error: unexpected redirect`

src/http/fetch.rs
```diff
@@ -61,33 +63,48 @@ pub(crate) fn init(ctx: &Ctx<'_>, globals: &Object) -> Result<()> {
...
+                    };
 
-                if let Some(headers) = options.headers {
-                    for (key, value) in headers.iter() {
-                        req = req.header(key, value)
+                    match res.headers().get(HeaderName::from_static("location")) {
+                        Some(location_headers) => {
+                            if let Ok(location_str) = location_headers.to_str() {
+                                uri = location_str.parse().or_throw(&ctx)?;
+                                ensure_url_access(&ctx, &uri)?;
```
richarddavison, line 90: Nice!

src/http/fetch.rs
```diff
@@ -168,6 +210,18 @@ fn get_fetch_options<'js>(
         {
             abort_receiver = Some(signal.borrow().sender.subscribe());
         }
+
+        if let Some(redirect_opt) = get_option::<String>("redirect", &arg_opts, &resource_opts)? {
+            redirect = Some(match redirect_opt.as_str() {
```
richarddavison, line 215: We can probably do something like:
```rust
let redirect_str = redirect_opt.as_str();
if !matches!(redirect_str, "follow" | "manual" | "error"){
   return Err(....)
}
```

src/http/fetch.rs
```diff
@@ -117,6 +157,8 @@ fn get_fetch_options<'js>(
     let mut method = None;
     let mut body = None;
     let mut abort_receiver = None;
+    let mut redirect = None;
```
richarddavison, line 160: Maybe an empty mut string instead and then `push_str() if option is set`. Then we can avoid `unwrap_or_default.to_string()`?

richarddavison (changes requested):

src/http/fetch.rs
nabetti1720: We have corrected the problem, please confirm.

src/http/fetch.rs
nabetti1720: We have corrected the problem, please confirm.

src/http/fetch.rs
nabetti1720: We have corrected the problem, please confirm.

src/http/fetch.rs
nabetti1720: We have corrected the problem, please confirm.

src/http/fetch.rs
nabetti1720: We have already fixed it, please review.

nabetti1720:
We have revised our policy a bit.

> 1. Return network error response.

Since it has already been pointed out in the review that simply returning an Error is sufficient, we are not going to do this in the immediate future.
If necessary, we would like to implement a method to generate and return a [Network error response](https://fetch.spec.whatwg.org/#concept-network-error) by calling something like `Respose::error(ctx)`.

> 2. Replace only the url when build_request() is called multiple times.

Reading through the [HTTP-redirect fetch](https://fetch.spec.whatwg.org/#http-redirect-fetch), we see that we need to control multiple request parameters, not just the url, depending on the method and status code.
To control this, we would like to keep the `build_request()` to generate the next request parameter.
After this, commit additional modifications.(I wish I could have implemented it a little better...)

richarddavison:
Thanks again for the PR. I think it LGTM. 🚀

> We have revised our policy a bit.
> 
> > 1. Return network error response.
> 
> Since it has already been pointed out in the review that simply returning an Error is sufficient, we are not going to do this in the immediate future. If necessary, we would like to implement a method to generate and return a [Network error response](https://fetch.spec.whatwg.org/#concept-network-error) by calling something like `Response::error(ctx)`.
> 

What value would this bring?

> > 2. Replace only the url when build_request() is called multiple times.
> 
> Reading through the [HTTP-redirect fetch](https://fetch.spec.whatwg.org/#http-redirect-fetch), we see that we need to control multiple request parameters, not just the url, depending on the method and status code. To control this, we would like to keep the `build_request()` to generate the next request parameter. After this, commit additional modifications.
> 

You can ignore the “cors” logic as it’s mainly relevant to browsers and is not planned to be supported.

> > 3. Add test code
> 
> I am wondering if it is possible to do a `fetch()` redirect test without accessing the actual site. Any good ideas? Note that the minimum test for the pattern necessary to make a decision was implemented in rust.

Yes, runtime_client.rs has tests using wiremock:
https://github.com/awslabs/llrt/blob/bfaa41ecd20df9de971aa0358571e30c2a46ae3f/src/runtime_client.rs#L500

nabetti1720:
> Thanks again for the PR. I think it LGTM. 🚀
> 
> > We have revised our policy a bit.
> > > 1. Return network error response.
> > 
> > 
> > Since it has already been pointed out in the review that simply returning an Error is sufficient, we are not going to do this in the immediate future. If necessary, we would like to implement a method to generate and return a [Network error response](https://fetch.spec.whatwg.org/#concept-network-error) by calling something like `Response::error(ctx)`.
> 
> What value would this bring?

Yes, I have not found a use case for `llrt` at this time. So I do not think that implementation is necessary.

> > > 2. Replace only the url when build_request() is called multiple times.
> > 
> > 
> > Reading through the [HTTP-redirect fetch](https://fetch.spec.whatwg.org/#http-redirect-fetch), we see that we need to control multiple request parameters, not just the url, depending on the method and status code. To control this, we would like to keep the `build_request()` to generate the next request parameter. After this, commit additional modifications.
> 
> You can ignore the “cors” logic as it’s mainly relevant to browsers and is not planned to be supported.

Understood. So, would it be safe to assume that we have roughly completed the implementation?
Please point out any specifications that are of concern.

> > > 3. Add test code
> > 
> > 
> > I am wondering if it is possible to do a `fetch()` redirect test without accessing the actual site. Any good ideas? Note that the minimum test for the pattern necessary to make a decision was implemented in rust.
> 
> Yes, runtime_client.rs has tests using wiremock:
> 
> https://github.com/awslabs/llrt/blob/bfaa41ecd20df9de971aa0358571e30c2a46ae3f/src/runtime_client.rs#L500

Thank you for sharing your excellent samples!
This is the first implementation I've seen, so it may take some time, but I'll get to work on creating a mock server as soon as possible :)

nabetti1720:
I am creating a test with a mock server as soon as possible, but the `fetch()` call part is not working properly.
I am thinking of using `fetch()` defined in the `globals` of `vm`. Is the procedure correct?

``` rust
let vm = Vm::new().await.unwrap();

async_with!(vm.ctx => |ctx| {
    async move {
        let result = ctx.globals().fetch(format!("http://{}/expect/200/", mock_server.address()), None).await;
        assert!(result.is_ok());
    }
})
.await;

vm.runtime.idle().await;
```

```bash
error[E0599]: no method named `fetch` found for struct `rquickjs::Object<'_>` in the current scope
   --> src/http/fetch.rs:490:44
    |
490 |                 let result = ctx.globals().fetch(format!("http://{}/expect/200/", mock_server.address()), None).await;
    |                                            ^^^^^ method not found in `Object<'_>`
```

richarddavison:
> ```rust
>  ctx.globals().fetch(
> ```

You would have to access the global object like this:

```rust
let fetch = ctx.globals().get("fetch")?;
let res: Promise<Value> = fetch(....)?;
let response = res.await?;
```

nabetti1720:
My apologies. I have tried everything, but I cannot resolve the error that the type of fetch is not specified. What is the type of fetch in this case?

```rust
let vm = Vm::new().await.unwrap();

async_with!(vm.ctx => |ctx| {
    async move {
        let fetch = ctx.globals().get("fetch")?;
        let res: Promise<Value> = fetch(format!("http://{}/expect/200/", mock_server.address()), None)?;
        let response = res.await?;
        assert!(response.is_ok());
        Ok(())
    }
})
.await;

vm.runtime.idle().await;
```

```bash
error[E0282]: type annotations needed
   --> src/http/fetch.rs:490:21
    |
490 |                 let fetch = ctx.globals().get("fetch")?;
    |                     ^^^^^
491 |                 let res: Promise<Value> = fetch(format!("http://{}/expect/200/", mock_server.address()), None)?;
    |                                           -------------------------------------------------------------------- type must be known at this point
    |
help: consider giving `fetch` an explicit type
    |
490 |                 let fetch: /* Type */ = ctx.globals().get("fetch")?;
    |                          ++++++++++++
```

richarddavison:
> My apologies. I have tried everything, but I cannot resolve the error that the type of fetch is not specified. What is the type of fetch in this case?

No need to apologize. Fetch is a `Function` :)
```rust
let fetch: Function = ...
```

nabetti1720:
We are continuing to try things, but there are two issues.
1. When I try to use `assert!` in `async_with!, I get an error. Therefore, I am wondering if it is possible to bring the response outside.
2. Is it OK if the type is not specified in `response` as in `fetch`?(Issue No. 1 has not been resolved and not confirmed.)

```rust
let vm = Vm::new().await.unwrap();

async_with!(vm.ctx => |ctx| {
    async move {
        let fetch: Function = ctx.globals().get("fetch")?;
        let url = format!("http://{}/expect/200/", mock_server.address());
        let options = {};
        let res: Promise<Value> = fetch.call((url, options))?;
        let response = res.await?;
        // The assert macro cannot be used in asynchronous functions.
        // assert!(response.is_ok()); 
        Ok(())
    }
})
.await;

vm.runtime.idle().await;
}
```

```bash
error[E0282]: type annotations needed
   --> src/http/fetch.rs:497:17
    |
497 |                 Ok(())
    |                 ^^ cannot infer type of the type parameter `E` declared on the enum `Result`
    |
help: consider specifying the generic arguments
    |
497 |                 Ok::<(), E>(())
    |                   +++++++++
```

nabetti1720:
> Yes, you can return the result from the await right away:

I'm sorry. I continue to have trouble getting res out of the async block.
I asked ChatGPT and got several suggestions for how to write L494, but they were all no good...

```bash
error[E0277]: the `?` operator can only be used in an async block that returns `Result` or `Option` (or another type that implements `FromResidual`)
   --> src/http/fetch.rs:490:65
    |
489 | /             async move {
490 | |                 let fetch: Function = ctx.globals().get("fetch")?;
    | |                                                                 ^ cannot use the `?` operator in an async block that returns `rquickjs::Value<'_>`
491 | |                 let url = format!("http://{}/expect/200/", mock_server.address());
492 | |                 let options = {};
493 | |                 let res: Promise<Value> = fetch.call((url, options))?;
494 | |                 res.await.catch(&ctx).unwrap()
495 | |             }}).await;
    | |_____________- this function should return `Result` or `Option` to accept `?`
    |
    = help: the trait `FromResidual<std::result::Result<Infallible, rquickjs::Error>>` is not implemented for `rquickjs::Value<'_>`

error[E0277]: the `?` operator can only be used in an async block that returns `Result` or `Option` (or another type that implements `FromResidual`)
   --> src/http/fetch.rs:493:69
    |
489 | /             async move {
490 | |                 let fetch: Function = ctx.globals().get("fetch")?;
491 | |                 let url = format!("http://{}/expect/200/", mock_server.address());
492 | |                 let options = {};
493 | |                 let res: Promise<Value> = fetch.call((url, options))?;
    | |                                                                     ^ cannot use the `?` operator in an async block that returns `rquickjs::Value<'_>`
494 | |                 res.await.catch(&ctx).unwrap()
495 | |             }}).await;
    | |_____________- this function should return `Result` or `Option` to accept `?`
    |
    = help: the trait `FromResidual<std::result::Result<Infallible, rquickjs::Error>>` is not implemented for `rquickjs::Value<'_>`

error[E0308]: mismatched types
   --> src/http/fetch.rs:488:43
    |
488 |            let fetch_result: Result<Value> = async_with!(vm.ctx => |ctx| {
    |  ____________________________________________^
489 | |/             async move {
490 | ||                 let fetch: Function = ctx.globals().get("fetch")?;
491 | ||                 let url = format!("http://{}/expect/200/", mock_server.address());
492 | ||                 let options = {};
493 | ||                 let res: Promise<Value> = fetch.call((url, options))?;
494 | ||                 res.await.catch(&ctx).unwrap()
495 | ||             }}).await;
    | ||_____________-_^ expected `Result<Value<'_>, Error>`, found `async` block
    |  |_____________|
    |                the found `async` block
    |
    = note:       expected enum `std::result::Result<rquickjs::Value<'_>, rquickjs::Error>`
            found `async` block `{async block@src/http/fetch.rs:489:13: 495:14}`
    = note: this error originates in the macro `async_with` (in Nightly builds, run with -Z macro-backtrace for more info)
```

nabetti1720:
> Sorry my misstake, I forgot to remove `unwrap`. You need to do that if you want to return Result using `?` from the async block. Options is also wrong here: `{}` in rust is a closure, not a JS object. You can pass a `Request` object or a JS object like this:

Thank you very much. However, when creating the initial object for the `options`, it seems that `ctx` is not allowed in this position because of the `AsyncContext`.

I also removed the `unwrap()` on the last line of the `async` block, but the error did not change significantly.

The codes and errors at this time are as follows.

```rust
let vm = Vm::new().await.unwrap();

let fetch_result: Result<Value> = async_with!(vm.ctx => |ctx| {
    async move {
        let fetch: Function = ctx.globals().get("fetch")?;
        let options = Object::new(ctx).unwrap();
        options.set("redirect","manual").unwrap();

        let url = format!("http://{}/expect/200/", mock_server.address());

        let res: Promise<Value> = fetch.call((url, options))?;
        res.await.catch(&ctx)
    }}).await;

let fetch_response = Class::<Response>::from_value(fetch_result.unwrap()).unwrap();
let response = fetch_response.borrow();
assert!(response.status() == 200);

vm.runtime.idle().await;
```

```bash
error[E0277]: `?` couldn't convert the error to `CaughtError<'_>`
   --> src/http/fetch.rs:490:65
    |
490 |                 let fetch: Function = ctx.globals().get("fetch")?;
    |                                                                 ^ the trait `std::convert::From<rquickjs::Error>` is not implemented for `CaughtError<'_>`, which is required by `std::result::Result<rquickjs::Value<'_>, CaughtError<'_>>: FromResidual<std::result::Result<Infallible, rquickjs::Error>>`
    |
    = note: the question mark operation (`?`) implicitly performs a conversion on the error value using the `From` trait
    = help: the following other types implement trait `FromResidual<R>`:
              <std::result::Result<T, F> as FromResidual<Yeet<E>>>
              <std::result::Result<T, F> as FromResidual<std::result::Result<Infallible, E>>>
    = note: required for `std::result::Result<rquickjs::Value<'_>, CaughtError<'_>>` to implement `FromResidual<std::result::Result<Infallible, rquickjs::Error>>`

error[E0277]: `?` couldn't convert the error to `CaughtError<'_>`
   --> src/http/fetch.rs:496:69
    |
496 |                 let res: Promise<Value> = fetch.call((url, options))?;
    |                                                                     ^ the trait `std::convert::From<rquickjs::Error>` is not implemented for `CaughtError<'_>`, which is required by `std::result::Result<rquickjs::Value<'_>, CaughtError<'_>>: FromResidual<std::result::Result<Infallible, rquickjs::Error>>`
    |
    = note: the question mark operation (`?`) implicitly performs a conversion on the error value using the `From` trait
    = help: the following other types implement trait `FromResidual<R>`:
              <std::result::Result<T, F> as FromResidual<Yeet<E>>>
              <std::result::Result<T, F> as FromResidual<std::result::Result<Infallible, E>>>
    = note: required for `std::result::Result<rquickjs::Value<'_>, CaughtError<'_>>` to implement `FromResidual<std::result::Result<Infallible, rquickjs::Error>>`

error[E0308]: mismatched types
   --> src/http/fetch.rs:488:43
    |
488 |            let fetch_result: Result<Value> = async_with!(vm.ctx => |ctx| {
    |  ____________________________________________^
489 | |/             async move {
490 | ||                 let fetch: Function = ctx.globals().get("fetch")?;
491 | ||                 let options = Object::new(ctx).unwrap();
492 | ||                 options.set("redirect","manual").unwrap();
...   ||
497 | ||                 res.await.catch(&ctx)
498 | ||             }}).await;
    | ||_____________-_^ expected `Result<Value<'_>, Error>`, found `async` block
    |  |_____________|
    |                the found `async` block
    |
    = note:       expected enum `std::result::Result<rquickjs::Value<'_>, rquickjs::Error>`
            found `async` block `{async block@src/http/fetch.rs:489:13: 498:14}`
    = note: this error originates in the macro `async_with` (in Nightly builds, run with -Z macro-backtrace for more info)
```

nabetti1720:
The following lines are also available in `?` -> `.unwrap()`, the error that occurred in this line disappeared.

> ```rust
>         ...
>         let fetch: Function = ctx.globals().get("fetch")?;
>         ...
>         let res: Promise<Value> = fetch.call((url, options))?;
>         ...
> ```

I cannot resolve the L498 error.

richarddavison:
No more pseudo code! I've tested this and it compiles. We make all the assertions inside the async block and catch the exceptions inside the async_with! macro as they can't be returned outside of it. 
```rust
async_with!(vm.ctx => |ctx| {
    let globals = ctx.globals();
    let run = async {
        let options = Object::new(ctx.clone())?;
        options.set("redirect", "manual")?;

        let fetch: Function = globals.get("fetch")?;
        let url = format!("http://{}/expect/200/", "url");
        let response_promise: Promise<Value> = fetch.call((url,options))?;
        let response = response_promise.await?;
        let response = Class::<Response>::from_value(response)?;
        let response = response.borrow();

        //do more tests and assertions in here
        assert_eq!(response.status(),200);

        Ok(())
    };
    run.await.catch(&ctx).unwrap();
})
.await;
```

nabetti1720:
Hi @richarddavison, Thank you for your dedication and support!
Finally, the `cargo test` started to pass when accessing the mock server I created.

richarddavison (approved):
LGTM

PR #358, opened 2024-05-03
feat: Add Response: redirected property

nabetti1720:
### Description of changes

- Added redirected property to response object.
See also. https://developer.mozilla.org/en-US/docs/Web/API/Response/redirected

### Checklist


llrt_core/src/modules/http/fetch.rs
```diff
@@ -73,7 +73,7 @@ pub(crate) fn init(ctx: &Ctx<'_>, globals: &Object) -> Result<()> {
 
                 let mut redirect_count = 0;
                 let mut response_status = 0;
-                let (res, uri) = loop {
+                let (res, uri, redirected) = loop {
```
richarddavison, line 76: Can we avoid returning a tuple here and just mutate variables outside the loop? Whats the benefit of this approach?
```rust
let mut uri;
let mut redirected;
```

richarddavison (commented):
Awesome, LGTM. Just a minor question/feedback

llrt_core/src/modules/http/fetch.rs
nabetti1720: Thanks for the review.

Of course it is possible. There is no deep reason for this, I just thought it would be clearer as a variable to use after this.

In addition, I noticed that the `redirect_count`, which is the basis for determining the variable `redirected` is also declared in the same way, so by setting the argument of `Response::from_incoming()` to `!matches!(redirect_count, 0)`, I noticed that the declaration itself is not needed.

Which do you prefer?

richarddavison (approved):

PR #359, opened 2024-05-04
fix: Avoid re-running build.rs on every build, even when nothing changed

refi64:
### Description of changes

from commit message:

Because build.rs depends on the bundle/ directory, whenever it also writes into it, the directory's timestamp is updated. This leads Cargo to always consider it outdated and re-run build.rs every time.

To fix this, separate the inputs to build.rs (the JS files) into a separate directory, bundle/js/, and move the compiled .lrt files to bundle/lrt/. This will keep their timestamps separate.

### Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,9 +27,9 @@ jobs:
         run: cargo fmt --all -- --check
       - name: Clippy
         run: | #create mock js files
-          mkdir -p bundle
+          mkdir -p bundle/js
           for i in {1..5}; do
-            echo "console.log(123);" > "bundle/test$i.js"
+            echo "console.log(123);" > "bundle/js/test$i.js"
           done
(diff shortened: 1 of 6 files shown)


richarddavison (approved):
Thanks for fixing. This has been a bit annoying.

richarddavison:
@refi64 CI is failing because it requires som files in `bundle/js` to generate compression dict, see:
https://github.com/awslabs/llrt/blob/80daaadabeb314e06ebff0574da562c362e08e2a/.github/workflows/ci.yml#L31C10-L33C15
This feels a bit hacky, but I don't want project to build if the required js files haven't been provided.

refi64:
@richarddavison ah thanks, was just trying to figure out why I couldn't repro it on my end :sweat_smile: looks like that's the only use of `bundle/` in the CI scripts so I think it should be working now.

PR #360, opened 2024-05-05
feat: Implementation of decompression in fetch()

nabetti1720:
### Description of changes

- Implementation of decompression in fetch()
Supported Encodings: gzip, deflate, br, zstd

### Execution Sample
content-encoding.js:
```javascript
const main = async () => {
  try {
    const response = await fetch('https://www.meta.com/', {headers: {"accept-encoding": process.argv[2]}});
    console.log(response.headers.get('content-encoding'));
    const responseText = await response.text();
    console.log('vvv');
    console.log(responseText.substring(0, 100));
    console.log('^^^');
  } catch(e) {
    console.log(e);
  }
}

main();
```

1. accept-encoding: `zstd`
```
shinya@MBA2022M2 llrt-test % ./llrt content-encoding.js zstd   
zstd
vvv
<!DOCTYPE html><html id="facebook" class="_a4ch _aber" lang="ja" dir="ltr"><head><link data-default-
^^^
```
2. accept-encoding: `br`
```
shinya@MBA2022M2 llrt-test % ./llrt content-encoding.js br  
br
vvv
<!DOCTYPE html><html id="facebook" class="_a4ch _aber" lang="ja" dir="ltr"><head><link data-default-
^^^
```
3. accept-encoding: `gzip`
```
shinya@MBA2022M2 llrt-test % ./llrt content-encoding.js gzip
gzip
vvv
<!DOCTYPE html><html id="facebook" class="_a4ch _aber" lang="ja" dir="ltr"><head><link data-default-
^^^
```
4. accept-encoding: `deflate`
```
shinya@MBA2022M2 llrt-test % ./llrt content-encoding.js deflate
undefined       // NOTE: This website does not support deflate.
vvv
<!DOCTYPE html><html id="facebook" class="_a4ch _aber" lang="ja" dir="ltr"><head><link data-default-
^^^
```
5. accept-encoding: `gzip, deflate, br, zstd`
```
shinya@MBA2022M2 llrt-test % ./llrt content-encoding.js "gzip, deflate, br, zstd"
zstd            // NOTE: The encoding is determined on the server side.
vvv
<!DOCTYPE html><html id="facebook" class="_a4ch _aber" lang="ja" dir="ltr"><head><link data-default-
^^^
```

### ToDo

- [x] Add test code. (`Response.text()` or `response.json()` does not work well in Mock Server testing...)
- ~~[ ] Verify `deflate` decompression on the actual website. (I would like to know if anyone has found it anywhere. :)~~
- [x] Default setting for `accept-encoding`. (Is there a specification that cannot be overridden by a `header` method when `Request::builder()` is executed? Also, what about the encoding priority order?)

### Checklist

### DIFF
--- a/llrt_core/src/modules/http/response.rs
+++ b/llrt_core/src/modules/http/response.rs
@@ -1,6 +1,6 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
-use std::time::Instant;
+use std::{collections::HashMap, io::Read, time::Instant};
 
 use http_body_util::BodyExt;
 use hyper::{body::Incoming, header::HeaderName};
@@ -20,7 +20,10 @@ use crate::{
 use super::{blob::Blob, headers::Headers};
 
 use once_cell::sync::Lazy;
-use std::collections::HashMap;
+
+use brotlic::DecompressorReader as BrotliDecoder;
+use flate2::read::{GzDecoder, ZlibDecoder};
+use zstd::stream::read::Decoder as ZstdDecoder;
 
 static STATUS_TEXTS: Lazy<HashMap<u16, &'static str>> = Lazy::new(|| {
     let mut map = HashMap::new();
@@ -115,10 +118,24 @@ impl<'js> Response<'js> {
             }
         }
 
+        let mut content_encoding = None;
+        if let Some(content_encoding_header) =
+            response_headers.get(HeaderName::from_static("content-encoding"))
+        {
+            if let Ok(content_encoding_header) = content_encoding_header.to_str() {
+                content_encoding = Some(content_encoding_header.to_owned())
+            }
+        }
+
         let headers = Headers::from_http_headers(response.headers())?;
         let headers = Class::instance(ctx.clone(), headers)?;
 
         let status = response.status();
+
+        let body_attributes = BodyAttributes {
+            content_type,
+            content_encoding,
+        };
 
         Ok(Self {
             body: Some(BodyVariant::Incoming(Some(response))),
@@ -129,7 +146,7 @@ impl<'js> Response<'js> {
             status_text: None,
             redirected,
             headers,
-            content_type,
+            body_attributes,
             abort_receiver,
         })
     }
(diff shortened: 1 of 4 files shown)


llrt_core/src/modules/http/response.rs
```diff
@@ -149,7 +166,21 @@ impl<'js> Response<'js> {
                     body.body_mut().collect().await.or_throw(ctx)?.to_bytes()
                 };
 
-                bytes.to_vec()
+                if self.body_attributes.content_encoding.is_none() {
+                    return Ok(Some(bytes.to_vec()));
+                }
+
+                let content_encoding = self.body_attributes.content_encoding.clone();
+                let mut data: Vec<u8> = Vec::new();
```
richarddavison, line 174: How will the decoders below allocate? Do they reserve based on max size or does it append the the vec? Would it make sense to set `Vec::with_capactity` to at least since of bytes to reduce allocations?

llrt_core/src/modules/http/response.rs
```diff
@@ -149,7 +166,21 @@ impl<'js> Response<'js> {
                     body.body_mut().collect().await.or_throw(ctx)?.to_bytes()
                 };
 
-                bytes.to_vec()
+                if self.body_attributes.content_encoding.is_none() {
```
richarddavison, line 169: We can do the other way around and do `if let Some(encoding) =  self.body_attributes.content_encoding {...}` which would avoid the unwrap below.

richarddavison (changes requested):
This is fantastic, thanks! Simple and clean implementation. Now we just need some tests. I suggest to encode (or hardcode) gzip, deflate, br and zstd "hello world" bytes and read the response as plain texts. Ideally test in Rust and not JS so we can mock the http responses.

llrt_core/src/modules/http/response.rs
nabetti1720: Thank you for pointing this out! 
It would be a more prospective code. I will present it in the next commit.

llrt_core/src/modules/http/response.rs
nabetti1720: I was wondering what to do about pre-allocating vectors, too.

Checking the `std::io::read_to_end()` implementation (actually `std::io::default_read_to_end()`), it looks like it does `Vec::try_reserve()` by 32 bytes if there is insufficient capacity. So, as you say, it would be effective to allocate in advance.

- std::io::read_to_end()`:  https://doc.rust-lang.org/src/std/io/mod.rs.html#851-854
- std::io::default_read_to_end()`:  https://doc.rust-lang.org/src/std/io/mod.rs.html#410-523

The best thing to do would have been to reserve the capacity in advance with `content-length`, but the response returned in chunked format is not given `content-length`, and it does not seem to be possible to know the just size in advance.

An idea to minimize the additional allocation as much as possible is to reserve "roughly twice" the size of the response body in advance, for example, and `Vec::shrink_to_fit()` it later.

What do you think?

nabetti1720:
> I suggest to encode (or hardcode) gzip, deflate, br and zstd "hello world" bytes and read the response as plain texts. Ideally test in Rust and not JS so we can mock the http responses.

Yes, I added the test that way.
One of the remaining ToDo, checking deflate decompression on the actual website, can be substituted for this test if it seems difficult.

The only remaining issue is the default `accept-encoding`!

richarddavison:
> Yes, I added the test that way. One of the remaining ToDo, checking deflate decompression on the actual website, can be substituted for this test if it seems difficult.

It's not ideal to have CI depend on external resources. These tests should suffice.

nabetti1720:
> It's not ideal to have CI depend on external resources. These tests should suffice.

Sorry for the misunderstanding.
We didn't intend to incorporate it into CI testing, we just wanted to see how it worked on a real website. :)

richarddavison:
@nabetti1720 please rebase from main. I have a mentor changed CI

nabetti1720:
> @nabetti1720 please rebase from main. I have a minor changed CI

I ran it as soon as possible.

richarddavison (approved):

llrt_core/src/modules/http/response.rs
richarddavison: I think we can start to reserve at least the size of the compressed bytes

nabetti1720: I will go with that idea.

nabetti1720:
All issues were addressed!

PR #395, opened 2024-05-31
feat: Improved compliance of crypto object

nabetti1720:
### Description of changes

Add the following methods to improve the compliance of `crypto` Object.

- crypto.getRandomValues()

### Checklist

### DIFF
--- a/llrt_core/src/modules/crypto/mod.rs
+++ b/llrt_core/src/modules/crypto/mod.rs
@@ -13,7 +13,7 @@ use rquickjs::{
     function::{Constructor, Opt},
     module::{Declarations, Exports, ModuleDef},
     prelude::{Func, Rest},
-    Class, Ctx, Error, Exception, Function, IntoJs, Null, Object, Result, TypedArray, Value,
+    Class, Ctx, Error, Exception, Function, IntoJs, Null, Object, Result, Value,
 };
 
 use crate::{
@@ -116,7 +116,7 @@ fn random_fill_sync<'js>(
 ) -> Result<Object<'js>> {
     let offset = offset.unwrap_or(0);
 
-    if let Some(array_buffer) = obj_to_array_buffer(&ctx, &obj)? {
+    if let Some(array_buffer) = obj_to_array_buffer(&obj)? {
         let checked_len = get_checked_len(array_buffer.len(), size.0, offset);
 
         let raw = array_buffer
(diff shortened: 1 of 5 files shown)


llrt_core/src/modules/crypto/mod.rs
```diff
@@ -132,6 +133,49 @@ fn random_fill_sync<'js>(
...
+            let val_bytes = $rng.gen::<$ty>().to_ne_bytes();
+            chunk.copy_from_slice(&val_bytes[..size]);
+        }
+    }};
+}
+
+fn get_random_values<'js>(ctx: Ctx<'js>, obj: Object<'js>) -> Result<Object<'js>> {
+    let mut rng = rand::thread_rng();
+
+    if let Some(array_buffer) = obj_to_array_buffer(&ctx, &obj)? {
```
richarddavison, line 149: I would suggest to refactor this method:
https://github.com/nabetti1720/llrt/blob/724f74f198fb93f3f674810d1728e203a4e8f75c/llrt_core/src/utils/object.rs#L142 to avoid matching on class names and instead to `if let Ok(array) = TypedArray::<i8>::from_object(obj.clone())` for each type and return the buffer.

Then use the same principle below.

llrt_core/src/modules/crypto/mod.rs
```diff
@@ -132,6 +133,49 @@ fn random_fill_sync<'js>(
     Ok(obj)
 }
 
+macro_rules! fill_typed_array {
+    ($ty:ty, $bytes:expr, $rng:expr) => {{
+        let size = mem::size_of::<$ty>();
+        for chunk in $bytes.chunks_exact_mut(size) {
```
richarddavison, line 139: Why do we need to chunk? Can't we fill an entire buffer?

richarddavison (changes requested):
Looking great so far, thanks for the PR. Just some questions :)

llrt_core/src/modules/crypto/mod.rs
nabetti1720: Are you suggesting that we fill the entire length of the array with random numbers, one byte at a time?

Each array buffer type has a primitive type size for each element, and I thought that if I filled a completely random number, it might end up with a value that could not exist in that type.

I tried that approach in this effort (it may have been an issue with my implementation...), it sometimes resulted in out-of-range asserts in the tests.

llrt_core/src/modules/crypto/mod.rs
nabetti1720: Is the correction as expected? Please review. :)

llrt_core/src/modules/crypto/mod.rs
richarddavison: I think we can use something like this no?

https://rust-random.github.io/rand/rand/trait.RngCore.html#tymethod.fill_bytes

tests/unit/crypto.test.ts
```diff
@@ -140,4 +140,101 @@ describe("random", () => {
...
+      expect(byte >= 0 && byte <= 4294967295).toBeTruthy();
+    }
+  });
+  it("should generate random bytes synchronously into a Float32Array using getRandomValues", () => {
+    const float32Array = new Float32Array(10);
+    defaultImport.getRandomValues(float32Array);
+    expect(float32Array.length).toEqual(10);
+    for (const byte of float32Array) {
+      expect(
+        (byte >= -3.4028234663852886e38 && byte <= 3.4028234663852886e38) ||
```
richarddavison, line 212: Can we use hex notation instead for all tests?
For exemple test between 7F7FFFFFF to FF7FFFFFF

richarddavison (changes requested):

llrt_core/src/modules/crypto/mod.rs
richarddavison: And since no true random is needed we can use the global inited SYSTEM_RANDOM variable

tests/unit/crypto.test.ts
nabetti1720: It looks like there are 9bytes, is this comparison correct for the 36bits range?

`Float32Array` is generated from `rng.fill_bytes()`, but neither of the following two conditions pass.
Note that the original implementation no longer passes the test.

```javascript
    for (const byte of float32Array) {
      expect((byte >= 0x7F7FFFFFF && byte <= 0xFF7FFFFFF)).toBeTruthy();
    }

```
```javascript
    for (const byte of float32Array) {
      expect((byte >= 0x7F7FFFFF && byte <= 0xFF7FFFFF)).toBeTruthy();
    }
```

To begin with, can IEEE 754 be expressed in hexadecimal?

nabetti1720: For tests of `Int...` and `Uint...` types, we were able to replace all of them with hexadecimal representations.

For the `Float...` types, I replaced it with a hexadecimal representation (e.g. Float32Array as above), but the test now fails.

llrt_core/src/modules/crypto/mod.rs
nabetti1720: > And since no true random is needed we can use the global inited SYSTEM_RANDOM variable

I do not understand how using SYSTEM_RANDOM would solve this problem...

nabetti1720:
> I might be wrong but I think it’s the view (the type array) that determines how the underlying array buffer is behaving. So it’s always a u8 array of numbers but viewed as a for example Float64Array those u8s would represent 64 bit floats instead. But it might overflow for floats in IEEE format

For this very reason, the specification does not seem to support `Float32Array` and `Float64Array`.

I asked ChatGPT and they suggested the following workaround in Javascript. :)

```Javascript
function getRandomFloat32Array(length) {
    var buffer = new ArrayBuffer(length * Float32Array.BYTES_PER_ELEMENT);
    var floatArray = new Float32Array(buffer);

    var uintArray = new Uint32Array(buffer);
    crypto.getRandomValues(uintArray);

    for (var i = 0; i < length; i++) {
        // Convert 32-bit unsigned integer to float between 0 and 1
        floatArray[i] = uintArray[i] / (0xFFFFFFFF + 1);
    }

    return floatArray;
}

// Example usage
var floatArray = getRandomFloat32Array(10);
console.log(floatArray);
```

Since we cannot find a good solution at this time, we would like to exclude `Float32Array` and `Float64Array` from support.

llrt_core/src/modules/crypto/mod.rs
richarddavison: It won’t solve this problem just another perf suggestion. We can skip Float Typed Arrays and fill buffers using SYSTEM_RANDOM

richarddavison:
I think we can go ahead and simply exclude support for float typed arrays

nabetti1720:
Thank you for your consideration.
It has already been completed in the last [commit](https://github.com/awslabs/llrt/pull/395/commits/284d12f658e593f11fa3c2a9cb850140512ddfe0).

llrt_core/src/modules/crypto/mod.rs
nabetti1720: Hi @richarddavison , I have responded, please review.

richarddavison (approved):

PR #412, opened 2024-06-07
fix: get method does not return null

nabetti1720:
### Description of changes

In the `get()` method of the `Headers` object, `undefined` is returned when a non-existent key is specified.

```shell
% ./llrt -e "console.log(new Headers().get('aaa'));"
undefined
```
Normally, `null` should be returned.

https://developer.mozilla.org/en-US/docs/Web/API/Headers/get
> The get() method of the [Headers](https://developer.mozilla.org/en-US/docs/Web/API/Headers) interface returns a byte string of all the values of a header within a Headers object with a given name. If the requested header doesn't exist in the Headers object, it returns null.

### Checklist

### DIFF
--- a/llrt_core/src/modules/http/headers.rs
+++ b/llrt_core/src/modules/http/headers.rs
@@ -4,8 +4,8 @@ use std::collections::BTreeMap;
 
 use hyper::HeaderMap;
 use rquickjs::{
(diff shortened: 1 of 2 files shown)


richarddavison (approved):
LGTM!

PR #418, opened 2024-06-10
feat: Improved compliance of Request object

nabetti1720:
### Description of changes

Add the following method to improve the compliance of `Request` object.

- Request.blob()

### Checklist

### DIFF
--- a/llrt_core/src/modules/http/request.rs
+++ b/llrt_core/src/modules/http/request.rs
@@ -158,10 +158,13 @@ impl<'js> Request<'js> {
 
     async fn blob(&mut self, ctx: Ctx<'js>) -> Result<Blob> {
         let headers = Headers::from_value(&ctx, self.headers().unwrap().as_value().clone())?;
-        let mime_type = headers
-            .iter()
-            .find(|(k, _)| k == &"content-type")
-            .map(|(_, v)| v.clone());
+        let mime_type = headers.iter().find_map(|(k, v)| {
+            if k == "content-type" {
+                Some(v.clone())
+            } else {
+                None
+            }
+        });
         if let Some(bytes) = self.take_bytes(&ctx).await? {
             return Ok(Blob::from_bytes(bytes, mime_type));
         }
--- a/llrt_core/src/modules/http/request.rs
+++ b/llrt_core/src/modules/http/request.rs
@@ -37,6 +37,7 @@ pub struct Request<'js> {
     method: String,
     headers: Option<Class<'js, Headers>>,
     body: Option<Value<'js>>,
+    content_type: Option<String>,
     signal: Option<Class<'js, AbortSignal<'js>>>,
 }
 
@@ -60,6 +61,7 @@ impl<'js> Request<'js> {
             method: "GET".to_string(),
             headers: None,
             body: None,
+            content_type: None,
             signal: None,
         };
 
@@ -156,6 +158,13 @@ impl<'js> Request<'js> {
         TypedArray::new(ctx, Vec::<u8>::new()).map(|m| m.into_value())
     }
 
+    async fn blob(&mut self, ctx: Ctx<'js>) -> Result<Blob> {
+        if let Some(bytes) = self.take_bytes(&ctx).await? {
+            return Ok(Blob::from_bytes(bytes, self.content_type.clone()));
+        }
+        Ok(Blob::from_bytes(Vec::<u8>::new(), None))
+    }
+
     fn clone(&mut self, ctx: Ctx<'js>) -> Result<Self> {
         let headers = if let Some(headers) = &self.headers {
             Some(Class::<Headers>::instance(
@@ -171,6 +180,7 @@ impl<'js> Request<'js> {
             method: self.url.clone(),
             headers,
             body: self.body.clone(),
+            content_type: self.content_type.clone(),
             signal: self.signal.clone(),
         })
     }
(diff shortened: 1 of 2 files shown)


llrt_core/src/modules/http/request.rs
```diff
@@ -156,6 +158,13 @@ impl<'js> Request<'js> {
         TypedArray::new(ctx, Vec::<u8>::new()).map(|m| m.into_value())
     }
 
+    async fn blob(&mut self, ctx: Ctx<'js>) -> Result<Blob> {
+        if let Some(bytes) = self.take_bytes(&ctx).await? {
+            return Ok(Blob::from_bytes(bytes, self.content_type.clone()));
```
richarddavison, line 163: use content-type value from header if this is missing

richarddavison (changes requested):
Some minor changes and we're good 👍

llrt_core/src/modules/http/request.rs
```diff
@@ -60,6 +61,7 @@ impl<'js> Request<'js> {
             method: "GET".to_string(),
             headers: None,
             body: None,
+            content_type: None,
```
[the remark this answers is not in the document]
nabetti1720: Thanks for the review.
Your comment was difficult for me to understand(All this is due to my lack of English skills...), so I modified it as follows, while also checking the behavior of Bun.

1. If both blob's type and Headers' content-type are specified, blob's type is used.
2. If blob's type is not specified and Headers' content-type is specified, Headers' content-type is used.
3. If both are not specified, nothing is specified.

llrt_core/src/modules/http/request.rs
nabetti1720: I think I have addressed this by modifying `assign_request()`.

llrt_core/src/modules/http/request.rs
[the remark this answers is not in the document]
nabetti1720: Sorry. I forgot one more thing...

4. if the type of Blob is specified and the content-type of Headers is not specified, the content-type of Headers should be the type of Blob.

Have you been told about this?

llrt_core/src/modules/http/request.rs
```diff
@@ -207,18 +220,33 @@ fn assign_request<'js>(request: &mut Request<'js>, ctx: Ctx<'js>, obj: &Object<'
...
+                        Some(TypedArray::<u8>::new(ctx.clone(), blob.get_bytes())?.into_value());
+                    request.content_type = match blob.mime_type().len() {
+                        0 => None,
+                        _ => Some(blob.mime_type()),
+                    };
+                },
+                None => {
+                    request.body = Some(body);
+                    request.content_type = None;
+                },
```
richarddavison, line 236: This can be implemented like this instead:
```rust
if let Some(blob) = get_class::<Blob>(&body)? {
    let blob = blob.borrow();
    request.body = Some(TypedArray::<u8>::new(ctx.clone(), blob.get_bytes())?.into_value());
    request.content_type = if blob.mime_type().is_empty() {
        None
    } else {
        Some(blob.mime_type())
    };
} else {
    request.body = Some(body);
    request.content_type = None;
}
```

llrt_core/src/modules/http/request.rs
```diff
@@ -207,18 +220,33 @@ fn assign_request<'js>(request: &mut Request<'js>, ctx: Ctx<'js>, obj: &Object<'
...
             }
         }
     }
 
     if obj.contains_key("headers").unwrap() {
```
richarddavison, line 241: If think we can clean up `assign_request` class by avoiding unwraps and use:
```rust
 if let Some(headers) = obj.get_optional("headers"){
        
 }
```
Similarly for "body" and "signal"

llrt_core/src/modules/http/request.rs
```diff
@@ -207,18 +220,33 @@ fn assign_request<'js>(request: &mut Request<'js>, ctx: Ctx<'js>, obj: &Object<'
...
         let headers: Value = obj.get("headers")?;
         let headers = Headers::from_value(&ctx, headers)?;
+        if request.content_type.is_none() {
```
richarddavison, line 244: This is unnecessary since we only need to set `request.content_type` if blob is being used as body or if `blob()` method is called. It's thus better to calculate this from the blob method only. In this implementation we'll loop through all headers even if we don't ever use `blob()` method.

llrt_core/src/modules/http/request.rs
richarddavison: Yes and no. It's better to use "on-demand" than eagerly. See my comment :)

llrt_core/src/modules/http/request.rs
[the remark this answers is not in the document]
richarddavison: 1. No, it looks like the header value always sets the blobs value. If no is specified, it's just an empty type (even if blob as a type defined. Test this in Chrome:
```javascript
let b = new Blob(["test"],{type:"application/json"})
let r = new Request(b,{headers: {"content-type":"text/plain"}})

await r.blob();

let b = new Blob(["test"],{type:"application/json"})
let r = new Request(b)

await r.blob();

```
So if content-type header is present, it will be used when calling `blob()` otherwise blob will have empty type (even if specified when constructed)

This basically means that we don't need a `content_type` field, it's never used or derived from blob instance

2. See 1
3. Yes
4. No, headers are not touched.

richarddavison (changes requested):

llrt_core/src/modules/http/request.rs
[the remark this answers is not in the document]
nabetti1720: Indeed!
I see that the blob’s type is not used in both of the following patterns.

1.
image

2.
image

This way, there is no need to have a separate content-type inside the request object.
So we can just set the content-type of headers to the type used in blob(), regardless of whether it has a value or not.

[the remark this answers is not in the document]
nabetti1720: I was wondering if it would be possible to simplify the method of retrieving the content-type a bit more, but I didn't have any good ideas. I have made the fix and would appreciate a review.

llrt_core/src/modules/http/request.rs
nabetti1720: I have decided to refer to the content-type of Headers directly in Blob(), so this support is no longer necessary.

llrt_core/src/modules/http/request.rs
nabetti1720: I have decided to refer to the content-type of Headers directly in Blob(), so this support is no longer necessary.

llrt_core/src/modules/http/request.rs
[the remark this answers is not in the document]
richarddavison: The only other way I can think of is this:
```rust
let mime_type = headers.iter()
    .find_map(|(k, v)| if k == "content-type" { Some(v.clone()) } else { None });
```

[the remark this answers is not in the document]
nabetti1720: Thank you very much. I think I have now addressed all the issues.

richarddavison (approved):

PR #431, opened 2024-06-22
feat: Improved compliance of TextEncoder

nabetti1720:
### Description of changes

Add the following methods to improve the compliance of Encoding API.

- TextEncoder.encodeInto()

### Checklist

### DIFF
--- a/llrt_core/src/modules/encoding/text_encoder.rs
+++ b/llrt_core/src/modules/encoding/text_encoder.rs
@@ -3,7 +3,6 @@
 use rquickjs::{Ctx, Object, Result, TypedArray, Value};
 
 use crate::utils::{object::obj_to_array_buffer, result::ResultExt};
-use std::str::from_utf8;
 
 #[derive(rquickjs::class::Trace)]
 #[rquickjs::class]
@@ -46,7 +45,9 @@ impl TextEncoder {
 
             let mut enc = encoding_rs::UTF_8.new_encoder();
             (_, _, written, _) = enc.encode_from_utf8(string.as_str(), bytes, false);
-            read = from_utf8(&bytes[..written]).unwrap().encode_utf16().count();
+            read = string[..written]
+                .chars()
+                .fold(0, |acc, ch| acc + ch.len_utf16());
         }
 
         let obj = Object::new(ctx)?;
--- a/llrt_core/src/modules/encoding/text_encoder.rs
+++ b/llrt_core/src/modules/encoding/text_encoder.rs
@@ -1,12 +1,15 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
-use rquickjs::{Ctx, Result, TypedArray, Value};
+use rquickjs::{Ctx, Object, Result, TypedArray, Value};
+
+use crate::utils::{object::obj_to_array_buffer, result::ResultExt};
+use std::str::from_utf8;
 
 #[derive(rquickjs::class::Trace)]
 #[rquickjs::class]
 pub struct TextEncoder {}
 
-#[rquickjs::methods]
+#[rquickjs::methods(rename_all = "camelCase")]
 impl TextEncoder {
     #[qjs(constructor)]
     pub fn new() -> Self {
@@ -20,5 +23,35 @@ impl TextEncoder {
 
     pub fn encode<'js>(&self, ctx: Ctx<'js>, string: String) -> Result<Value<'js>> {
         TypedArray::new(ctx, string.as_bytes()).map(|m| m.into_value())
+    }
+
+    pub fn encode_into<'js>(
+        &self,
+        ctx: Ctx<'js>,
+        string: String,
+        obj: Object<'js>,
+    ) -> Result<Object<'js>> {
+        let mut read = 0;
+        let mut written = 0;
+
+        if let Some((array_buffer, source_length, source_offset)) = obj_to_array_buffer(&obj)? {
+            let raw = array_buffer
+                .as_raw()
+                .ok_or("ArrayBuffer is detached")
+                .or_throw(&ctx)?;
+
+            let bytes = unsafe {
+                std::slice::from_raw_parts_mut(raw.ptr.as_ptr().add(source_offset), source_length)
+            };
+
+            let mut enc = encoding_rs::UTF_8.new_encoder();
+            (_, _, written, _) = enc.encode_from_utf8(string.as_str(), bytes, false);
+            read = from_utf8(&bytes[..written]).unwrap().encode_utf16().count();
+        }
+
+        let obj = Object::new(ctx)?;
+        obj.set("read", read)?;
+        obj.set("written", written)?;
+        Ok(obj)
     }
 }
--- a/tests/unit/encoding.test.ts
+++ b/tests/unit/encoding.test.ts
@@ -28,7 +28,7 @@ describe("atoa & btoa", () => {
 });
 
 describe("TextDecoder", () => {
-  it("Should be able to decode even non UTF-8 labels ", () => {
+  it("should be able to decode even non UTF-8 labels", () => {
     const hono = "炎"; // hono - [炎] means flame🔥 in Japanese
     const honoSjis = new Uint8Array([0x89, 0x8a]);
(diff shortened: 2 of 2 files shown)


llrt_core/src/modules/encoding/text_encoder.rs
```diff
@@ -21,4 +24,34 @@ impl TextEncoder {
...
+        &self,
+        ctx: Ctx<'js>,
+        string: String,
+        obj: Object<'js>,
+    ) -> Result<Object<'js>> {
+        let mut read = 0;
+        let mut written = 0;
+
+        if let Some((array_buffer, source_length, source_offset)) = obj_to_array_buffer(&obj)? {
+            let raw = array_buffer
```
richarddavison, line 38: I don't think you need to access the raw bytes like this.
You can do `let bytes: &[u8] = array_buffer.as_ref()` instead and use offsets on slice instead.

llrt_core/src/modules/encoding/text_encoder.rs
```diff
@@ -21,4 +24,34 @@ impl TextEncoder {
...
+                .ok_or("ArrayBuffer is detached")
+                .or_throw(&ctx)?;
+
+            let bytes = unsafe {
+                std::slice::from_raw_parts_mut(raw.ptr.as_ptr().add(source_offset), source_length)
+            };
+
+            let mut enc = encoding_rs::UTF_8.new_encoder();
+            (_, _, written, _) = enc.encode_from_utf8(string.as_str(), bytes, false);
+            read = from_utf8(&bytes[..written]).unwrap().encode_utf16().count();
```
richarddavison, line 49: This seems inefficient as we're encoding twice. Can we do encoding only once?

richarddavison (changes requested):
Thanks for the PR!!! Fantastic job, some minor changes and I'll be good to merge 🙌

llrt_core/src/modules/encoding/text_encoder.rs
nabetti1720: The type of bytes must be `&mut [u8]`, not `&[u8]`.
I could not find a way to create `&mut [u8]` other than the current implementation...

llrt_core/src/modules/encoding/text_encoder.rs
nabetti1720: Does "twice" mean `from_utf8()` and `encode_utf16()`?
If so, we cannot avoid it for the following reasons at the moment.

1. To correctly calculate the number of units in UTF-16, at this time we need to run `encode_utf16()`.
2. And `encode_utf16()` must be performed on `&str`.
3. In addition, `&bytes[. .written]` is `&[u8]`, we need to run `from_utf8()` to convert from `&[u8]` to `&str`.

I may be overlooking something that could be done more easily.

llrt_core/src/modules/encoding/text_encoder.rs
richarddavison: Ah right, makes sense since we're writing into the typed array!

llrt_core/src/modules/encoding/text_encoder.rs
richarddavison: Would something like this work? Arn't rust strings always utf8?
```rust
pub fn encode_into<'js>(
    &self,
    ctx: Ctx<'js>,
    string: String,
    obj: Object<'js>,
) -> Result<Object<'js>> {
    let mut read = 0;
    let mut written = 0;

    if let Some((array_buffer, source_length, source_offset)) = obj_to_array_buffer(&obj)? {
        let raw = array_buffer
            .as_raw()
            .ok_or("ArrayBuffer is detached")
            .or_throw(&ctx)?;
        let bytes = unsafe {
            std::slice::from_raw_parts_mut(raw.ptr.as_ptr().add(source_offset), source_length)
        };

    
       read = string.chars().fold(0, |acc, ch| acc + ch.len_utf16());

        let utf8_bytes = string.as_bytes();
        written = utf8_bytes.len().min(bytes.len());
        bytes[..written].copy_from_slice(&utf8_bytes[..written]);
    }

    let obj = Object::new(ctx)?;
    obj.set("read", read)?;
    obj.set("written", written)?;
    Ok(obj)
}
```

richarddavison (changes requested):

llrt_core/src/modules/encoding/text_encoder.rs
nabetti1720: I tried it immediately, and there are some things that worked and some things that did not work.

```rust
written = utf8_bytes.len().min(bytes.len());
```
`wrtten` must return a valid number of digits when rounded out by that number of digits as UTF-8. This means that the number could be even smaller than the two minimums.
This decision is not easy, as it requires consideration of multibyte characters and surrogate pairs.
Therefore, I had to rely on `decoding_rs::UTF_8::encode_from_utf8()` to make this determination.

```rust
read = string.chars().fold(0, |acc, ch| acc + ch.len_utf16());
```
Assuming that `written` can be calculated correctly, we were able to incorporate this implementation by cutting out the string by the number of bytes indicated by `written`.

richarddavison (approved):

richarddavison:
Hi @nabetti1720. I think we need to revisit this. The `encoding_rs` crate adds significant size to the executable due to the vast number of encodings it supports. Node only supports "utf8", "utf16le" and "latin1":

> 'utf8' (alias: 'utf-8'): Multi-byte encoded Unicode characters. Many web pages and other document formats use [UTF-8](https://en.wikipedia.org/wiki/UTF-8). This is the default character encoding. When decoding a Buffer into a string that does not exclusively contain valid UTF-8 data, the Unicode replacement character U+FFFD � will be used to represent those errors.
> 
> 'utf16le' (alias: 'utf-16le'): Multi-byte encoded Unicode characters. Unlike 'utf8', each character in the string will be encoded using either 2 or 4 bytes. Node.js only supports the [little-endian](https://en.wikipedia.org/wiki/Endianness) variant of [UTF-16](https://en.wikipedia.org/wiki/UTF-16).
> 
> 'latin1': Latin-1 stands for [ISO-8859-1](https://en.wikipedia.org/wiki/ISO-8859-1). This character encoding only supports the Unicode characters from U+0000 to U+00FF. Each character is encoded using a single byte. Characters that do not fit into that range are truncated and will be mapped to characters in that range.

Is there a way to limit or strip all other encodings from `encoding_rs`? Or, if not, add BOM and utf16le support to my initial implementation (with reasonable complexity and good performance)?

nabetti1720:
Hi @richarddavison , thank you for your comment. Let me clear up a few things.

1. I found this quote in the section on code conversion in `Buffer`.
https://nodejs.org/api/buffer.html#buffers-and-character-encodings
2. However, the `Util.TextEncoder` and `Util.TextDecoder` in node.js (and that provided in globalThis) follows WebStandard.
https://nodejs.org/api/util.html#class-utiltextdecoder
https://nodejs.org/api/util.html#class-utiltextencoder
3. And indeed, in LLRT, I believe the code conversion logic of `Buffer` and `Text[En|De]coder` were common in previous implementations.
4. However, in this PR, `encoding_rs` was adopted because of the motivation to adapt `Text[En|De]coder` to WebStandard, but the `Buffer` implementation was left as is and should not be affected. (`Buffer` is an important component and we did not want to influence it.)

Based on the above, we do not believe that there is currently any difference between node.js and LLRT in terms of supported encoding policies.

If you understand these and still want to narrow down the encoding target:
- `encoding_rs` does not seem to have any mechanism to control it, like a function flag. However, if done well, it may be possible to replace it with encoding_rs::mem.
https://docs.rs/encoding_rs/latest/encoding_rs/mem/index.html
- Or as you say, revert to the previous implementation and add minimal functionality.

Is that an answer?

richarddavison:
I see you point but I think we should revert to previous implementation but add support for utf16le. Full ICU could be added as a feature gated implementation in the future if we see users asking for it. The reason is that full ICU and encoding support contributes significantly to startup time 🙂

nabetti1720:
Hi @richarddavison , I misjudged your performance and your passion for footprint reduction. :)
After this, I will issue a PR to revert back to the implementation before the introduction of `encoding_rs`. (But instead of cramming it into mod.rs as before, we will keep what we can, such as the current source structure and methods/properties).

PR #457, opened 2024-07-09
perf: replacement of @smithy/abort-controller

nabetti1720:
### Description of changes

Many `@aws-sdk` packages directly or indirectly depend on `@smithy/abort-controller`.
But if you check the implementation of `@smithy/abort-controller`, `AbortController` and `AbortSignal` are just defined.

https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/Package/-smithy-abort-controller/
https://github.com/smithy-lang/smithy-typescript/tree/main/packages/abort-controller/src

LLRT already provides these classes as native objects of Rust, so there is no need for them.
This PR replaces the `@smithy/abort-controller` import statement with an empty javascript file so that the native implementation can be used.
We also believe that replacing with native implementation will lead to some performance improvement.

### Checklist


richarddavison (approved):
That’s great. Thank you!

Issue #430, opened 2024-06-19
Is it possible to publish common modules as third-party packages, such as rquickjs-ext similar to deno?


ahaoboy:
Some common modules such as os, path, fs, etc., which are not closely coupled with the runtime, can they be extracted as separate third-party libraries? For example, rquickjs-ext, so that other projects that only use rquickjs can also use it

richarddavison:
It probably can! We "semi-started" to support this by moving out runtime specific code to its own crate. Right now there are a couple of blockers tho:

1 There is some lambda specific features (currently gated) that reside in `llrt_core`
2 The library depends on some JS bundled code to be present (for things like ESM), this should be moved and built into `core`
3 The core module will at compile time create a perfect hashmap and embed modules into its own executable.

I experimented with a different approach for 3 by appending a special data format that contained an index of all embedded files + the raw bytecode at the end of the executable. This proved to slightly increase cold start even though it made the embedding more flexible so was ruled out.

Sytten:
Hey I am pitching in since I started working on a utils crate for rquickjs (https://github.com/DelSkayn/rquickjs/pull/319). The goal is to eventually recreate all the api we expect for node/deno.

I would interested to help in whatever implementation like fetch.

richarddavison:
> Hey I am pitching in since I started working on a utils crate for rquickjs ([DelSkayn/rquickjs#319](https://github.com/DelSkayn/rquickjs/pull/319)). The goal is to eventually recreate all the api we expect from node/deno for quickjs.
> 

For sure! ⭐️

> I would interested to help in whatever implementation. @richarddavison do you guys have a chat system so we can talk about that?

I just dropped you an email!

gc-victor:
It would be great to know if there is any update about this 🙏

ahaoboy:
Some problems I encountered during the development process, I hope the new module mechanism can improve
1. Node protocol support, that is, support for two module protocols at the same time 
```js
import fs from 'fs'

import fs from 'node:fs'
```
 
2. Module dependency, for example, fetch depends on the net module. Currently, we can only manually ensure the import order of the modules. Is it possible to implement a mechanism that reports an error or automatically imports the net module if only fetch is imported?

richarddavison:
> Some problems I encountered during the development process, I hope the new module mechanism can improve
> 
> 1. Node protocol support, that is, support for two module protocols at the same time
> 
> ```js
> import fs from 'fs'
> 
> import fs from 'node:fs'
> ```
[...]

PR #458, opened 2024-07-09
Split module

Sytten:
### Issue

Related to #430

### Description of changes

The goal is to split off modules one by one so they can be re-used by other users of rquickjs.
It is easier to start with a fresh crate, I hesitated to name it `rquickjs-extra` since it would be nice if it could be discoverable.
I used feature flags to make sure the user can choose what they need.

* Splitting the FS, Buffer and Path modules (since they are a unit of work)
* Split the utils in its own crate, it was cleaner that way for `llrt_core` to reuse it but open to other ideas
* I kept the imports as-is in `llrt_core` as much as possible to reduce code changes
* I decided to split at the `ModuleInfo` level and let the user handle the registration in their context. Idea is to keep `llrt_modules` as "engine" agnostic as possible.

I can also craft a github action to publish to crates.io once we get through a first round of review.
I also want to add CI tests for each feature flag to ensure they are really independent.
Finally, I want to start typescript typing for llrt similar to `@types/node` (this will live in the `DefinitelyTyped`)

If we agree on the structure, I will spend the coming weeks porting the other modules and making the necessary changes in rquickjs (thinking about the `setInterval` case which will require callbacks in the event loop of `rquickjs`).

### Checklist


richarddavison (approved):
Thanks! LGTM!

Issue #437, opened 2024-06-27
Make URL and URLSearchParams more compliant with WHATWG URL specification


stephencroberts:
In trying to use the `URL` module, I quickly found that it's not compliant with the [URL standard](https://url.spec.whatwg.org/) for many normal use-cases. The first issue I came across was with the handling of the `#` for hashes (fragments) in both getting and setting the hash, so I decided to set up the [web-platform-tests for url](https://github.com/web-platform-tests/wpt/tree/b634532883c8dd54ac7ef4a29b5eccaf5125359f/url) as part of the LLRT unit test suite which uncovered a host of other issues for common web use-cases. I have a fork of LLRT that I'm using, and I've updated `URL` and `URLSearchParams` to pass a vast majority of the tests (the rest need to be fixed in the `Url` crate). Are you open to a Pull Request?

Sytten:
I also created another variant of URLSearchParams here for reference https://github.com/DelSkayn/rquickjs/blob/bf699539bab3caacc23cadb3abbde9f1ae95160f/util/src/url_search_params.rs

PR #461, opened 2024-07-10
chore: delete test code that is no longer needed

nabetti1720:
### Issue # (if available)

Related #437, #447

### Description of changes

The following tests have been replaced by the more sophisticated [web-platform-tests](https://github.com/web-platform-tests/wpt) .

- URL class
- URLsearchparams class

This PR removes test code that is duplicated and no longer needed.
If it is found that a test is missing, it should be reflected in the [web-platform-tests](https://github.com/web-platform-tests/wpt) in the future.

Note that we left the test as a module because we thought it was necessary as a minimum.

### Checklist


richarddavison:
Thanks for the PR. I'm a bit hesitant to remove these tests. Are all tests really duplicated? 🤔

nabetti1720:
Hi @richarddavison , thank you for your comment.
We did not exactly check to see if everything was duplicated. It would take some time to investigate that. It is also not reasonable to verify duplicates multiple times as web-platform-tests move forward.

I have reconsidered this matter.

Even if the ultimate goal is to pass web-platform-tests, it is still useful when creating test cases at the same time as implementation, and it has the advantage of squashing bugs before merging web-platform-tests.

With that, I would like to close this PR.

PR #462, opened 2024-07-10
feat(llrt_core): update encoding to pass more web-platform-tests

stephencroberts:
### Description of changes

Adds a bunch of web-platform-tests for `TextEncoder` and `TextDecoder` and a few tweaks to those classes to pass more tests.

#### Reviewing

The first commit is simply adding the web-platform-tests from the wpt repo, so you can essentially skip over it. The second commit is the interesting part.

### Checklist


stephencroberts:
Putting in draft because I'm trying to use it in our wasm runtime and having issues.

llrt_core/src/modules/encoding/text_decoder.rs
```diff
@@ -64,24 +63,31 @@ impl<'js> TextDecoder {
...
 
     pub fn decode(&self, ctx: Ctx<'js>, buffer: Value<'js>) -> Result<String> {
-        let bytes = get_bytes(&ctx, buffer)?;
-
-        let decoder = Encoding::for_label(self.encoding.as_bytes()).unwrap();
+        let bytes = if buffer.is_undefined() {
+            vec![]
+        } else {
+            get_bytes(&ctx, buffer)?
+        };
```
richarddavison, line 70: Should this be added into `get_bytes`? Is it expected behavior for other places using bytes from a buffer, typed array or or string?

llrt_core/src/modules/encoding/text_decoder.rs
```diff
@@ -64,24 +63,31 @@ impl<'js> TextDecoder {
...
-                false => decoder.decode_with_bom_removal(&bytes),
-                true => decoder.decode_without_bom_handling(&bytes),
-            }
-        } else {
-            (str, _, has_error) = decoder.decode(&bytes);
-        }
+        let (str, has_error) = match self.ignore_bom {
+            false => decoder.decode_with_bom_removal(&bytes),
+            true => decoder.decode_without_bom_handling(&bytes),
+        };
```
richarddavison, line 80: A bit strange to match on bools (I know this was in previous version). I prefer a if:
```rust
let (str, has_error) = if self.ignore_bom {
    decoder.decode_without_bom_handling(&bytes)
} else {
    decoder.decode_with_bom_removal(&bytes)
};
```

llrt_core/src/modules/encoding/text_encoder.rs
```diff
@@ -16,43 +17,56 @@ impl TextEncoder {
     }
 
     #[qjs(get)]
-    fn encoding(&self) -> String {
-        "utf-8".to_string()
+    fn encoding(&self) -> &str {
```
richarddavison, line 20: Nice!

llrt_core/src/modules/encoding/text_encoder.rs
```diff
@@ -16,43 +17,56 @@ impl TextEncoder {
...
-    pub fn encode<'js>(&self, ctx: Ctx<'js>, string: String) -> Result<Value<'js>> {
-        TypedArray::new(ctx, string.as_bytes()).map(|m| m.into_value())
+    pub fn encode<'js>(&self, ctx: Ctx<'js>, string: Opt<Value<'js>>) -> Result<Value<'js>> {
+        if let Some(string) = string.0 {
+            if let Some(string) = string.as_string() {
+                if let Ok(string) = string.to_string() {
+                    return TypedArray::new(ctx, string.as_bytes()).map(|m| m.into_value());
+                }
+            }
+        }
```
richarddavison, line 31: Should we really ignore errors here? If string conversion fails for instance we?

llrt_core/src/modules/encoding/text_encoder.rs
```diff
@@ -16,43 +17,56 @@ impl TextEncoder {
...
             };
 
             let mut enc = encoding_rs::UTF_8.new_encoder();
-            (_, _, written, _) = enc.encode_from_utf8(string.as_str(), bytes, false);
-            read = string[..written]
+            let (_, _, written, _) = enc.encode_from_utf8(&src, dst, false);
+            let read = src[..written]
                 .chars()
                 .fold(0, |acc, ch| acc + ch.len_utf16());
-        }
```
richarddavison, line 51: Can we avoid iterating over each char to get the count?

llrt_utils/src/result.rs
```diff
@@ -6,6 +6,7 @@ use rquickjs::{Ctx, Exception, Result};
 
 pub trait ResultExt<T> {
     fn or_throw_msg(self, ctx: &Ctx, msg: &str) -> Result<T>;
+    fn or_throw_range(self, ctx: &Ctx, msg: &str) -> Result<T>;
```
richarddavison, line 9: Nice!

richarddavison (changes requested):
This is fantastic, thanks a lot! Some minor questions/changes 🙌

llrt_core/src/modules/encoding/text_decoder.rs
stephencroberts: Good question. I'll look where else it's used. Once we implement the `stream` option, I think this behavior changes even more, but I haven't read the spec yet; but there are wpt test cases around flushing the streaming buffer when `buffer` is `undefined` at the end of a stream.

llrt_core/src/modules/encoding/text_decoder.rs
stephencroberts: I could go either way, but I wrote a fair amount of Erlang recently where everything is a `match` so I'm kinda used to seeing it. 🙃

llrt_core/src/modules/encoding/text_encoder.rs
stephencroberts: Admittedly, I haven't read the spec yet. With the previous implementation, the `String` parameter required that the value be a string and would throw if something like `undefined` was passed in. With this change, we pass a lot more of the tests, but I can't say for sure how this is supposed to behave according to the spec. We could pause this change until I find out, but my thought was that it's just a slight improvement to what was already there. What do you think?

llrt_core/src/modules/encoding/text_encoder.rs
stephencroberts: So...I spent quite a lot of time playing with a lot of different approaches here, from using `qjs` primitives to get the string bytes, recast as `&[u16]`, `encode_from_utf16()` to get the `read` and `written` correctly, and to handle all of the test cases around unicode and surrogates (and unpaired surrogates), to attempting to encode UTF-8 with fallback to UTF-16, but I couldn't get anything to fully work. According to [MDN docs](https://developer.mozilla.org/en-US/docs/Web/API/TextEncoder/encodeInto#read), `read` is the number of UTF-16 code points, even though we're encoding into UTF-8. There are improvements to be made here, but I would need to thoroughly read the spec and understand how those edge cases are supposed to be handled.

llrt_core/src/modules/encoding/text_encoder.rs
stephencroberts: Thinking about this more, maybe we should `throw` to be safe. I can see if that causes test failures and look at what behavior those tests expect.

stephencroberts:
@richarddavison Thanks for the review! I think I've addressed your feedback, and I was able to get it working in our WASM runtime. It was a very interesting problem, and one that I had already come across in our runtime. If you use `map_or` or `ok_or` on a `Result` or `Option` to throw an exception, the argument is eagerly evaluated so it _always_ throws an exception into the JavaScript context, even if it was `Ok`/`Some`. We have a check in our runtime for uncaught exceptions, so it was always throwing even though it was decoding just fine. We have to use `map_or_else` and `ok_or_else` to lazily evaluate.

Let me know what else you find in the review!

llrt_core/src/modules/encoding/text_encoder.rs
stephencroberts: Oh, I forgot to mention, I think the issue with surrogates and such is that `encoding_rs` only accepts a `&str` for `encode_from_utf8` which requires turning a byte array into a string slice, and Rust requires a string slice to be valid UTF-8. To handle those edge cases, I believe `encoding_rs` should accept a byte array and check for surrogates and whatever other edge cases.

llrt_core/src/modules/encoding/text_encoder.rs
richarddavison: I was more concerned about this part:
```rust
if let Some(string) = string.as_string() {
    if let Ok(string) = string.to_string() {
        return TypedArray::new(ctx, string.as_bytes()).map(|m| m.into_value());
    }
}
```
That could be written as
```rust
if let Some(string) = string.as_string() {
    let string = string.to_string()?
    return TypedArray::new(ctx, string.as_bytes()).map(|m| m.into_value());
}
```

richarddavison:
> We have to use `map_or_else` and `ok_or_else` to lazily evaluate.

Oh ofc, I didn't think of that, thanks for finding. And thanks to extension traits this fix will be applied everywhere 🎉

llrt_core/src/modules/encoding/text_encoder.rs
nabetti1720: > `read` is the number of UTF-16 code points, even though we're encoding into UTF-8. 

When I first implemented this specification, it was the hardest thing for me to understand. No matter how much research I did, I could not find a use case for it, so I still don't understand it. :(

We may have to go back to the specification development process to understand it.

Apart from that, your refactor content confirms that I was not wrong to bring in `encoding_rs`. Thank you very much. :)

richarddavison:
For some strange reason http server test keeps failing on aarch64. Rerunning the pipeline again

stephencroberts:
> The built in test runner executes all tests in parallel (or 5 at the time)

Oh fascinating, it appeared the tests running locally were executed sequentially which is the only way the globals could work. Is that just a difference with how we run `make test` vs how we're executing in CI? One option would be for me to make further modifications to the WPT test harness to avoid the globals which is probably the _right_ way to do it but might cause some conflicts to resolve in the future with WPT. I tried to make the minimal changes to make things work.

llrt_core/src/modules/encoding/text_encoder.rs
stephencroberts: @nabetti1720 Thanks for reviewing! Yeah, I think you did a wonderful job. We have our own basic implementation of `TextEncoder/Decoder` and we also used `encoding_rs`. I'm looking forward to switching ours out for this one once we get this PR merged so we don't have to maintain it ourselves. Would be cool to implement the streaming features at some point too.

richarddavison:
> > The built in test runner executes all tests in parallel (or 5 at the time)
> 
> Oh fascinating, it appeared the tests running locally were executed sequentially which is the only way the globals could work. Is that just a difference with how we run `make test` vs how we're executing in CI? One option would be for me to make further modifications to the WPT test harness to avoid the globals which is probably the _right_ way to do it but might cause some conflicts to resolve in the future with WPT. I tried to make the minimal changes to make things work.

Thank you! The difference from CI is that it runs in an emulated environment (since GitHub actions are x86 only) which is much slower, causing data races for this undefined behavior. I will work on the test runner to use child processes

stephencroberts:
> I will work on the test runner to use child processes

Sounds good. Let me know if you'd like me to update the WPT harness. Thanks for the help!

Sytten:
I dont know if it is relevant, but I use a lot [bstr](https://docs.rs/bstr/latest/bstr/index.html) to deal with encoding/decoding in my software since I can keep the stuff in `Vec<u8>` always instead of using `String`.

richarddavison:
> > I will work on the test runner to use child processes
> 
> Sounds good. Let me know if you'd like me to update the WPT harness. Thanks for the help!

Please update the WTP harness. As a temporary work around until the refactor is completed we can also run the web platform tests in its own directory and at run it separately using `llrt test -d <dir>`

stephencroberts:
> Please update the WTP harness.

I'll work on this today, shouldn't be _too_ hard.

> Also note that we recently removed encoding _rs

What do you mean removed? From other parts of LLRT? Should we try to remove it from `TextEncoder/Decoder` as well?

> limits support to only handle utf8, utf16le as per

Ah ok, that should make things simpler!

richarddavison:
> I'll work on this today, shouldn't be _too_ hard.

Fantastic, much appreciated!

> What do you mean removed? From other parts of LLRT? Should we try to remove it from `TextEncoder/Decoder` as well?
 
Yes, we're using a custom implementation that is simpler and faster (less checks). We only support minimal ICU from Node for performance reasons.

stephencroberts:
@richarddavison Oh I see, there are conflicts to resolve. I'll rebase and work through it, but might be tomorrow. I got the test harness updated though so every test can execute in parallel now!

stephencroberts (commented):
@richarddavison I rebased and worked through the conflicts, so it's ready for review again.

With the changes, there are a handful of additional test failures around BOM handling, labels, and a couple of others. I didn't see any easy fixes for them without understanding the spec better, so I've just commented them out.

richarddavison:
@stephencroberts this is awesome work. I can see in the test failures that we’re introducing a JS memory leak in the MacOS builds. This is a bit tricky to locate as it may be a an artifact of races or other issues leading to this. The error that occurs is because we’re dropping the runtime yet still have references to JavaScript objects (in Rust/c).

stephencroberts:
@richarddavison I tried to narrow it down and found that it's just somewhere in the WPT `testharness.js`, and I honestly don't think it's executing any of our Rust code. I get the leak by running an empty test which just bootstraps the test harness and executes the empty test, and I tried to narrow it down to a specific section but there's a lot of code to work through. As far as I can tell, the test harness is extremely bare bones as far as dependencies, and I didn't see anything it's using that is part of LLRT or even rquickjs which makes me think it could be a bug in quickjs. Got any tips for debugging further?

richarddavison:
@stephencroberts found the cause of the leak. It's the new `encode_into` function.

> `testharness.js`

Yeah this might be a QuickJS bug. Would it be possible to try to bootstrap a simple test (like assert(1==1) outside of the test runner? If that still doesn't work, use the raw quickjs runtime to do the same thing? and see if it's reproducible. The leaks are only showing in debug builds of qjs

richarddavison (approved):

stephencroberts:
@richarddavison Whew, glad you found it! I did get the test harness running with just `qjs` and it wasn't leaking. Thanks for getting that fixed for me!

PR #463, opened 2024-07-11
fix: rename hex module to @llrt/hex

nabetti1720:
### Description of changes

While reviewing the module implementation, I noticed that a `hex` module is provided that does not exist in the standard node.js API.

```shell
% node -e "const a = require('hex');"
node:internal/modules/cjs/loader:1148
  throw err;
  ^

Error: Cannot find module 'hex'
Require stack:
- /Users/shinya/Workspaces/llrt-test/[eval]
    at Module._resolveFilename (node:internal/modules/cjs/loader:1145:15)
    at Module._load (node:internal/modules/cjs/loader:986:27)
    at Module.require (node:internal/modules/cjs/loader:1233:19)
    at require (node:internal/modules/helpers:179:18)
    at [eval]:1:11
    at runScriptInThisContext (node:internal/vm:209:10)
    at node:internal/process/execution:118:14
    at [eval]-wrapper:6:24
    at runScript (node:internal/process/execution:101:62)
    at evalScript (node:internal/process/execution:133:3) {
  code: 'MODULE_NOT_FOUND',
  requireStack: [ '/Users/shinya/Workspaces/llrt-test/[eval]' ]
}

Node.js v20.13.1
```

There is a `hex` module in the npm package, but it exists for a completely different purpose.
https://www.npmjs.com/package/hex

~~Therefore, we felt that we should stop publishing the `hex` module within LLRT.~~
Avoid conflicts by renaming the `hex` module to `@llrt/hex`.

### Checklist

- [x] Created unit tests in `tests/unit` and/or in Rust for my feature if needed
- [x] Ran `make fix` to format JS and apply Clippy auto fixes
- [x] Made sure my code didn't add any additional warnings: `make check`
- [x] Updated documentation if needed ([API.md](API.md)/[README.md](README.md)/Other)

_By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice._

PR #468, opened 2024-07-12
fix: rename xml module to llrt:xml

nabetti1720:
### Issue # (if available)

Related #463 

### Description of changes
The following comment mentions the same issue for the `xml` module.
https://github.com/awslabs/llrt/pull/463#issuecomment-2224111842

In this PR, the same action is taken as was taken in the `hex` module.
In addition, change the prefix from `@llrt/<module>`to `llrt:<module>` to match the current trend of other javascript runtimes.

### Checklist


richarddavison:
> Did we consider using `llrt:<module>` namespace? I know you started with hex but since it is not released, there is still time to change it. I feel like`@llrt` really is for npm dependencies. Engines are starting to move toward that convention, node is the primary but others like Bun are [starting to do so](https://github.com/oven-sh/bun/blob/110849355cdd640d4d2494a0d0e96375a69c7f72/packages/bun-types/test.d.ts).

Yeah that would be more in-line with what other Runtimes do! @nabetti1720 can you change `@llrt` module prefixes to` llrt:`?

nabetti1720:
> > Did we consider using `llrt:<module>` namespace? I know you started with hex but since it is not released, there is still time to change it. I feel like`@llrt` really is for npm dependencies. Engines are starting to move toward that convention, node is the primary but others like Bun are [starting to do so](https://github.com/oven-sh/bun/blob/110849355cdd640d4d2494a0d0e96375a69c7f72/packages/bun-types/test.d.ts).
> 
> Yeah that would be more in-line with what other Runtimes do! @nabetti1720 can you change `@llrt` module prefixes to` llrt:`?

Of course it is possible. :)

Sytten:
If you want to move them that is less work I have to do. Just make sure they are feature flagged like I started doing :)

nabetti1720:
> If you want to move them that is less work I have to do. Just make sure they are feature flagged like I started doing :)

Thank you for your comment.
I could understand the direction that should be taken. I do not intend to increase your work :) , but let this PR be up to renaming as per your original purpose.

richarddavison (approved):

Issue #469, opened 2024-07-12
✨ Support Extra Root Certificates


Chagrilled:
LLRT should also expose an `LLRT_EXTRA_CA_CERTS`, mirroring [`NODE_EXTRA_CA_CERTS`](https://nodejs.org/api/cli.html#node_extra_ca_certsfile) to utilise user-provided CA certs so self-signed CAs can be used.

 Presumably this https://github.com/awslabs/llrt/blob/main/llrt_core/src/modules/net/mod.rs#L64 array should be extended with certs parsed from a .pem file

Chagrilled:
Added in #470

PR #470, opened 2024-07-12
add support for extra ca certs

georgesmith46:
### Issue #
#469 

### Description of changes

Adds support for extra CA certificates, mirroring the NODE_EXTRA_CA_CERTS functionality in Node.js. This is utilizing the rustls-pemfile crate for PEM file parsing.

The environment variable `LLRT_EXTRA_CA_CERTS` is used to configure the path to the extra certificates file.

I didn't see any relevant tests for this module. I'm happy to add them if required.

### Checklist

### DIFF
--- a/llrt_core/src/modules/net/mod.rs
+++ b/llrt_core/src/modules/net/mod.rs
@@ -68,11 +68,17 @@ pub static TLS_CONFIG: Lazy<ClientConfig> = Lazy::new(|| {
     }
 
     if let Ok(extra_ca_certs) = env::var(environment::ENV_LLRT_EXTRA_CA_CERTS) {
-        if let Ok(file) = File::open(extra_ca_certs) {
-            let mut reader = BufReader::new(file);
-            root_certificates.add_parsable_certificates(
-                rustls_pemfile::certs(&mut reader).filter_map(std::io::Result::ok),
-            );
+        match File::open(&extra_ca_certs) {
+            Ok(file) => {
+                let mut reader = BufReader::new(file);
+                root_certificates.add_parsable_certificates(
+                    rustls_pemfile::certs(&mut reader).filter_map(std::io::Result::ok),
+                );
+            },
+            _ => warn!(
+                r#"Ignoring extra certs from "{}", load failed"#,
+                extra_ca_certs
+            ),
         }
     }
 
--- a/llrt_core/src/modules/net/mod.rs
+++ b/llrt_core/src/modules/net/mod.rs
@@ -2,7 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 mod socket;
 
-use std::{env, time::Duration};
+use std::{env, fs::File, io::BufReader, time::Duration};
 
 use bytes::Bytes;
 use http_body_util::Full;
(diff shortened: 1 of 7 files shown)


llrt_core/src/modules/net/mod.rs
```diff
@@ -67,6 +67,21 @@ pub static TLS_CONFIG: Lazy<ClientConfig> = Lazy::new(|| {
...
 
+    if let Ok(extra_ca_certs) = env::var(environment::ENV_LLRT_EXTRA_CA_CERTS) {
+        match File::open(&extra_ca_certs) {
+            Ok(file) => {
+                let mut reader = BufReader::new(file);
+                root_certificates.add_parsable_certificates(
+                    rustls_pemfile::certs(&mut reader).filter_map(std::io::Result::ok),
+                );
+            },
+            _ => warn!(
```
richarddavison, line 78: Is warning sufficient? Isn't this a configuration error?

richarddavison (changes requested):
Hi @georgesmith46, thanks for the PR! Looks good just lacks some documentation. Could you add a section to README?

llrt_core/src/modules/net/mod.rs
georgesmith46: I know Node will just emit a warning whereas Deno will throw a JS error. I suspect throwing a JS error will involve some refactoring since these statics will need to be wrapped in a `Result`. I'll have a look into it 👍

georgesmith46: I've wrapped the statics in a `Result`. I've used `thiserror` to avoid the usual error boilerplate. Let me know what you think.

I noticed that exceptions thrown within the [init globals](https://github.com/awslabs/llrt/blob/f97d0b92798691458d183d594219c6bbb5406895/llrt_core/src/vm.rs#L426) stage are not manually caught and therefore it just prints `Error: Exception` instead of `Error: [message]`. I'm guessing this isn't intentional so I can raise an issue and fix this separately.

llrt_core/src/modules/net/mod.rs
```diff
@@ -41,41 +50,53 @@ pub fn get_pool_idle_timeout() -> u64 {
     pool_idle_timeout
 }
 
-pub static HTTP_CLIENT: Lazy<Client<HttpsConnector<HttpConnector>, Full<Bytes>>> =
-    Lazy::new(|| {
-        let pool_idle_timeout: u64 = get_pool_idle_timeout();
+pub static HTTP_CLIENT: Lazy<
+    ClientConfigResult<Client<HttpsConnector<HttpConnector>, Full<Bytes>>>,
```
richarddavison, line 54: Can we return an ioresult instead and avoid pulling in a dependency?

richarddavison (changes requested):
Looks great, thank you! Just a minor modification!

llrt_core/src/modules/net/mod.rs
georgesmith46: Sure, I've changed this to a `std::io::Result`. However, since `std::io::Error` doesn't implement Clone, the TLS config clone for the HTTP client is a bit more verbose than I'd like.

richarddavison (approved):

PR #474, opened 2024-07-14
fix: revert Encoding API

nabetti1720:
### Issue # (if available)

https://github.com/awslabs/llrt/pull/431#issuecomment-2227203324

### Description of changes

Revert to the implementation prior to applying `encoding_rs` for optimal performance and footprint.

- The source structure (e.g., file separation) maintains the most recent structure.
- Methods and properties that can be retained are retained. However, there are some that are not functional.
- ~~Support for `utf16le` is not included in this PR because we are not sure if it can be done immediately.~~
For UTF-16LE, only the conversions required by TextDecoder are supported.

### Checklist


llrt_utils/src/encoding.rs
```diff
@@ -113,3 +128,13 @@ pub fn bytes_to_hex_string(bytes: &[u8]) -> String {
 pub fn bytes_to_string(bytes: &[u8]) -> String {
     String::from_utf8_lossy(bytes).to_string()
 }
+
+pub fn bytes_to_utf16_string(bytes: &[u8]) -> Result<String, String> {
+    let data16 = bytes
+        .chunks(2)
```
richarddavison, line 134: There is a nighly “array_chunks” method that can be used. It can be feature gated so it uses that only on nightly (see json_escape for reference on feature gate nightly)

nabetti1720: I don't know if I have the correct way to turn off `warning`, but I have responded.
Also, I didn't know which type of build this would be used for.

nabetti1720: This is my first time using feature gate, so I am not confident that I have done it well, but I have responded.

nabetti1720:
All but TextEncoder.encodeInto() are now supported.

nabetti1720:
I think we could have `utf-16le` support without using `encoding_rs`. :)

llrt_core/src/modules/encoding/text_decoder.rs
```diff
@@ -65,25 +60,27 @@ impl<'js> TextDecoder {
...
+            .or_throw(&ctx)?;
+
+        if self.fatal
+            && str
+                .as_bytes()
+                .windows(3)
+                .any(|window: &[u8]| window == [0xEF, 0xBF, 0xBD])
+        {
+            return Err(Exception::throw_type(&ctx, "Fatal error"));
         }
```
richarddavison, line 83: We're already checking the string and replacing unknown chars to � in the https://github.com/nabetti1720/llrt/blob/8618194fafd85fc16eed82e2be0a6588d2338897/llrt_core/src/modules/encoding/encoder.rs#L113 method. It would be more optimal to not do `String::from_utf8_lossy` if `self.fatal`. This avoids iterating through the string twice.

llrt_utils/src/encoding.rs
```diff
@@ -73,6 +76,7 @@ impl Encoder {
             Self::Hex => bytes_from_hex(&bytes),
             Self::Base64 => bytes_from_b64(&bytes),
             Self::Utf8 | Self::Iso88591 => Ok(bytes),
+            Self::Utf16le => Err(String::from("This method is not yet supported")),
```
richarddavison, line 79: Forgot to implement?

llrt_utils/src/encoding.rs
```diff
@@ -82,6 +86,17 @@ impl Encoder {
             Self::Hex => bytes_from_hex(string.as_bytes()),
             Self::Base64 => bytes_from_b64(string.as_bytes()),
             Self::Utf8 | Self::Iso88591 => Ok(string.into_bytes()),
+            Self::Utf16le => Err(String::from("This method is not yet supported")),
```
richarddavison, line 89: Forgot to implement?

richarddavison (changes requested):
Almost there! Just some minor changes!

llrt_core/src/modules/encoding/text_decoder.rs
nabetti1720: Indeed it is! I have corrected it and would appreciate a review.

llrt_utils/src/encoding.rs
nabetti1720: I did not forget it, but I did not implement it because it was not needed in Text[En|De]coder and I was not confident in implementing it correctly.

Please check back with me once I have fixed it.

nabetti1720:
Hi @richarddavison , All have been addressed. :)

richarddavison (approved):
Awesome 👏

llrt_utils/Cargo.toml
```diff
@@ -18,3 +18,6 @@ rquickjs = { version = "0.6.2", features = [
   "array-buffer",
 ], default-features = false }
 tokio = { version = "1", optional = true }
+
```
richarddavison, line 21: Nice!

PR #486, opened 2024-07-17
Windows support for modules + move os 

Sytten:
### Description of changes

-  Added windows support for the modules
- Move OS module and added windows support
- Added CI for modules for windows, macos and linux (x86 only, it is sufficient)

### Checklist

### DIFF
--- a/llrt_modules/Cargo.toml
+++ b/llrt_modules/Cargo.toml
@@ -6,11 +6,18 @@ license-file = "LICENSE"
 
 [features]
 default = ["all"]
-all = ["buffer", "fs", "os", "path"]
+all = ["buffer", "fs", "os", "path", "process"]
 
 buffer = ["llrt_utils/encoding"]
 fs = ["tokio/fs", "llrt_utils/fs", "ring", "buffer", "path"]
-os = ["process", "libc", "once_cell"]
+os = [
+  "process",
+  "libc",
+  "windows-result",
+  "windows-registry",
+  "windows-version",
+  "once_cell",
+]
 path = []
 process = []
 
@@ -28,6 +35,11 @@ tokio = { version = "1", features = ["rt", "io-util"] }
 
 [target.'cfg(unix)'.dependencies]
 libc = { version = "0.2", optional = true }
+
+[target.'cfg(windows)'.dependencies]
+windows-registry = { version = "0.2", optional = true }
+windows-result = { version = "0.2", optional = true }
+windows-version = { version = "0.1", optional = true }
 
 [dev-dependencies]
 nanoid = "0.4.0"
--- a/llrt_modules/Cargo.toml
+++ b/llrt_modules/Cargo.toml
@@ -10,11 +10,13 @@ all = ["buffer", "fs", "os", "path"]
 
 buffer = ["llrt_utils/encoding"]
 fs = ["tokio/fs", "llrt_utils/fs", "ring", "buffer", "path"]
-os = ["libc"]
+os = ["process", "libc", "once_cell"]
 path = []
+process = []
 
 [dependencies]
 either = "1"
+once_cell = { version = "1", optional = true }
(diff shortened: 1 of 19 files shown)


llrt_modules/Cargo.toml
```diff
@@ -23,6 +33,14 @@ rquickjs = { version = "0.6", features = [
 ring = { version = "0.17", optional = true }
 tokio = { version = "1", features = ["rt", "io-util"] }
 
+[target.'cfg(unix)'.dependencies]
+libc = { version = "0.2", optional = true }
+
+[target.'cfg(windows)'.dependencies]
+windows-registry = { version = "0.2", optional = true }
+windows-result = { version = "0.2", optional = true }
+windows-version = { version = "0.1", optional = true }
```
Sytten, line 42: Those are official microsoft utils, I started by writing my own low level calls but they are packaging them now so it's better to use that IMO

llrt_modules/src/modules/os/windows.rs
```diff
@@ -0,0 +1,44 @@
...
+
+    let registry_key = LOCAL_MACHINE.open("SOFTWARE\\Microsoft\\Windows NT\\CurrentVersion")?;
+    let value = registry_key.get_value("ProductName")?;
+    let Value::String(mut value) = value else {
+        return Err(Error::empty());
+    };
+
+    // Windows 11 shares dwMajorVersion with Windows 10
+    // this workaround tries to disambiguate that by checking
+    // if the dwBuildNumber is from Windows 11 releases (>= 22000).
```
Sytten, line 38: I took this logic from libuv

richarddavison (approved):

richarddavison:
LGTM!

llrt_modules/src/modules/process.rs
```diff
@@ -0,0 +1,16 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+use std::env;
+
+// This is not the full module for now since we are not yet sure
+// how to implement the TIME_ORIGIN Atomic in what are supposed to
+// be independant modules.
```
Sytten, line 7: If you have suggestions on how to do that @richarddavison I am all hears.
It is very much tied to the VM, which is a core concept that we don't want in the modules.

nabetti1720: Other than vm.rs, TIME_ORIGIN is used for processes and performance. I believe that `process` is a Node.js API, so it is eligible to be moved to `llrt_modules`, but `performance` is a WinterCG API, so it will remain in `llrt_core` (although perf_hooks API may be implemented in the future...).

So we think it is appropriate to move the following instruction to `llrt_utils/src/lib.rs`. What do you think?

```rust
pub static TIME_ORIGIN: AtomicUsize = AtomicUsize::new(0);
```

Sytten: Probably but the problem really is that its not obvious for a user of the modules that it needs to set that thing.
Thats why I am saying it should done at module declaration probably?

richarddavison: Hum. Can we move it to utils with a lazy init once_cell or similar? Then call the init from all models where it used? This will effectively only call it once with negligible overhead

Sytten: Probably not since we need the time the "VM" started. So there needs to be an explicit call from the user. 
The more I think about it, the more I feel this is something we should put in `rquickjs`.

richarddavison: Right I see what you mean. It doesn’t have to tho. It’s just relative time. So it’s fine to init lazy. In LLRT runtime we should just run the init ASAP in the VM by getting the value

Sytten: Good idea, I will move it once this gets merged.

PR #488, opened 2024-07-20
add perf_hooks module 

ahaoboy:
### Description of changes
Add node perf_hooks module, currently only the most commonly used function performance.now is added
https://nodejs.org/api/perf_hooks.html#performancenow

### Checklist


richarddavison:
Hi @ahaoboy thanks for the PR. We already have this https://github.com/awslabs/llrt/blob/main/llrt_core/src/modules/performance.rs. How is it different? If not, I see no reason to add this API also since users can use the performance module.

ahaoboy:
> How is it different

I guess it should be a historical problem of nodejs, and the process type can be used through modules or through global variables
Some libraries use the performance.now function in the form of modules

ahaoboy:
They are same, so can I merge them to llrt_modules?
```
import { performance as perf } from 'perf_hooks'
console.log(performance === perf)
```

nabetti1720:
Hi @ahaoboy , The reason there have been no `perf_hooks` is simply that the Performance API was created in the process of conformance to WinterCG and was not aimed at the goal of being a Node.js compatible API. It would be great if these implementations could be shared and used.

I am also very interested in how to split the implementation into `lrt_core` and `llrt_modules` when sharing implementation. This would be helpful for future migration of modules with shared implementations.

ahaoboy:
Unfortunately, many npm libraries are designed based on the nodejs api, so many of them use performance in the following form:

```
import { performance } from 'perf_hooks'
```

Sytten:
I was going to take care of the global variable, we had a good solution for it in my windows PR.
Also if you can add the typing in the `type` that would be nice!

ahaoboy:
1. Move the code in llrt_core/performance to llrt_modules/perf_hooks
2. Move TIME_ORIGIN in the vm module to perf_hooks and initialize it when creating a performance instance
3. Due to the limitation of the life cycle, there is no easy way to create a global variable `assert(globalThis.performance === require('perf_hooks').performance)`

richarddavison:
There is still a lot of duplication in this module. If it’s simply an alias it should import the existing performance module. Also `get_time_origin` can be moved to utils and called from VM the constructor as well

ahaoboy:
> There is still a lot of duplication in this module. 

Can you be more specific? I'm not very familiar with the llrt project.

richarddavison:
Hi @ahaoboy sorry for the delay. I think the way forward here is if the hooks module imports performance module so they refer to the same object.

ahaoboy:
> I think the way forward here is if the hooks module imports performance module so they refer to the same object.

It is very difficult to implement shared global variables on the rust side. The only way I can think of is to put the performance object into the global object, such as `globalThis.__performance__ = performance`, but this is not elegant. Maybe there is a better way?

The console module also has similar issues, but llrt does not seem to have this feature implemented yet.
```
import {Console} from 'console'

// true
console.log(Console === console.Console)
```

llrt_modules/src/modules/perf_hooks.rs
```diff
@@ -0,0 +1,169 @@
...
+    obj.set("timeOrigin", get_time_origin())?;
+
+    Ok(obj)
+}
+
+pub fn new_performance(ctx: Ctx<'_>) -> Result<Object<'_>> {
+    if TIME_ORIGIN.load(Ordering::Relaxed) == 0 {
+        let time_origin = Utc::now().timestamp_nanos_opt().unwrap_or_default() as usize;
+        TIME_ORIGIN.store(time_origin, Ordering::Relaxed)
+    }
```
richarddavison, line 41: Move this out and create an "init" function that sets time origin if not already set. Then call this `init` function from vm.rs

richarddavison (changes requested):
Almost there! Just some minor comments

llrt_modules/src/modules/perf_hooks.rs
ahaoboy: Since TIME_ORIGIN has been moved to the perf_hooks module, the peformance module can now also be moved to llrt_modules
Added perf_hooks/init function, called when the VM is initialized

richarddavison (approved):

Sytten:
@ahaoboy I fixed it in my PR #494

richarddavison:
@ahaoboy please rebase and force push and we'll merge :)

PR #490, opened 2024-07-20
fix: throwing exceptions during module initialization

### LINKED ISSUE #482: Exceptions thrown during module initialization are not caught
Exceptions thrown during the initialization of a global or module are not caught. This means exceptions thrown at this point will be missing the exception message.

How to replicate
---------
Running this
```
LLRT_NET_DENY=/ ./llrt -e 'console.log("123")'
```
Outputs
```
Error: Exception
```
But it should output
```
ReferenceError: LLRT_NET_DENY env contains an invalid URI
```

How to fix
----------
This is where the exceptions should probably be caught:
https://github.com/awslabs/llrt/blob/9e21153db5b7c6a3107462e97989b3c6a0e868c4/llrt_core/src/vm.rs#L426

Example of catching an exception: https://github.com/awslabs/llrt/blob/9e21153db5b7c6a3107462e97989b3c6a0e868c4/llrt_core/src/vm.rs#L448

nabetti1720:
### Issue # (if available)

Fixed #482 

### Description of changes

This PR will allow the correct output of the content of errors during module initialization.

After the revision:
```shell
% LLRT_NET_DENY=/ ./llrt -e 'console.log("123")'
ReferenceError: LLRT_NET_DENY env contains an invalid URI
```

### Checklist

### DIFF
--- a/llrt_core/src/vm.rs
+++ b/llrt_core/src/vm.rs
@@ -423,10 +423,16 @@ impl Vm {
         let ctx = AsyncContext::full(&runtime).await?;
         ctx.with(|ctx| {
             for init_global in init_globals {
-                init_global(&ctx)?;
+                if let Err(err) = init_global(&ctx).catch(&ctx) {
+                    Self::print_error_and_exit(&ctx, err);
+                }
             }
-            timers::init_timers(&ctx)?;
-            init(&ctx, module_names)?;
+            if let Err(err) = timers::init_timers(&ctx).catch(&ctx) {
+                Self::print_error_and_exit(&ctx, err);
+            }
+            if let Err(err) = init(&ctx, module_names).catch(&ctx) {
+                Self::print_error_and_exit(&ctx, err);
+            }
             Ok::<_, Error>(())
         })
         .await?;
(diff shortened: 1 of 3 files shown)


llrt_core/src/vm.rs
```diff
@@ -423,10 +423,16 @@ impl Vm {
         let ctx = AsyncContext::full(&runtime).await?;
         ctx.with(|ctx| {
             for init_global in init_globals {
-                init_global(&ctx)?;
+                if let Err(err) = init_global(&ctx).catch(&ctx) {
```
richarddavison, line 426: Since we’re handling the error the same way we can put everything in a closure and handle the error once:

```rust
({
//all init code
}).catch()…
```

richarddavison (changes requested):
Thanks for the fix! I think we can merge all error handling for init!

llrt_core/src/vm.rs
nabetti1720: Thanks for pointing that out. It was a little confusing, but there was a similar implementation, so we modified it based on the following. :)

https://github.com/awslabs/llrt/blob/9824e4163fb8cf103066d7b5938929740304a4ff/llrt_core/src/compiler.rs#L65-L66

richarddavison (approved):

PR #494, opened 2024-07-23
Move child_process and events to llrt_modules

Sytten:
### Description of changes

- Move child process and events modules to llrt_modules
- Generalize the `spawn_exit` to allow users to not exit on error

### Checklist


llrt_utils/src/ctx.rs
```diff
@@ -0,0 +1,73 @@
...
+                        error_handler(&ctx, err);
+                    }
+                },
+            }
+        });
+        Ok(join_channel_rx)
+    }
+}
+
+pub fn set_error_handler<F>(handler: F)
```
richarddavison, line 68: This is great! However not quite sure about the naming. It is only being used for spawned/async tasks. Maybe `set_spawn_error_handler` or `set_async_error_handler`, suggestions welcome?

richarddavison (changes requested):
Fantastic! Only a small comment!

llrt_utils/src/ctx.rs
Sytten: I like `set_spawn_error_handler`, pushing a fix now

richarddavison (approved):

Issue #496, opened 2024-07-26
types support node: protocol


ahaoboy:
Currently only legacy module types are supported

```
declare module "os" {
}
```

Adding Type Declarations
```
declare module "node:os" {
    export * from "os";
}
```

richarddavison:
Closed as explained in https://github.com/awslabs/llrt/pull/497#issuecomment-2253466831

PR #497, opened 2024-07-26
types: add node protocol

ahaoboy:
### Issue # (if available)
https://github.com/awslabs/llrt/issues/496

### Description of changes

Add node protocol type support

### Checklist

### DIFF
--- a/types/buffer.d.ts
+++ b/types/buffer.d.ts
@@ -166,3 +166,7 @@ declare module "buffer" {
 
   export { Buffer };
 }
+
+declare module "node:buffer" {
+  export * from "buffer";
+}
--- a/types/fs.d.ts
+++ b/types/fs.d.ts
@@ -326,3 +326,7 @@ declare module "fs" {
    */
   export function accessSync(path: PathLike, mode?: Mode): void;
 }
+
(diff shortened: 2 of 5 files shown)


Sytten:
I explicitly decided not to add the `node:*` version since I think it should probably be `llrt:*` instead but I will leave @richarddavison to decide want he wants to do with that

ahaoboy:
I think it is not a good idea to use the llrt protocol for general modules. This will cause a large number of npm libraries to fail to run. Unless it is a specific module like llrt:ffi, using the llrt protocol will make it more difficult for library developers to write code that can run on platforms such as node, deno, llrt, etc.

richarddavison:
It's a bit of a mixed bag here, pros and cons with both approaches. I'm a bit skeptical to add `node:` prefix to typings because it explicitly states that we're targeting/having module compatibility with node. On the other hand, we already strip the `node:` prefix when trying to resolve modules. We by design tries to mimic the Node API for the modules we support but there are no guarantess for full compatibility. For instance there are a lot of `buffer` APIs in Node that we don't support (yet, and some we never will). Hope this makes sense!

PR #498, opened 2024-07-26
feat: Encoding API Improvements

nabetti1720:
### Description of changes

To improve compatibility with encodings supported by Node.js, the following actions have been taken.
https://nodejs.org/api/util.html#whatwg-supported-encodings

- For encodings supported by LLRT, modified so that they can also be specified by aliases.
- Reduced excessive normalization and made label judgments a little stricter.
- Since `iso-8859-1` was an alias, it was managed by `windows-1252`.
- `ENCODING_MAP` was introduced with the expectation of improved maintenance ~and lookup performance~.

### Checklist


llrt_utils/src/encoding.rs
```diff
@@ -2,62 +2,67 @@
...
-    pub enum Encoder {
-        Hex,
-        Base64,
-        Utf8,
-        Iso88591,
-        Utf16le,
-    }
-}
+static ENCODING_MAP: Lazy<HashMap<&'static str, Encoder>> = Lazy::new(|| {
+    let mut map = HashMap::new();
```
richarddavison, line 18: We should create with capacity as we know up front how many items to put in there

richarddavison (approved):
Hi @nabetti1720! I like the aliasing and more simple approach. Not sure if this yields much performance benefit (except for maybe one less string allocation to normalize the key) but it's more maintainable for sore.

nabetti1720:
Hi @richarddavison .
Yes. In the previous implementation, it was necessary to create a new enum when dealing with aliases that exceeded the normalization range (e.g., treating `unicode-1-1-utf8` as `utf-8`), and this proposal was made to avoid this.

We believe that by minimizing the normalization, we can now evaluate with the same string as the actual label. 

Finally, as for performance, this level of performance would not make a difference. :)

types/buffer.d.ts
```diff
@@ -1,12 +1,9 @@
...
   export type BufferEncoding =
-    | "utf8"
     | "utf-8"
-    | "utf16le"
     | "utf-16le"
+    | "windows-1252"
     | "base64"
-    | "iso88591"
-    | "iso-8859-1"
     | "hex";
```
Sytten, line 7: We might want to keep the typing for the variations

nabetti1720: It is not accurate to leave only some aliases. We now enumerate all labels that can be mapped.

richarddavison (approved):

PR #501, opened 2024-07-29
Move llrt_core/process to llrt_modules/process

ahaoboy:
### Description of changes

Move llrt_core/process to llrt_modules/process 

### Checklist


llrt_modules/src/lib.rs
```diff
@@ -11,3 +12,7 @@ pub mod stream;
 #[cfg(test)]
 mod test;
 mod utils;
+
+pub const VERSION: &str = env!("CARGO_PKG_VERSION");
+
+pub static TIME_ORIGIN: AtomicUsize = AtomicUsize::new(0);
```
Sytten, line 18: I think we need to have the discussion on that static at https://github.com/awslabs/llrt/issues/503

llrt_modules/src/lib.rs
```diff
@@ -11,3 +12,7 @@ pub mod stream;
 #[cfg(test)]
 mod test;
 mod utils;
+
+pub const VERSION: &str = env!("CARGO_PKG_VERSION");
```
Sytten, line 16: I would like to be able to override that version somehow, but it can wait

llrt_modules/src/modules/process.rs
```diff
@@ -14,3 +23,152 @@ pub fn get_platform() -> &'static str {
...
+fn env_proxy_setter<'js>(
+    target: Object<'js>,
+    prop: Value<'js>,
+    value: Coerced<String>,
+) -> Result<bool> {
+    target.set(prop, value.to_string())?;
+    Ok(true)
+}
+
+pub fn init(ctx: &Ctx<'_>) -> Result<()> {
```
Sytten, line 81: I think down the line we need to allow module settings to say not allow the module to load the process env.

Sytten:
Can you also add typescript typing :)

richarddavison (commented):
Thanks. Just a minor comment

llrt_modules/src/modules/process.rs
```diff
@@ -14,3 +23,152 @@ pub fn get_platform() -> &'static str {
         _ => platform,
     }
 }
+
+fn cwd() -> String {
```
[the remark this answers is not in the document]
ahaoboy: This seems like it would cause a lifetime error, or is there some other way to return a &str?

```
fn cwd() -> &str {
     env::current_dir().unwrap().to_string_lossy().as_ref()
}
```

```
error[E0515]: cannot return value referencing temporary value
  --> llrt_modules/src/modules/process.rs:28:6
   |
28 |      env::current_dir().unwrap().to_string_lossy().as_ref()
   |      ---------------------------------------------^^^^^^^^^
   |      |
   |      returns a value referencing data owned by the current function
   |      temporary value created here
```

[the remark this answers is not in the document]
Sytten: Yeah you can avoid the double alloc by returning a `String<'js>` built from the `&str`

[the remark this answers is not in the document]
ahaoboy: Am I understanding this correctly?
```
fn cwd(ctx: Ctx<'_>) -> std::io::Result<Result<rquickjs::String<'_>>> {
    env::current_dir().map(|s| rquickjs::String::from_str(ctx, s.to_string_lossy().as_ref()))
}
```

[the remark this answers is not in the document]
Sytten: use an `and_then` to avoid nesting your result, otherwise yes

types/process.d.ts
```diff
@@ -14,5 +18,305 @@ declare module "process" {
...
+      | "loong64"
+      | "mips"
+      | "mipsel"
+      | "ppc"
+      | "ppc64"
+      | "riscv64"
+      | "s390"
+      | "s390x"
+      | "x64";
+  }
```
Sytten, line 48: Lets be diligent about typing, we dont support nearly half of those platforms and architecture.

types/process.d.ts
```diff
@@ -14,5 +18,305 @@ declare module "process" {
...
+  interface ProcessVersions extends Dict<string> {
+    http_parser: string;
+    node: string;
+    v8: string;
+    ares: string;
+    uv: string;
+    zlib: string;
+    modules: string;
+    openssl: string;
+  }
```
Sytten, line 75: AFAIK we only set `llrt` right now, please update

types/process.d.ts
```diff
@@ -14,5 +18,305 @@ declare module "process" {
...
+  interface ProcessEnv extends Dict<string> {}
+  interface HRTime {
+    (time?: [number, number]): [number, number];
+  }
+  interface ProcessRelease {
+    name: string;
+    sourceUrl?: string | undefined;
+    headersUrl?: string | undefined;
+    libUrl?: string | undefined;
+    lts?: string | undefined;
```
Sytten, line 64: We only support `name`

types/process.d.ts
```diff
@@ -14,5 +18,305 @@ declare module "process" {
...
+     *
+     * ```js
+     * {
+     *   name: 'node',
+     *   lts: 'Hydrogen',
+     *   sourceUrl: 'https://nodejs.org/download/release/v18.12.0/node-v18.12.0.tar.gz',
+     *   headersUrl: 'https://nodejs.org/download/release/v18.12.0/node-v18.12.0-headers.tar.gz',
+     *   libUrl: 'https://nodejs.org/download/release/v18.12.0/win-x64/node.lib'
+     * }
+     * ```
```
Sytten, line 194: Change all the examples please to remove node references

Sytten (changes requested):
The typing needs to be improved quite a bit.
@richarddavison Can we add a line on that in the PT template, the goal of having our own typing is to reflect what we support so contributors must be diligent about adding only the parameters, functions, return values, etc that llrt supports.

llrt_modules/src/modules/process.rs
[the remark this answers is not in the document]
ahaoboy: Where do I add and_then? and_then needs to wrap with Ok, perhaps using current_dir().unwrap(), which we all know almost never fails.
```
env::current_dir().and_then(|path|Ok(rquickjs::String::from_str(ctx, path.to_string_lossy().as_ref())))
```

fmt
```
env::current_dir().and_then(|path| {
    Ok(rquickjs::String::from_str(
        ctx,
        path.to_string_lossy().as_ref(),
    ))
})
```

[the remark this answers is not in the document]
richarddavison:
```rust
fn cwd<'js>(ctx: Ctx<'js>) -> Result<rquickjs::String<'js>> {
    let cwd = env::current_dir().or_throw(&ctx)?;
    let cwd = cwd.to_string_lossy();
    let cwd = cwd.as_ref();
    rquickjs::String::from_str(ctx, cwd)
}
```

richarddavison (approved):

PR #515, opened 2024-08-01
move exceptions and navigator to llrt_modules

ahaoboy:
### Description of changes

move exceptions and navigator to llrt_modules

In the test file, use function globalThis.DOMException() to create an instance, but both node and the web only support creating instances in the form of a constructor, I don't know if there's a special scenario.

### Checklist


ahaoboy:
> Unsure what caused the windows test to fail, this error usually means we tried to access some memory we were not supposed to. There is probably a bug in child or fs module...

Using x86_64-pc-windows-gnu on windows platform is more convenient than x86_64-pc-windows-msvc, I can pass the test with gnu locally.

ahaoboy:
From current experience, windows-gnu and Linux-musl are more stable

.github/workflows/ci.yml
```diff
@@ -65,15 +65,18 @@ jobs:
...
             arch: x86_64
+            toolchain: stable
           - os: macos-latest
             platform: darwin
             arch: x86_64
+            toolchain: stable
           - os: windows-latest
             platform: windows
             arch: x86_64
+            toolchain: stable-x86_64-pc-windows-gnu
```
richarddavison, line 76: do we have to specify this?

richarddavison (changes requested):

.github/workflows/ci.yml
Sytten: yeah otherwise the default on windows is msvc

ahaoboy: I would recommend using the gnu toolchain on windows, especially if you need to compile a c project, as there are fewer bugs than with msvc.

richarddavison (approved):

richarddavison:
> There is a small optimization that I didn't add, I don't know if you will like it
> 
> ```
> fn get_user_agent() -> String {
>     ["llrt ", VERSION].concat()
> }
> ```
> 
>  ```
> fn get_user_agent() -> &'static str {
>    concat!("llrt ", env!("CARGO_PKG_VERSION"))
> }
> ```

Yes this is great! Please add and I'll merge

richarddavison:
@ahaoboy sorry please rebase again. I was not fast enough!

llrt_modules/src/modules/navigator.rs
```diff
@@ -2,10 +2,8 @@
 // SPDX-License-Identifier: Apache-2.0
 use rquickjs::{Ctx, Object, Result};
 
-use crate::VERSION;
-
-fn get_user_agent() -> String {
-    ["llrt ", VERSION].concat()
+fn get_user_agent() -> &'static str {
+    concat!("llrt ", env!("CARGO_PKG_VERSION"))
```
Sytten, line 6: You should be able to use the constant with concat

llrt_modules/Cargo.toml
```diff
@@ -19,6 +19,8 @@ all = [
   "process",
   "perf_hooks",
   "performance",
+  "exceptions",
+  "navigator",
```
Sytten, line 23: @richarddavison Those APIs are not node related, so thats why I think some kind of marker would be good

Sytten (changes requested):
Can you add them to the `README` too, thanks :)

llrt_modules/Cargo.toml
Sytten: Maybe something like
```rust
/// This is my module blablabla
///
/// # Reference
/// Source: node
```
It would be cool to produce a diagram of APIs supported from each reference implementation by scrapping the documentation.

Otherwise a macro like
```rust
#[ref(source: node)]
impl ModuleDef for MyModule {
  ...
}
```

Then we can generate an `impl ModuleSource for MyModule` and aggregate those easily.

llrt_modules/src/modules/navigator.rs
ahaoboy: [concat macro](https://doc.rust-lang.org/std/macro.concat.html) only supports literals

llrt_modules/Cargo.toml
ahaoboy: There are two kinds of APIs, node and browser. Most of the APIs on browsers are introduced in the form of global variables. Nodejs recommends modular usage.

PR #523, opened 2024-08-04
feat: implement zlib module

nabetti1720:
### Description of changes

Implement a portion of the Node.js zlib module ([Convenience methods](https://nodejs.org/api/zlib.html#convenience-methods)).

It is still rough around the edges at this point and some TODOs remain.

- ~Need to support more than just Brotli.~
All except unzip/unzipSync are now supported.
- ~Options also need to be supported as much as possible.~
For the time being, only the compression level has been addressed.
- ~Types/Tests/Docs also.~
All have been addressed.

### Checklist

### DIFF
--- a/API.md
+++ b/API.md
@@ -296,16 +296,6 @@ export class URLSearchParams {
 
 [brotliDecompressSync](https://nodejs.org/api/zlib.html#zlibbrotlidecompresssyncbuffer-options)
 
-
-> [!NOTE]
-> The following functions are LLRT's own implementation of Zstandard format compression/decompression.
-
-- zstandardCompress
-- zstandardCompressSync
-- zstandardDecompress
-- zstandardDecompressSync
-
-
 ## llrt:hex
 
 ```typescript
--- a/API.md
+++ b/API.md
@@ -262,6 +262,50 @@ export class URLSearchParams {
 
 [TextEncoder](https://nodejs.org/api/util.html#class-utiltextdecoder)
 
+## zlib
+
+[deflate](https://nodejs.org/api/zlib.html#zlibdeflatebuffer-options-callback)
+
+[deflateSync](https://nodejs.org/api/zlib.html#zlibdeflatesyncbuffer-options)
+
+[deflateRaw](https://nodejs.org/api/zlib.html#zlibdeflaterawbuffer-options-callback)
+
+[deflateRawSync](https://nodejs.org/api/zlib.html#zlibdeflaterawsyncbuffer-options)
+
+[gzip](https://nodejs.org/api/zlib.html#zlibgzipbuffer-options-callback)
+
+[gzipSync](https://nodejs.org/api/zlib.html#zlibgzipsyncbuffer-options)
+
+[inflate](https://nodejs.org/api/zlib.html#zlibinflatebuffer-options-callback)
+
+[inflateSync](https://nodejs.org/api/zlib.html#zlibinflatesyncbuffer-options)
+
+[inflateRaw](https://nodejs.org/api/zlib.html#zlibinflaterawbuffer-options-callback)
+
+[inflateRawSync](https://nodejs.org/api/zlib.html#zlibinflaterawsyncbuffer-options)
+
+[gunzip](https://nodejs.org/api/zlib.html#zlibgunzipbuffer-options-callback)
+
+[gunzipSync](https://nodejs.org/api/zlib.html#zlibgunzipsyncbuffer-options)
+
+[brotliCompress](https://nodejs.org/api/zlib.html#zlibbrotlicompressbuffer-options-callback)
+
+[brotliCompressSync](https://nodejs.org/api/zlib.html#zlibbrotlicompresssyncbuffer-options)
+
+[brotliDecompress](https://nodejs.org/api/zlib.html#zlibbrotlidecompressbuffer-options-callback)
+
+[brotliDecompressSync](https://nodejs.org/api/zlib.html#zlibbrotlidecompresssyncbuffer-options)
+
+
+> [!NOTE]
+> The following functions are LLRT's own implementation of Zstandard format compression/decompression.
+
+- zstandardCompress
+- zstandardCompressSync
+- zstandardDecompress
+- zstandardDecompressSync
+
+
 ## llrt:hex
 
 ```typescript
--- a/llrt_modules/Cargo.toml
+++ b/llrt_modules/Cargo.toml
@@ -20,6 +20,7 @@ all = [
   "process",
   "exceptions",
   "navigator",
+  "zlib",
 ]
 
 buffer = ["llrt_utils/encoding"]
@@ -38,6 +39,8 @@ process = []
 perf-hooks = []
 navigator = []
 exceptions = []
+zlib = []
+
 __bytearray-buffer = ["tokio/sync"]
 __stream = ["buffer", "__bytearray-buffer"]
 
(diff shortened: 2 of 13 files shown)


types/zlib.d.ts
```diff
@@ -0,0 +1,183 @@
+/**
+ * The `zlib` module provides compression functionality implemented using
+ * Gzip, Deflate/Inflate, Brotli and Zstandard.
+ *
+ * To access it:
+ *
+ * ```js
+ * const zlib = require('zlib');
```
Sytten, line 8: Prefer the examples using esm imports since llrt_modules will we used by other people and ideally they can offer the typing to their users

llrt_modules/src/modules/zlib.rs
```diff
@@ -0,0 +1,293 @@
...
+            ctx: Ctx<'js>,
+            value: Value<'js>,
+            options: Opt<Object<'js>>,
+        ) -> Result<Value<'js>> {
+            $converter(ctx.clone(), value, options, $command)
+        }
+    };
+}
+
+macro_rules! define_cb_function {
```
Sytten, line 38: Thats like a very cool macro, do you think we could generalize it and put it in the util? I know fs is missing all the functions with callbacks and that could help

llrt_modules/Cargo.toml
```diff
@@ -59,6 +61,11 @@ rquickjs = { version = "0.6", features = [
 ring = { version = "0.17", optional = true }
 tokio = { version = "1", features = ["rt", "io-util"] }
 tracing = "0.1"
+flate2 = { version = "1.0.30", features = [
```
Sytten, line 64: All those deps need to be optional and enabled by the feature

llrt_modules/src/modules/zlib.rs
```diff
@@ -0,0 +1,293 @@
...
+            default.set("gunzip", Func::from(gunzip))?;
+            default.set("gunzipSync", Func::from(gunzip_sync))?;
+
+            default.set("brotliCompress", Func::from(br_comp))?;
+            default.set("brotliCompressSync", Func::from(br_comp_sync))?;
+
+            default.set("brotliDecompress", Func::from(br_decomp))?;
+            default.set("brotliDecompressSync", Func::from(br_decomp_sync))?;
+
+            default.set("zstandardCompress", Func::from(zstd_comp))?;
```
Sytten, line 275: Idea @richarddavison we might want to add a comment or macro to tag APIs to know if they respect a certain spec or existing engine. That way we could auto generate a report or something. Like this method is good but its not a node thing (yet). Maybe a rustdoc tag like `/// @engine ` or something.

Sytten (changes requested):

types/zlib.d.ts
nabetti1720: I replaced it with this example, is it correct?
```js
import * as zlib from 'zlib';
```

llrt_modules/Cargo.toml
nabetti1720: Cargo.toml was modified.

llrt_modules/src/modules/zlib.rs
nabetti1720: If it is to be generalized, I would need to make additional considerations regarding the number of parameters other than callback functions.

I have done my best to create a macro for my current use case, but after this PR is merged, I would welcome this macro to be further generalized to work elsewhere. :)

llrt_modules/src/modules/zlib.rs
nabetti1720: Hi @Sytten , Sorry, I did not know what exactly to do about this point...
Should I put the annotation in the rust source? Or should I put the annotations in the types?

As for Zstandard, it is not so difficult to implement, and once Node.js supports it, there is a way to support it in LLRT as well.

Sytten: Do nothing for now :)
This is just an idea so we can track better from which spec each API comes from.

API.md
```diff
@@ -262,6 +262,50 @@ export class URLSearchParams {
...
+
+[brotliDecompress](https://nodejs.org/api/zlib.html#zlibbrotlidecompressbuffer-options-callback)
+
+[brotliDecompressSync](https://nodejs.org/api/zlib.html#zlibbrotlidecompresssyncbuffer-options)
+
+
+> [!NOTE]
+> The following functions are LLRT's own implementation of Zstandard format compression/decompression.
+
+- zstandardCompress
```
richarddavison, line 303: While these are nice to have I'm a bit unsure wether we should add them to the runtime. We have historically tried to stayed clear of custom LLRT APIs (expect for some Lambda specific implementations) as these are not existent in Node.

llrt_modules/src/modules/zlib.rs
```diff
@@ -0,0 +1,293 @@
...
+        pub fn $fn_name<'js>(
+            ctx: Ctx<'js>,
+            value: Value<'js>,
+            args: Rest<Object<'js>>,
+        ) -> Result<()> {
+            let mut args_iter = args.0.into_iter().rev();
+            let cb: Function = args_iter
+                .next()
+                .and_then(|v| <rquickjs::Value<'_> as Clone>::clone(&v).into_function())
+                .or_throw_msg(&ctx, "Callback required")?;
```
richarddavison, line 49: Doesn't `into_function() `automatically clone? Also error message is a bit misleading. "Callback parameter is not a function" or similar is more clear.

llrt_modules/src/modules/zlib.rs
```diff
@@ -0,0 +1,293 @@
...
+    Gunzip,
+}
+
+fn zlib_converter<'js>(
+    ctx: Ctx<'js>,
+    value: Value<'js>,
+    options: Opt<Object<'js>>,
+    command: ZlibCommand,
+) -> Result<Value<'js>> {
+    let src = if value.is_string() {
```
richarddavison, line 87: Why not use `get_bytes()` from utils? Then value can be whatever (buffer, string, typed array etc)

llrt_modules/src/modules/zlib.rs
```diff
@@ -0,0 +1,293 @@
...
+    };
+
+    let mut level = Compression::default();
+    if let Some(options) = options.0 {
+        if let Some(opt) = options.get_optional("level")? {
+            level = Compression::new(opt);
+        }
+    }
+
+    let mut dst: Vec<u8> = Vec::with_capacity(src.len());
```
richarddavison, line 102: Do we need this? Doesn't the encoders reserve based on the source size? Please verify! if not, there are algorithms to calculate approximate size for most compression formats.

richarddavison (changes requested):

llrt_modules/src/modules/zlib.rs
Sytten: I use the same library in our codebase, it is a stream based API so it only allocates a small buffer internally but you have to provide the aggregate buffer.

API.md
Sytten: Thats why I was suggesting we tag the APIs, I imagine some people will want to add WinterGC APIs that are not Node based too at some point.
Anyway this zlib is a good step but we will really need a stream based API for anything serious (that you cant decompress all in RAM).

nabetti1720: @richarddavison and @Sytten , I apologize for the confusion caused by my unnecessary implementation. I will remove the Zstandard implementation until the policy is clarified.

If this becomes necessary before the policy is clarified, I believe it can be provided from the llrt namespace.

llrt_modules/src/modules/zlib.rs
nabetti1720: Wow. I completely missed this one.

llrt_modules/src/modules/zlib.rs
nabetti1720: Sorry, the type of the `options` parameter was mixed between `Object<'js>` and `Value<'js>` and needed to be converted. We have simplified the process by unifying it with `Values<'js>`.

I also reviewed the error messages.

nabetti1720:
Thanks @richarddavison  and @Sytten , I believe I have addressed all unresolved conversations. :)

richarddavison (approved):

PR #525, opened 2024-08-04
Rework time management & remove inter module dependency

Sytten:
### Description of changes

- Remove `os` dependency on `process`
- Remove `chrono` (using `SystemTime`)
- Remove `performance` (merged into `perf_hooks`)
- Rework time to be independent of `perf_hook`
- Switched `TIME_ORIGIN` to an `AtomicU64` and removed casting (ensuring i686 works correctly) 

### Checklist

### DIFF
--- a/llrt_modules/src/time.rs
+++ b/llrt_modules/src/time.rs
@@ -1,0 +1,26 @@
+use std::{
+    sync::atomic::{AtomicU64, Ordering},
+    time::SystemTime,
+};
+
+pub(crate) static TIME_ORIGIN: AtomicU64 = AtomicU64::new(0);
+
+/// Get the current time in nanoseconds.
+///
+/// # Safety
+/// - Good until the year 2554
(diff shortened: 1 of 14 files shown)


llrt_modules/src/time.rs
```diff
@@ -0,0 +1,26 @@
+use std::{
+    sync::atomic::{AtomicU64, Ordering},
+    time::SystemTime,
+};
+
```
richarddavison, line 5: Why can’t we keep usize?

Sytten: usize is a u32 on a 32bits system, while I dont claim support for those it still is not a good practice to convert between usize and u64. It wont make a difference on 64bits system.

richarddd: Sure. 32bit support is not prioritized anyway!

richarddavison (approved):

PR #534, opened 2024-08-10
chore: Correction of target version

nabetti1720:
### Description of changes

The target version has been revised as ES2023 is now supported in LLRT 0.2.0-beta.

### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,12 +43,6 @@ ENV LAMBDA_HANDLER "app.handler"
 CMD [ "llrt" ]
 ```
 
-That's it 🎉
-
-> [!IMPORTANT]
-> Even though LLRT supports [ES2020](https://262.ecma-international.org/11.0/) it's **NOT** a drop in replacement for Node.js. Consult [Compatibility matrix](#compatibility-matrix) and [API](API.md) for more details.
-> All dependencies should be bundled for a `browser` platform and mark included `@aws-sdk` packages as external.
-
 ### Option 4: AWS SAM
 
 The following [example project](example/llrt-sam/) sets up a lambda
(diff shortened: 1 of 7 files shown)


Sytten:
Is all of es2023 supported? I feel like the async iterator didnt work last time I tried.

nabetti1720:
> Is all of es2023 supported? I feel like the async iterator didnt work last time I tried.

We have not tried every feature of the ES2023 from one end to the other. Unfortunately, I do not have the skills to do so...
We felt it was acceptable to list ES2023 as supported for generally the following reasons.

1. ES2023 is mentioned for the first time in the [CHANGELOG](https://github.com/awslabs/llrt/commit/875ba405eee392296460dc5b0a1dcafd08c42544) of LLRT 0.2.0-beta (although it does not say "fully supported"...).

2. And this is what the QuickJS [website](https://bellard.org/quickjs/) says as well.
>> **Introduction**
>>
>> QuickJS is a small and embeddable Javascript engine. It supports the [ES2023](https://tc39.github.io/ecma262/2023) specification including modules, asynchronous generators, proxies and BigInt.

Looking at [test262.fyi](https://test262.fyi/#), it is true that QuickJS ES2023 compliance is not 100%, but then even V8 is 97%.

image

Especially since QuickJS says it supports ES2023, I thought we should separate this from the issue of individual functions not working. Unless it is truly fatal..

FYI:
There is a Web site called [runtime-compat](https://runtime-compat.unjs.io/). On the surface, the screen looks at WebAPI compliance, but the survey itself extends to Javascript build-in functionality.

As far as the status of LLRT's [survey](https://github.com/unjs/runtime-compat/blob/main/generator/runtimes/llrt/data.json), it appears to be compliant as the results of several asyncIterator surveys.

Do the latest test results show any difference?

richarddavison (approved):
Thank you

PR #536, opened 2024-08-11
Add ci support for windows

ahaoboy:
### Description of changes

Add ci support for windows

### Checklist

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -14,7 +14,9 @@ on:
       release:
         required: false
         type: string
-
+      toolchain:
+        required: true
+        type: string
 jobs:
   build:
     runs-on: ${{ inputs.os }}
@@ -41,13 +43,31 @@ jobs:
         run: |
           /bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/HEAD/install.sh)"
           brew install zig make
-      - name: Install Windows dependencies
+      - name: Install zig on windows
         if: inputs.platform == 'windows'
         shell: pwsh
         run: |
           Set-ExecutionPolicy -ExecutionPolicy RemoteSigned -Scope CurrentUser
           iex "& {$(irm get.scoop.sh)} -RunAsAdmin"
-          scoop install make zig
+          scoop install zig
+      - name: Install Windows dependencies
+        if: inputs.platform == 'windows'
+        uses: msys2/setup-msys2@v2
+        with:
+          msystem: MINGW64
+          update: true
+          path-type: inherit
+          install: >-
+            mingw-w64-x86_64-make
+            mingw-w64-x86_64-cmake
+            mingw-w64-x86_64-gcc
+            mingw-w64-x86_64-clang
+            mingw-w64-x86_64-perl
+            zip
+            zstd
+            patch
+            unzip
+
       - name: Install JavaScript dependencies
         run: |
           corepack enable
@@ -55,9 +75,9 @@ jobs:
       - name: Setup Rust
         uses: actions-rust-lang/setup-rust-toolchain@v1
         with:
-          toolchain: nightly
+          toolchain: ${{ inputs.toolchain }}
       - name: Run tests
-        if: inputs.arch != 'aarch64' || inputs.platform != 'linux'
+        if: (inputs.arch != 'aarch64' || inputs.platform != 'linux') && inputs.platform != 'windows'
         run: |
           make test-ci 2>&1
       - name: Run tests
@@ -83,6 +103,15 @@ jobs:
         if: inputs.release && inputs.platform == 'darwin'
         run: |
           make llrt-darwin-${{ inputs.release }}.zip
+      - name: Build Windows binaries
+        if: inputs.release && inputs.platform == 'windows'
+        shell: msys2 {0}
+        run: |
+          # HACK: Add scoop/shims to PATH
+          export PATH="$PATH:/c/Users/$USER/scoop/shims"
+          make stdlib
+          make libs-${{ inputs.release }}
+          make llrt-windows-${{ inputs.release }}.zip
       - name: Upload artifacts
         if: inputs.release
         uses: actions/upload-artifact@v4
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -39,23 +39,38 @@ jobs:
       fail-fast: ${{ startsWith(github.ref, 'refs/tags/') }}
       matrix:
         include:
+          - os: windows-latest
+            platform: windows
+            arch: x86_64
+            release: x64
+            toolchain: nightly-x86_64-pc-windows-gnu
           - os: ubuntu-latest
             platform: linux
             arch: x86_64
+            release: x64
+            toolchain: nightly
           - os: ubuntu-latest
             platform: linux
             arch: aarch64
+            release: arm64
+            toolchain: nightly
           - os: macos-latest
             platform: darwin
             arch: x86_64
+            release: x64
+            toolchain: nightly
           - os: macos-latest
             platform: darwin
             arch: aarch64
+            release: arm64
+            toolchain: nightly
     uses: ./.github/workflows/build.yml
     with:
       os: ${{ matrix.os }}
       platform: ${{ matrix.platform }}
       arch: ${{ matrix.arch }}
+      release: ${{ matrix.release }}
+      toolchain: ${{ matrix.toolchain }}
   modules:
     needs:
       - check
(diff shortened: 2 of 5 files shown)


ahaoboy:
Skipped the windows test for now, because there are too many failures to fix completely in a short period of time.

https://github.com/awslabs/llrt/actions/runs/10340140048/job/28620380619

.github/workflows/ci.yml
```diff
@@ -39,23 +39,31 @@ jobs:
...
+            toolchain: nightly-x86_64-pc-windows-gnu
           - os: ubuntu-latest
             platform: linux
             arch: x86_64
+            toolchain: nightly
           - os: ubuntu-latest
             platform: linux
             arch: aarch64
+            toolchain: nightly
           - os: macos-latest
```
nabetti1720, line 54: It appears that toolchain is not specified.

.github/workflows/ci.yml
```diff
@@ -39,23 +39,31 @@ jobs:
...
             platform: darwin
             arch: x86_64
```
nabetti1720, line 56: It appears that toolchain is not specified.

ahaoboy: The default is nightly, but it is not needed here because the specific toolchain is set in the makefile

nabetti1720: I see, you are indeed running `rustup` at the beginning of `Run Tests`.

Sytten:
Nice job!!

.github/workflows/ci.yml
ahaoboy: For other platforms, the default stable/nightly is OK. This is mainly for using the gnu toolchain on the windows platform to be able to build c related code

ahaoboy:
Although Windows still has some bugs, it can be used to test its performance. Here, llrt and tjs are used to measure the rendering events of a command line library. Since both use qjs, the difference is not obvious.

https://[ahaoboy.github.io/r-tui-bench/](https://ahaoboy.github.io/r-tui-bench/)

Sytten:
Ha yeah I see some of the errors, mostly filesystem and process related. I think the llrt_modules is pretty universal but the code seems to need to some help

richarddavison:
> Although Windows still has some bugs, it can be used to test its performance. Here, llrt and tjs are used to measure the rendering events of a command line library. Since both use qjs, the difference is not obvious.
> 
> https://[ahaoboy.github.io/r-tui-bench/](https://ahaoboy.github.io/r-tui-bench/)

This is awesome and very interesting. I suspect the increased (relative) performance in `tis` comes from `tjs` Uses `QuickJS-ng` and Bellards QuickJS. However these two repos will eventually converge and we'll see more activity and openness from this amazing project.

Makefile
```diff
@@ -7,10 +8,17 @@ TOOLCHAIN = +$(RUST_VERSION)
...
+ifeq ($(DETECTED_OS),windows)
+ZSTD_LIB_ARGS = -j lib-nomt UNAME=Windows ZSTD_LIB_COMPRESSION=0 ZSTD_LIB_DICTBUILDER=0 AR="zig ar"
+ZSTD_LIB_CC_ARGS = -s -O3 -flto
+ZSTD_LIB_CC_x64 = CC="zig cc -target x86_64-windows-gnu $(ZSTD_LIB_CC_ARGS)"
+else
 ZSTD_LIB_ARGS = -j lib-nomt UNAME=Linux ZSTD_LIB_COMPRESSION=0 ZSTD_LIB_DICTBUILDER=0 AR="zig ar"
 ZSTD_LIB_CC_ARGS = -s -O3 -flto
 ZSTD_LIB_CC_arm64 = CC="zig cc -target aarch64-linux-musl $(ZSTD_LIB_CC_ARGS)"
 ZSTD_LIB_CC_x64 = CC="zig cc -target x86_64-linux-musl $(ZSTD_LIB_CC_ARGS)"
+endif
```
richarddavison, line 21: This is a Lambda only optimization where we pack LLRT into a self extrating zstandard compressed executable. These args should not differ on windows as we should be able to build LLRT for Linux x64 and ARM64 on Windows (theoretically)

richarddavison (approved):
Thanks for this. Just a small comment. CI is also failing.

Makefile
ahaoboy: I am not familiar with zstd related compilation parameters. After testing, ZSTD_LIB_CC_x64 must be distinguished, otherwise ```error: the argument -femit-implib is allowed only when building a Windows DLL```

.github/workflows/build.yml
```diff
@@ -41,23 +43,40 @@ jobs:
         run: |
           /bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/HEAD/install.sh)"
           brew install zig make
+      # TODO: msys2 doesn't have zig pkg
+      - name: Install zig on windows
+        if: inputs.platform == 'windows'
+        uses: mlugg/setup-zig@v1
```
Sytten, line 49: I would really not use some random unofficial github action

ahaoboy:
> Ha yeah I see some of the errors, mostly filesystem and process related. I think the llrt_modules is pretty universal but the code seems to need to some help


Yes, there are a lot of tests that fail, mostly platform-specific path handling, file, and even some network library errors.

Sytten (changes requested):

.github/workflows/build.yml
ahaoboy: I haven't found a simple way to install zig. The zig package of msys2 cannot work. Or I can manually download zig through the command line and add it to the global environment variable, which is a bit troublesome.

ahaoboy:
It seems to work correctly ~
image

.github/workflows/build.yml
Sytten: Do we need msys2? If there is no alternative we can use a smaller GHA, but use a commit hash not a tag. And we need to make sure it doesnt do funny business.

ahaoboy: When there are c projects as dependencies, msys2 is currently the best solution, because projects like rquickjs rely on some linux tools, such as patch, perl, etc.
setup-msys2 does not seem to provide a version lock config, we can only trust it like we trust npm that it will not do funny business

nabetti1720:
Hi @ahaoboy , very nice work!!

By the way, does `make release-all` work without any errors at all on your local laptop now? You have added a `windows` condition to the Makefile this time and I am wondering if it is affected.

ahaoboy:
> By the way, does `make release-all` work without any errors at all on your local laptop now? You have added a `windows` condition to the Makefile this time and I am wondering if it is affected.

It doesn't work on windows, because windows cannot compiled to linux. It is best to use cross for multi-platform compilation.

nabetti1720:
> > By the way, does `make release-all` work without any errors at all on your local laptop now? You have added a `windows` condition to the Makefile this time and I am wondering if it is affected.
> 
> It doesn't work on windows, because windows cannot compiled to linux. It is best to use cross for multi-platform compilation.

I see. I don't know the result because I can't try it yet, but on the other hand, would it work on `linux` or `macos`?
Even if it works, I feel that I have already lost the motivation to keep `make release-all` as long as I can already generate artifacts for each architecture on GitHub.

.github/workflows/build.yml
richarddavison: This would install zig if running on powershell:
https://github.com/awslabs/llrt/blob/a80dfe08f2dd8071723cfa385757aa4958f3023c/.github/workflows/build.yml#L44

richarddavison (changes requested):

.github/workflows/build.yml
ahaoboy: Ok, we need to add one more environment variable, Let's see if it works
```
    export PATH="$PATH:/c/Users/$USER/scoop/shims"
```

ahaoboy:
There is a random socket test error

https://github.com/awslabs/llrt/actions/runs/10348831471/job/28641947417?pr=536

richarddavison:
> There is a random socket test error
> 
> https://github.com/awslabs/llrt/actions/runs/10348831471/job/28641947417?pr=536
> 
> https://github.com/awslabs/llrt/actions/runs/10352573355/job/28653525847

This feels like a race condition. Happens for other builds as well. I suspect we prematurely close the connection before the client receives response in a test. It’s a bit hard to reproduce as it only happens in CI.

ahaoboy:
> This feels like a race condition. Happens for other builds as well. I suspect we prematurely close the connection before the client receives response in a test. It’s a bit hard to reproduce as it only happens in CI.

This error will directly cause the program to exit, so it is not certain which test case failed.

richarddavison:
> > This feels like a race condition. Happens for other builds as well. I suspect we prematurely close the connection before the client receives response in a test. It’s a bit hard to reproduce as it only happens in CI.
> 
> This error will directly cause the program to exit, so it is not certain which test case failed.

Yes, thats why I’m refactoring the test runner to expose a server that spawns child processes to execute each test. That way we’ll know what test fails and can hide console output etc.

richarddavison:
@ahaoboy please rebase

richarddavison (approved):

PR #538, opened 2024-08-11
perf: Reduced copying of bytes from objects such as ArrayBuffer or TypedArrays

richarddavison:
### Description of changes

Introduces `ObjectBytes` util as a replacement for `get_bytes`. `ObjectBytes` is created from any JS value and owns that value as its true type (for example a `TypedArray` or `ArrayBuffer`). Since it owns its data it can also return a shared slice or Cow, leading to reduced copies for buffer based data types. This should increase performance and lower memory as less data is copied/duplicated/allocated.

### Checklist

### DIFF
--- a/llrt_core/src/modules/crypto/mod.rs
+++ b/llrt_core/src/modules/crypto/mod.rs
@@ -5,6 +5,7 @@ mod md5_hash;
 mod sha_hash;
 use std::slice;
 
+use llrt_utils::bytes::ObjectBytes;
 use once_cell::sync::Lazy;
 use rand::prelude::ThreadRng;
 use rand::Rng;
@@ -25,8 +26,7 @@ use crate::{
         module::export_default,
     },
     utils::{
-        class::get_class_name,
-        object::{bytes_to_typed_array, get_start_end_indexes, obj_to_array_buffer},
+        object::{bytes_to_typed_array, get_start_end_indexes},
         result::ResultExt,
     },
     vm::{CtxExtension, ErrorExtensions},
@@ -116,13 +116,16 @@ fn random_fill_sync<'js>(
 ) -> Result<Object<'js>> {
     let offset = offset.unwrap_or(0);
 
-    if let Some((array_buffer, source_length, source_offset)) = obj_to_array_buffer(&obj)? {
-        let (start, end) = get_start_end_indexes(source_length, size.0, offset);
-
+    if let Some(object_bytes) = ObjectBytes::from_array_buffer(&obj)? {
+        let (array_buffer, source_length, source_offset) =
+            object_bytes.get_array_buffer()?.unwrap();
         let raw = array_buffer
             .as_raw()
             .ok_or("ArrayBuffer is detached")
             .or_throw(&ctx)?;
+
+        let (start, end) = get_start_end_indexes(source_length, size.0, offset);
+
         let bytes = unsafe { slice::from_raw_parts_mut(raw.ptr.as_ptr(), source_length) };
 
         SYSTEM_RANDOM
@@ -134,7 +137,16 @@ fn random_fill_sync<'js>(
 }
 
 fn get_random_values<'js>(ctx: Ctx<'js>, obj: Object<'js>) -> Result<Object<'js>> {
-    if let Some((array_buffer, source_length, source_offset)) = obj_to_array_buffer(&obj)? {
+    if let Some(object_bytes) = ObjectBytes::from_array_buffer(&obj)? {
+        if matches!(
+            object_bytes,
+            ObjectBytes::F64Array(_) | ObjectBytes::F32Array(_)
+        ) {
+            return Err(Exception::throw_message(&ctx, "Unsupported TypedArray"));
+        }
+
+        let (array_buffer, source_length, source_offset) =
+            object_bytes.get_array_buffer()?.unwrap();
         let raw = array_buffer
             .as_raw()
             .ok_or("ArrayBuffer is detached")
@@ -151,13 +163,7 @@ fn get_random_values<'js>(ctx: Ctx<'js>, obj: Object<'js>) -> Result<Object<'js>
             std::slice::from_raw_parts_mut(raw.ptr.as_ptr().add(source_offset), source_length)
         };
 
-        match get_class_name(&obj)?.unwrap().as_str() {
-            "Int8Array" | "Uint8Array" | "Uint8ClampedArray" | "Int16Array" | "Uint16Array"
-            | "Int32Array" | "Uint32Array" | "BigInt64Array" | "BigUint64Array" => {
-                SYSTEM_RANDOM.fill(bytes).unwrap()
-            },
-            _ => return Err(Exception::throw_message(&ctx, "Unsupported TypedArray")),
-        }
+        SYSTEM_RANDOM.fill(bytes).unwrap()
     }
 
     Ok(obj)
--- a/llrt_core/src/modules/crypto/mod.rs
+++ b/llrt_core/src/modules/crypto/mod.rs
@@ -78,6 +78,9 @@ fn get_random_int(_ctx: Ctx, first: i64, second: Opt<i64>) -> Result<i64> {
     Ok(random_number)
 }
 
+const NOT_ARRAY_BUFFER_ERROR: &'static str = "Not an ArrayBuffer";
+const ARRAY_BUFFER_DETACHED_ERROR: &'static str = "ArrayBuffer is detached";
+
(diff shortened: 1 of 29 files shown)


Sytten:
Do we want to merge that with the ArrayBufferView? Apart from string they do a similar job. V8 has a complex hierarchy of objects that can have bytes in them.
If not I would comment on when each should be used.

llrt_utils/src/bytes.rs
```diff
@@ -1,8 +1,247 @@
...
+        Err(Exception::throw_message(
+        ctx,
+        "value must be typed DataView, Buffer, ArrayBuffer, Uint8Array or interpretable as string",
+    ))
+    }
+
+    pub fn get_bytes(&mut self) -> Cow<'_, [u8]> {
+        let cow = match self {
+            ObjectBytes::U8Array(array) => {
+                Cow::Borrowed(array.as_bytes().expect(Self::DETACHED_ERROR))
```
Sytten, line 66: I would really prefer to return a `Result` since in my context if a user plugin script is screwing up, I don't want to crash my whole process. Though the buffer should not be detached since we hold a reference.

llrt_utils/src/bytes.rs
```diff
@@ -1,8 +1,247 @@
...
+    pub fn from_offset(
+        ctx: &Ctx<'js>,
+        value: Value<'js>,
+        offset: usize,
+        length: Option<usize>,
+    ) -> Result<Self> {
+        if value.is_undefined() {
+            return Ok(ObjectBytes::Vec(Some(vec![])));
+        }
+        if let Some(bytes) = get_string_bytes(&value, offset, length)? {
```
Sytten, line 40: A lot of the current llrt APIs are not fully compliant with node because we miss the encoding part for strings. If you want to bridge that gap we will need to have another input for that function

llrt_utils/src/encoding.rs
```diff
@@ -71,11 +71,11 @@ impl Encoder {
         }
     }
 
-    pub fn decode(&self, bytes: Vec<u8>) -> Result<Vec<u8>, String> {
+    pub fn decode(&self, bytes: &[u8]) -> Result<Vec<u8>, String> {
```
Sytten, line 74: I would use `fn decode<B: Into<Cow<'_, [u8]>>>`(&self, bytes: B`, that will save you a clone on the third branch if you pass in a Vec. AFAIK it is transparent for the caller.

llrt_modules/src/modules/buffer.rs
```diff
@@ -61,7 +61,7 @@ impl<'js> Buffer {
     ) -> Result<Value<'js>> {
         if let Some(encoding) = encoding {
             let encoder = Encoder::from_str(&encoding).or_throw(ctx)?;
-            bytes = encoder.decode(bytes).or_throw(ctx)?;
+            bytes = encoder.decode(&bytes).or_throw(ctx)?;
```
Sytten, line 64: Dont add this if you follow my trait Into cow

llrt_modules/src/modules/fs/write_file.rs
```diff
@@ -1,6 +1,6 @@
+use llrt_utils::bytes::ObjectBytes;
```
Sytten, line 1: Should be after the license

llrt_utils/src/bytes.rs
```diff
@@ -1,8 +1,247 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
+
```
Sytten, line 3: Nit space, might want a linter rule for imports. They are available in unstable fmt

llrt_utils/src/bytes.rs
```diff
@@ -1,8 +1,247 @@
...
+            ObjectBytes::F32Array(array) => {
+                Cow::Borrowed(array.as_bytes().expect(Self::DETACHED_ERROR))
+            },
+            ObjectBytes::F64Array(array) => {
+                Cow::Borrowed(array.as_bytes().expect(Self::DETACHED_ERROR))
+            },
+            ObjectBytes::DataView(array_buffer) => {
+                Cow::Borrowed(array_buffer.as_bytes().expect(Self::DETACHED_ERROR))
+            },
+            ObjectBytes::Vec(bytes) => Cow::Owned(bytes.take().expect("Bytes already taken")),
```
Sytten, line 98: I feel like we could always borrow in this method and get rid of the cow. Currently it is a mix of both which will be confusing since there is no guarantee that a second call will return data.

Sytten (commented):
This is a good improvement!!
I have opinions on stuff but you know better :)

llrt_utils/src/bytes.rs
richarddavison: I was considering this myself and you have a valid point. We should not be able to detach a buffer unless we do some really strange stuff. Also there are methods in `rquickjs` that also panics if buffer is detected so leaving this as is for now 🙂

llrt_utils/src/bytes.rs
richarddavison: I think most does. This utility does not handle encodings as it’s a lower level utility just to get the raw bytes. As you can see in the encoder related structs we’re transforming the bytes with respect to their encoding property

llrt_utils/src/bytes.rs
richarddavison: It’s for performance reasons. If we only borrow we need an additional clone to get get the underlying Vec. Agree it’s a bit strange that calling get_bytes() a second time would panic if already called but I couldn’t think of another way. Suggestions are welcome!

Sytten: Yes I see we have two use cases. So lets do two functions.
- When you are passing the return value has a reference to say `.write` use the `as_bytes` method which takes `&self` and returns a slice.
- When you are using an owned version say in the hasher, use `.into_bytes` which consumes `self` and returns a `vec`.

llrt_core/src/modules/crypto/mod.rs
```diff
@@ -134,7 +137,16 @@ fn random_fill_sync<'js>(
...
+    if let Some(object_bytes) = ObjectBytes::from_array_buffer(&obj)? {
+        if matches!(
+            object_bytes,
+            ObjectBytes::F64Array(_) | ObjectBytes::F32Array(_)
+        ) {
+            return Err(Exception::throw_message(&ctx, "Unsupported TypedArray"));
+        }
+
+        let (array_buffer, source_length, source_offset) =
+            object_bytes.get_array_buffer()?.unwrap();
```
Sytten, line 149: Not a fan of unwrap even if this is safe right now. Just takes a refactor to making it crash. At minimum lets use an expect or better turn that into a result.

llrt_core/src/modules/http/fetch.rs
```diff
@@ -282,8 +283,11 @@ fn get_fetch_options<'js>(
         if let Some(body_opt) =
             get_option::<Value>("body", arg_opts.as_ref(), resource_opts.as_ref())?
         {
-            let bytes = get_bytes(ctx, body_opt)?;
-            body = Some(Full::from(bytes));
+            let mut bytes = ObjectBytes::from(ctx, &body_opt)?;
+            let bytes = bytes.get_bytes();
+            //this is fine since we're only holding on to bytes in this function
```
Sytten, line 288: This is not true though because Bytes will hold onto the static slice forever. It wont copy it. So it will escape this function and then we could end up with a use after free problem since we will drop the Value for the ArrayBuffer.

Sytten (changes requested):

llrt_core/src/modules/http/fetch.rs
Sytten: I contributed to the Bytes crate so I kinda know the internals of it, but essentially if you see this:
https://github.com/tokio-rs/bytes/blob/master/src/bytes.rs#L166

It uses a custom vtable and it will store a pointer to the slice. If there are clones of it then it changes it to a an Arc-style system but still wont copy it.

llrt_core/src/modules/crypto/mod.rs
```diff
@@ -78,6 +78,9 @@ fn get_random_int(_ctx: Ctx, first: i64, second: Opt<i64>) -> Result<i64> {
     Ok(random_number)
 }
 
+const NOT_ARRAY_BUFFER_ERROR: &'static str = "Not an ArrayBuffer";
+const ARRAY_BUFFER_DETACHED_ERROR: &'static str = "ArrayBuffer is detached";
```
Sytten: That we could put somewhere to be reused

Sytten (approved):

PR #543, opened 2024-08-14
refactor: move crypto to llrt_modules

nabetti1720:
### Description of changes

Moving crypto ~~and llrt~~ to llrt_modules. ~~However, the creation of types is just the beginning...~~

### Checklist


Sytten:
Could use a different naming than `llrt` even if it is importable by default as `llrt:something`. Same idea that we dont have a module called node. Instead we could use the tag system I proposed to mark the source of those APIs.

nabetti1720:
> Could use a different naming than `llrt` even if it is importable by default as `llrt:something`. Same idea that we dont have a module called node. Instead we could use the tag system I proposed to mark the source of those APIs.

Hi @Sytten , Thank you for your comment.

What other uses exist besides importing with the designation `llrt:something`? Also, for your proposed tag system, if it is already in use somewhere in this project, could you please tell me where?

Sorry, I don't have English and technical skills and would appreciate your time and tips to help me understand your thoughts. :)

Sytten:
I will speak with @richarddavison about it, dont worry about it. I would split the PR to port crypto first since there is not debate around that one and lets postpone the llrt module transfer.

nabetti1720:
> I will speak with @richarddavison about it, dont worry about it. I would split the PR to port crypto first since there is not debate around that one and lets postpone the llrt module transfer.

Thank you for your consideration. There was a place in the crypto module that used llrt:uuid, so I had no choice but to move it at the same time.

nabetti1720:
I wanted to clean up this PR quickly, so I modified the move to llrt_modules to only use the crypto module. It was only a small part of the crypto module's dependency on llrt:uuid, so we decided to use the uuid crate directly.

richarddavison (approved):
Thanks LGTM

PR #547, opened 2024-08-16
feat: Multi-flavor support for LLRT artifacts in AWS Lambda

### LINKED ISSUE #39: Full SDK Binaries
In order to keep the binary size to a minimum only a subset clients of the full AWS SDK v3 is bundled currently.
Multiple binaries

We could via feature flags build multiple flavors of LLRT:

- llrt-full - The full SDK bundled
- llrt-min - No SDK bundled
- llrt - The default version (current) with the most popular packages bundled

Futhermore, we could allow customers to download the bundled and optimized SDKs as .lrt files. .lrt files is zstd compressed QuickJS bytecode that can be imported in llrt (already supported)

Size

L

nabetti1720:
### Issue # (if available)

Closed #39 

### Description of changes

This PR is to provide further support for `@aws-sdk`.

- Since the API Endpoint management was already starting to get confusing, we will incorporate the following website categorization.
AWS services by category : https://docs.aws.amazon.com/whitepapers/latest/aws-overview/amazon-web-services-cloud-platform.html
- The packages added this time were selected from the top of those sorted by Popularity in `npmjs`.
- We have left a comment on how to define it in case there are new SDKs that we would like to support. 
- We have also organized the SDKs covered by the bundle in the README.
- The fact that the added constants in `build.mjs` are in camelCase is a bit awkward, but please forgive us as it is the result of our priority on clarity.

Addendum 1:
- The environment variable `SDK_BUNDLE_MODE` is now supported. This is a milestone for future releases of binaries with different bundle targets. When set to `STD`, the SDKs that have been supported up to now will be bundled. In case of `FULL`, all SDKs added this time will be bundled.

Addendum 2:
- Modify Makefile to support the following three binary creation patterns for lambda.
  1. `make release-[x64|arm64]-full-sdk` : The program is executed with `SDK_BUNDLE_MODE` set to `FULL`.
  2. `make release-[x64|arm64]` : The program is executed with `SDK_BUNDLE_MODE` set to `STD`.
  3. `make release-[x64|arm64]-no-sdk` : `SDK_BUNDLE_MODE` is executed other than the above. ~~In addition, `--features no-sdk` is specified when running `cargo build`.~~ If you specify `--features no-sdk` in `cargo build`, `@aws-sdk` etc. will not be bundled, but test code will be bundled. Bundling of `@aws-sdk` etc. is controlled by `SDK_BUNDLE_MODE`, so we will leave `--features lambda` as it is.

~~After the separate Windows CI support is merged, we will work on integrating it into GitHub Actions (in a next PR).~~
This PR even supports GithubActions.

### Checklist


nabetti1720:
Now that we have made a break, we would appreciate your review. :)

nabetti1720:
Now that we have made a break, we would appreciate your review. :)

build.mjs
```diff
@@ -75,32 +75,208 @@ const ES_BUILD_OPTIONS = {
...
+  ...FULL({ "client-sso-admin": ["SSOAdmin", ["sso", "identitystore"]] }),
+  ...FULL({ "client-sso-oidc": ["SSOOIDC", ["sso", "identitystore"]] }),
+  ...STD({ "client-sts": ["STS", ["sts"]] }),
+};
+const _Storage = {
+  ...FULL({ "client-efs": ["EFS", ["elasticfilesystem"]] }),
+  ...STD({ "client-s3": ["S3", ["s3"]] }),
+  ...STD({ "lib-storage": ["Upload", ["s3"]] }),
+};
+
```
richarddavison, line 251: This seems overly complex. Since we have a hierarchical structure i would structure the data differently.
We can keep the original `SDK_DATA` and set an optional boolean to indicate if it's `full` or not `std`. We only need to distinguish included SDKs between `full` and `std` since `no-sdk` contains none of these:
```javascript
//format: {CLIENT_IMPORT_NAME}: [{SDK_NAME},{ENDPOINT_NAME},{FULL_SDK_ONLY}]
const SDK_DATA = {
  "client-eventbridge": ["EventBridge", "events", true],
  "client-dynamodb": ["DynamoDB", "dynamodb"],
  ,...
  }
```

We can also distinguish them via comments rather than separate variables:
```javascript
// Application integration
...
//Blockchain
...
//Compute services


```

richarddavison (changes requested):
Thanks for the PR! This is great. Should we also add `full-sdk` binaries to automated releases?

build.mjs
nabetti1720: > This seems overly complex. Since we have a hierarchical structure i would structure the data differently.
> We can keep the original `SDK_DATA` and set an optional boolean to indicate if it's `full` or not `std`. We only need to distinguish included SDKs between `full` and `std` since `no-sdk` contains none of these:

You pointed out that by using `cargo --features no-sdk`, we could just use the `full` or `std` binary decision, but I don't think that would work. At this time, we are `using cargo --features lambda` for binary generation for lambda, even in the case of `no-sdk`. This is because we do not want to bundle test code.

image

By reviewing the above process, we could exclude the test code in the `no-sdk` case, but this time it would affect the binary for linux/darwin (llrt test would not work)...

Therefore, at the present time, it is desirable to be able to make a three-value judgment here: no | std | full.

> We can also distinguish them via comments rather than separate variables:

Your point here makes sense! I will review it immediately. :)

richarddavison: I think I wasn't expressing myself clear :) We have 3 "flavors" when building the JS like you described:
 no | std | full.
 However we only need to distinguish between std & full when building the `SDK_DATA` list. If we have `no` we simply ignore the values `SDK_DATA` in the list.

Makefile
```diff
@@ -51,48 +51,72 @@ release-linux: | lambda-all llrt-linux-x64.zip llrt-linux-arm64.zip
...
+	cargo $$(BUILD_ARG) --target $$(TARGET_linux_$$(RELEASE_ARCH_NAME_${1})) --features lambda
+	./pack target/$$(TARGET_linux_$$(RELEASE_ARCH_NAME_${1}))/release/llrt target/$$(TARGET_linux_$$(RELEASE_ARCH_NAME_${1}))/release/bootstrap
+	@rm -rf llrt-lambda-${1}-full-sdk.zip
+	zip -j llrt-lambda-${1}-full-sdk.zip target/$$(TARGET_linux_$$(RELEASE_ARCH_NAME_${1}))/release/bootstrap
+	cargo $$(BUILD_ARG) --target $$(TARGET_linux_$$(RELEASE_ARCH_NAME_${1})) --features lambda,uncompressed
+	mv target/$$(TARGET_linux_$$(RELEASE_ARCH_NAME_${1}))/release/llrt llrt-container-${1}-full-sdk
+
+llrt-lambda-${1}: export SDK_BUNDLE_MODE = STD
+llrt-lambda-${1}: | clean-js js
+	cargo $$(BUILD_ARG) --target $$(TARGET_linux_$$(RELEASE_ARCH_NAME_${1})) --features lambda
```
richarddavison, line 88: I think we can reduce a lot of code by adding more variables to the template.
```
define release_template
release-${1}-${2}
```

build.mjs
```diff
@@ -75,29 +76,160 @@ const ES_BUILD_OPTIONS = {
...
+  ...{},
+  // ComputeServices
+  "client-auto-scaling": ["AutoScaling", ["autoscaling"], true],
+  "client-batch": ["Batch", ["batch"], true],
+  "client-ec2": ["EC2", ["ec2"], true],
+  "client-lambda": ["Lambda", ["lambda"], false],
+  // CustomerEnablement
+  ...{},
+  // Containers
+  "client-ecr": ["ECR", ["ecr", "api.ecr"], true],
```
richarddavison, line 151: Why is there two arrays here? `["ecr", "api.ecr"]`

richarddavison (commented):
Almost ready!

build.mjs
nabetti1720: The definition is based on the description in the following URL.

https://docs.aws.amazon.com/general/latest/gr/ecr.html

> The ecr and api.ecr endpoints are used for calls to the Amazon ECR API. API actions such as DescribeImages and CreateRepository go to this endpoint. While the two endpoints function the same, the api.ecr endpoint is recommended and the default when using the AWS CLI or AWS SDKs. When connecting to Amazon ECR through an AWS PrivateLink VPC endpoint, you must use the api.ecr endpoint to make API calls.

Makefile
nabetti1720: Please confirm that you have made the correction.

build.mjs
nabetti1720: This looks like it could be unified with `api.ecr`.

There seems to be something else.
https://docs.aws.amazon.com/general/latest/gr/step-functions.html

> client-sfn": ["SFN", ["states", "sync-states"], false],

richarddavison (approved):

PR #549, opened 2024-08-20
Use unchecked_unwrap where safe

richarddavison:
### Description of changes

Uses `unchecked_unwrap` where safe to do. Also improves some error handling.

### Checklist


Sytten:
I usually prefer an expect with a message so we know when this "safe" turned out to not be safe where to look 😅

richarddavison:
> I usually prefer an expect with a message so we know when this "safe" turned out to not be safe where to look 😅

True, but in this case unsafe is used to add more compiler optimizations. `expect` and `unwrap` has the same complexity while `unwrap_unchecked` saves a branch.

PR #550, opened 2024-08-22
perf: Dynamic timer sleep

richarddavison:
### Description of changes

Use dynamic sleep instead of fixed for timer & interval sleeps. This significantly reduces wakes in the async scheduler since timers are very likely to be above 4ms (previous min value).

### Checklist

### DIFF
--- a/llrt_core/src/modules/timers.rs
+++ b/llrt_core/src/modules/timers.rs
@@ -1,12 +1,14 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 use std::{
+    pin::{pin, Pin},
     ptr::NonNull,
+    rc::Rc,
     sync::{
-        atomic::{AtomicBool, AtomicUsize, Ordering},
-        Mutex,
+        atomic::{AtomicUsize, Ordering},
+        Mutex, MutexGuard,
     },
-    time::{Duration, SystemTime, UNIX_EPOCH},
+    time::Duration,
 };
 
 use once_cell::sync::Lazy;
@@ -15,18 +17,60 @@ use rquickjs::{
     prelude::{Func, Opt},
     qjs, CatchResultExt, Ctx, Function, Persistent, Result,
 };
+use tokio::{
+    select,
+    sync::Notify,
+    time::{Instant, Sleep},
+};
 
 use crate::{module_builder::ModuleInfo, modules::module::export_default, vm::Vm};
 
 static TIMER_ID: AtomicUsize = AtomicUsize::new(0);
-static TIME_POLL_ACTIVE: AtomicBool = AtomicBool::new(false);
-pub struct TimeoutRef {
+static RT_TIMER_STATE: Lazy<Mutex<Vec<RuntimeTimerState>>> = Lazy::new(|| Mutex::new(Vec::new()));
+
+pub struct RuntimeTimerState {
+    timers: Vec<Timeout>,
+    rt: *mut qjs::JSRuntime,
+    running: bool,
+    deadline: Instant,
+    notify: Rc<Notify>,
+}
+impl RuntimeTimerState {
+    fn new(rt: *mut qjs::JSRuntime) -> Self {
+        let deadline = Instant::now() + Duration::from_secs(86400 * 365 * 30);
+        Self {
+            timers: Default::default(),
+            rt,
+            deadline,
+            running: false,
+            notify: Default::default(),
+        }
+    }
+}
+
+unsafe impl Send for RuntimeTimerState {}
+
+#[derive(Clone)]
+pub struct Timeout {
     callback: Option<Persistent<Function<'static>>>,
-    pub expires: usize,
-    ctx: NonNull<qjs::JSContext>,
+    deadline: Instant,
+    raw_ctx: NonNull<qjs::JSContext>,
     id: usize,
     repeating: bool,
-    delay: usize,
+    interval: u64,
+}
+
+impl Default for Timeout {
+    fn default() -> Self {
+        Self {
+            callback: None,
+            deadline: Instant::now(),
+            raw_ctx: NonNull::dangling(),
+            id: 0,
+            repeating: false,
+            interval: 0,
+        }
+    }
 }
 
 fn set_immediate(cb: Function) -> Result<()> {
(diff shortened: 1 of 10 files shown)


llrt_core/src/modules/timers.rs
```diff
@@ -15,103 +17,143 @@ use rquickjs::{
...
-    create_spawn_loop(ctx, rt)?;
 
     Ok(id)
 }
 
+fn get_timer_state<'a>(
+    state_ref: &'a mut MutexGuard<Vec<RuntimeTimerState>>,
+    rt: *mut qjs::JSRuntime,
+) -> &'a mut RuntimeTimerState {
+    let rt_timers = state_ref.iter_mut().find(|state| state.rt == rt);
```
Sytten, line 135: Thats ok for now but we should switch to the new data method to store the state on the runtime itself once that is released. It will be a lot cleaner.

llrt_core/src/modules/timers.rs
```diff
@@ -15,103 +17,143 @@ use rquickjs::{
...
 use crate::{module_builder::ModuleInfo, modules::module::export_default, vm::Vm};
 
 static TIMER_ID: AtomicUsize = AtomicUsize::new(0);
-static TIME_POLL_ACTIVE: AtomicBool = AtomicBool::new(false);
-pub struct TimeoutRef {
+static RT_TIMER_STATE: Lazy<Mutex<Vec<RuntimeTimerState>>> = Lazy::new(|| Mutex::new(Vec::new()));
+
+pub struct RuntimeTimerState {
+    timers: Vec<Timeout>,
+    rt: *mut qjs::JSRuntime,
```
Sytten, line 33: Since this is runtime agnostic do you want to move it to modules? I can also do it for you later.

llrt_core/src/modules/timers.rs
```diff
@@ -15,103 +17,143 @@ use rquickjs::{
...
+    }
+    drop(rt_timer);
+
+    if !task_running {
+        create_spawn_loop(
+            rt_ptr,
+            ctx,
+            unsafe { abort_timer_rx.unwrap_unchecked() },
+            deadline,
+        )?;
```
Sytten, line 125: Move that in the else clause, you will avoid this unwrap. Duplicate the drop of rt worst case.

llrt_core/src/modules/timers.rs
```diff
@@ -15,103 +17,143 @@ use rquickjs::{
...
     Ok(id)
 }
 
+fn get_timer_state<'a>(
+    state_ref: &'a mut MutexGuard<Vec<RuntimeTimerState>>,
+    rt: *mut qjs::JSRuntime,
+) -> &'a mut RuntimeTimerState {
+    let rt_timers = state_ref.iter_mut().find(|state| state.rt == rt);
+
+    (unsafe { rt_timers.unwrap_unchecked() }) as _
```
Sytten, line 137: Thats ugly 😅
I dont think an expect would be too bad here

llrt_core/src/modules/timers.rs
```diff
@@ -15,103 +17,143 @@ use rquickjs::{
...
-        if let Some(timeout) = state.timers.iter_mut().find(|t| t.id == id) {
-            if let Some(timeout) = timeout.callback.take() {
-                timeout.restore(&ctx)?; //prevent memory leaks
-            }
-            timeout.expires = 0;
-            timeout.repeating = false;
+    let state = get_timer_state(&mut rt_timers, rt);
+    if let Some(timeout) = state.timers.iter_mut().find(|t| t.id == id) {
+        if let Some(timeout) = timeout.callback.take() {
+            timeout.restore(&ctx)?; //prevent memory leaks
```
Sytten, line 147: A drop is sufficient, restore is just some fancy lifetime changes. The drop trait of Value is what decreases the ref count.

llrt_core/src/modules/timers.rs
```diff
@@ -15,103 +17,143 @@ use rquickjs::{
...
         }
+        timeout.repeating = false;
+        timeout.deadline = Instant::now() - Duration::from_secs(1);
+        state.notify.notify_one()
```
Sytten, line 151: Why not just removing the timeout? Would warrant a comment since its not obvious why the task should be awoken.

llrt_core/src/vm.rs
```diff
@@ -630,14 +631,19 @@ fn init(ctx: &Ctx<'_>, module_names: HashSet<&'static str>) -> Result<()> {
...
+            let mut executing_timers = Vec::new();
 
             let imported_object = loop {
                 if let Some(x) = import_promise.result::<Object>() {
                     break x?;
                 }
 
-                poll_timers(rt, &mut executing_timers, &mut false);
+                if deadline < Instant::now() {
+                    poll_timers(rt, &mut executing_timers, None, Some(&mut deadline));
```
Sytten, line 644: Is that not conflicting with your background task? Why is it still needed?

Sytten (commented):
Good stuff!

llrt_core/src/modules/timers.rs
richarddavison: There is a simple reason for this. Say we have code like this:
```javascript
const timeout = setTimeout(() => {
    console.log("hello");
},30*1000)

clearTimeout(timeout)
```
Without the wake we're idling for 30 seconds before exiting even though we have nothing left to do

llrt_core/src/vm.rs
richarddavison: It's not conflicting because require is synchronous and might load a file containing something like:
```javascript
await new Promise((res) => setTimeout(res, 0));
```
Which never will be resolved since we're not polling the futures yet which happens in here: https://github.com/awslabs/llrt/blob/6cbad7193b68e0c01fc893cd357d84ea0a243215/llrt_core/src/vm.rs#L489

llrt_core/src/modules/timers.rs
```diff
@@ -15,103 +17,143 @@ use rquickjs::{
...
+pub struct RuntimeTimerState {
+    timers: Vec<Timeout>,
+    rt: *mut qjs::JSRuntime,
+    running: bool,
+    deadline: Instant,
+    notify: Rc<Notify>,
+}
+impl RuntimeTimerState {
+    fn new(rt: *mut qjs::JSRuntime) -> Self {
+        let deadline = Instant::now() + Duration::from_secs(86400 * 365 * 30);
```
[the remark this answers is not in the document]
richarddavison: This value comes from https://docs.rs/tokio/latest/src/tokio/time/instant.rs.html#57. Duration::MAX will overflow since its added to the instant

PR #558, opened 2024-08-28
refactor: Improvement of ObjectBytes by implementing FromJs

nabetti1720:
### Description of changes

- We organized the existing code by implementing `FromJs` for `ObjectBytes`.

### Checklist

### DIFF
--- a/llrt_modules/src/modules/fs/write_file.rs
+++ b/llrt_modules/src/modules/fs/write_file.rs
@@ -2,18 +2,17 @@
 // SPDX-License-Identifier: Apache-2.0
 
 use llrt_utils::{bytes::ObjectBytes, result::ResultExt};
-use rquickjs::{Ctx, Result, Value};
+use rquickjs::{Ctx, Result};
 use tokio::fs;
 use tokio::io::AsyncWriteExt;
 
-pub async fn write_file<'js>(ctx: Ctx<'js>, path: String, data: Value<'js>) -> Result<()> {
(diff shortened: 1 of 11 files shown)


llrt_modules/src/modules/fs/write_file.rs
```diff
@@ -2,18 +2,17 @@
...
     let mut file = fs::File::create(&path)
         .await
         .or_throw_msg(&ctx, &["Can't create file \"", &path, "\""].concat())?;
 
     let write_error_message = &["Can't write file \"", &path, "\""].concat();
 
-    let bytes = ObjectBytes::from(&ctx, &data)?;
     file.write_all(bytes.as_bytes())
         .await
         .or_throw_msg(&ctx, write_error_message)?;
```
richarddavison, line 18: Here is an optimization not to accept ObjectBytes<'js> as we check if we can create the file first and then do the ObjectBytes calculation.  Also I saw now that we can reuse:
`let write_error_message = &["Can't write file \"", &path, "\""].concat();`

richarddavison (changes requested):
Thanks, this is very convenient! Just a minor nitpick

llrt_modules/src/modules/fs/write_file.rs
nabetti1720: Understood. And the renovations are complete.

richarddavison (approved):

PR #562, opened 2024-08-29
Add typing for timers

Sytten:
### Description of changes

Add typescript for timers. Added them to global since this is like expected to be there.

### Checklist

### DIFF
--- a/types/timers.d.ts
+++ b/types/timers.d.ts
@@ -16,11 +16,9 @@ declare module "timers" {
      * Schedules execution of a one-time `callback` after `delay` milliseconds.
      *
      * The `callback` will likely not be invoked in precisely `delay` milliseconds.
-     * Caido makes no guarantees about the exact timing of when callbacks will fire,
+     * LLRT makes no guarantees about the exact timing of when callbacks will fire,
      * nor of their ordering. The callback will be called as close as possible to the
-     * time specified.
-     *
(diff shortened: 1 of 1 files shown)


types/timers.d.ts
```diff
@@ -0,0 +1,71 @@
...
+
+    /**
+     * Schedules execution of a one-time `callback` after `delay` milliseconds.
+     *
+     * The `callback` will likely not be invoked in precisely `delay` milliseconds.
+     * Caido makes no guarantees about the exact timing of when callbacks will fire,
+     * nor of their ordering. The callback will be called as close as possible to the
+     * time specified.
+     *
+     * When `delay` is less than `4`, the `delay` will be set to `4`.
```
Sytten, line 23: This is not true in the current code, but I feel like it should? @richarddavison thoughts

richarddavison: Not really. It's a bit more complex than that. If one timer has been executed and the next one is scheduled to be executed in 1-3ms it will be rounded up to 4ms. If it's scheduled time already has passed it will be executed immediately. We can write something like "the precision of timers are 4ms". This is ofc something we in the future can override or set via an env if needed to. But it feels pretty OK for now

Sytten: Will do sorry been very busy

richarddavison (approved):

PR #567, opened 2024-09-01
feat(fetch): support data: URLs

### LINKED ISSUE #565: fetch support octet-stream;base64
This is not a valid http url, but in some scenarios it is convenient to use this form to embed binary files in the program.

```js
const s = "hello"
const base64 = Buffer.from(s).toString("base64");
const dataURIPrefix = "data:application/octet-stream;base64,";
const url = dataURIPrefix + base64;
const resp = await fetch(url);
const buf = await resp.arrayBuffer();
const str = Buffer.from(buf).toString("ascii");
console.log(str === s);
```

nabetti1720:
### Issue # (if available)

Closed #565

### Description of changes

Support for `data: urls`(https://fetch.spec.whatwg.org/#data-urls) has been added.

Sample 1:
```javascript
// data-url1.js
const s = "hello"
const base64 = Buffer.from(s).toString("base64");
const dataURIPrefix = "data:application/octet-stream;base64,";
const url = dataURIPrefix + base64;
const resp = await fetch(url);
const buf = await resp.arrayBuffer();
const str = Buffer.from(buf).toString("ascii");
console.log(str === s);
```
```shell
% llrt data-url1.js
true
```

Sample 2: The source of the base64 string is, https://www.rfc-editor.org/rfc/rfc2397
```javascript
// data-url2.js
import fs from 'fs';
const imgUrl = "data:image/gif;base64,R0lGODdhMAAwAPAAAAAAAP///ywAAAAAMAAwAAAC8IyPqcvt3wCcDkiLc7C0qwyGHhSWpjQu5yqmCYsapyuvUUlvONmOZtfzgFzByTB10QgxOR0TqBQejhRNzOfkVJ+5YiUqrXF5Y5lKh/DeuNcP5yLWGsEbtLiOSpa/TPg7JpJHxyendzWTBfX0cxOnKPjgBzi4diinWGdkF8kjdfnycQZXZeYGejmJlZeGl9i2icVqaNVailT6F5iJ90m6mvuTS4OK05M0vDk0Q4XUtwvKOzrcd3iq9uisF81M1OIcR7lEewwcLp7tuNNkM3uNna3F2JQFo97Vriy/Xl4/f1cf5VWzXyym7PHhhx4dbgYKAAA7";
const res = await fetch(imgUrl);
const imgData = await res.blob();
console.log(imgData);
fs.writeFileSync('Larry.gif', await imgData.arrayBuffer());
```
```shell
% llrt data-url2.js
Blob {
  size: 273,
  type: 'image/gif'
}
```

image

### Checklist

### DIFF
--- a/llrt_core/src/modules/http/fetch.rs
+++ b/llrt_core/src/modules/http/fetch.rs
@@ -139,24 +139,26 @@ fn parse_data_url<'js>(ctx: &Ctx<'js>, data_url: &str) -> Result<Response<'js>> 
         .split_once(',')
         .ok_or_else(|| Exception::throw_type(ctx, "Invalid data URL format"))?;
 
-    let (is_base64, mime_type) = if mime_type.contains("base64") {
-        (
-            true,
-            mime_type
-                .trim()
-                .trim_end_matches("base64")
-                .trim_end_matches(|c| c == ' ' || c == ';'),
-        )
-    } else {
-        (false, mime_type.trim())
-    };
+    let mut is_base64 = false;
+    let mut content_type = String::with_capacity(10);
+    for (i, part) in mime_type.split(";").enumerate() {
+        let part = part.trim();
+        if i == 1 || i == 2 {
+            if part == "base64" {
+                is_base64 = true;
+                break;
+            }
+            content_type.push(';');
+        }
+        content_type.push_str(part)
+    }
 
-    let content_type = if mime_type.starts_with(';') {
-        ["text/plain", mime_type].concat()
-    } else if mime_type.is_empty() {
+    let content_type = if content_type.starts_with(';') {
+        ["text/plain", &content_type].concat()
+    } else if content_type.is_empty() {
         "text/plain;charset=US-ASCII".to_string()
     } else {
-        mime_type.to_string()
+        content_type
     };
 
     let body = if is_base64 {
--- a/llrt_core/src/modules/http/fetch.rs
+++ b/llrt_core/src/modules/http/fetch.rs
@@ -1,29 +1,28 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
+use std::{collections::HashSet, time::Instant};
+
 use bytes::Bytes;
 use http_body_util::Full;
 use hyper::{header::HeaderName, Method, Request, Uri};
-
-use llrt_utils::bytes::ObjectBytes;
+use llrt_utils::{bytes::ObjectBytes, encoding::bytes_from_b64};
 use rquickjs::{
     atom::PredefinedAtom,
     function::{Opt, This},
     prelude::{Async, Func},
-    Class, Coerced, Ctx, Exception, FromJs, Function, Object, Result, Value,
+    Class, Coerced, Ctx, Exception, FromJs, Function, IntoJs, Object, Result, Value,
 };
 use tokio::select;
-
-use std::{collections::HashSet, time::Instant};
 
 use crate::{
     environment,
     modules::events::abort_signal::AbortSignal,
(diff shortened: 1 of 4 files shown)


llrt_core/src/modules/http/fetch.rs
```diff
@@ -131,6 +134,50 @@ pub(crate) fn init(ctx: &Ctx<'_>, globals: &Object) -> Result<()> {
     Ok(())
 }
 
+fn parse_data_url<'js>(ctx: &Ctx<'js>, data_url: &String) -> Result<Response<'js>> {
+    let (mime_type, data) = data_url
+        .trim_start_matches("data:")
+        .split_once(',')
+        .ok_or_else(|| Exception::throw_type(ctx, "Invalid data URL format"))?;
+
+    let (is_base64, mime_type) = if mime_type.contains("base64") {
```
richarddavison, line 143: Could we simplify this? split_once by ";" and check if second part start with `base64`?

richarddavison (changes requested):
Thanks, some minor changes to the url parsing

llrt_core/src/modules/http/fetch.rs
```diff
@@ -131,6 +134,50 @@ pub(crate) fn init(ctx: &Ctx<'_>, globals: &Object) -> Result<()> {
     Ok(())
 }
 
+fn parse_data_url<'js>(ctx: &Ctx<'js>, data_url: &String) -> Result<Response<'js>> {
+    let (mime_type, data) = data_url
+        .trim_start_matches("data:")
```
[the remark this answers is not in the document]
nabetti1720: Here we just want to remove the `data:` that is not needed for processing. As far as I know, the code would be a bit verbose to remove it in any way other than `trim_start_matches`.

llrt_core/src/modules/http/fetch.rs
nabetti1720: `base64` does not always come in second. If `charset` is attached to `mime_type`, it may come third because `;` is used as a separator. If we start considering the second or third, the code is going to be redundant...

This method is basically the same as described in the [Fetch Standard](https://fetch.spec.whatwg.org/#data-urls), with a few modifications.

llrt_core/src/modules/http/fetch.rs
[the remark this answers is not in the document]
richarddavison: we should pass a string slice to the function instead of a string reference:
```rust
fn parse_data_url<'js>(ctx: &Ctx<'js>, data_url: &str){
    let (mime_type, data) = data_url["data:".len()..].split_once(',')...
}
```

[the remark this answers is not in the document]
nabetti1720: Thank you very much. I have fixed it with this idea.

llrt_core/src/modules/http/fetch.rs
richarddavison: I think a better approach here would be to do something like this:
```rust
  let mut is_base_64 = false;
  let mut content_type = String::with_capacity(10);
  for (i, part) in data.split(";").enumerate(){
      let part = part.trim();
      if  i == 1 || i == 2 {
          if part == "base64" {
              is_base_64 = true;
              break;
          }
          content_type.push(';');
      }
      content_type.push_str(part)
  }
```

richarddavison (changes requested):

llrt_core/src/modules/http/fetch.rs
nabetti1720: Thank you. It seems to be working well in several patterns that I can think of. :)

nabetti1720: Thank you. It seems to be working well in several patterns that I can think of. :)

richarddavison (approved):

PR #580, opened 2024-09-08
support require json file

ahaoboy:
### Description of changes
Support require json files

```js
console.log( require('package.json') )

```
### Checklist

### DIFF
--- a/llrt_core/src/vm.rs
+++ b/llrt_core/src/vm.rs
@@ -33,10 +33,13 @@ use zstd::{bulk::Decompressor, dict::DecoderDictionary};
 
 include!(concat!(env!("OUT_DIR"), "/bytecode_cache.rs"));
 
-use crate::modules::{
-    console,
-    crypto::SYSTEM_RANDOM,
-    path::{dirname, join_path, resolve_path},
+use crate::{
+    json_loader::JSONLoader,
+    modules::{
+        console,
+        crypto::SYSTEM_RANDOM,
+        path::{dirname, join_path, resolve_path},
+    },
 };
 
 use crate::{
@@ -391,6 +394,7 @@ impl Vm {
 
         let loader = LoaderContainer::new((
             module_loader,
+            JSONLoader,
             BinaryLoader,
             BuiltinLoader::default(),
             ScriptLoader::default()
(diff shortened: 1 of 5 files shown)


richarddavison (changes requested):
This is a great addition but should be done at import level instead of only supporting require.

richarddavison:
> > This is a great addition but should be done at import level instead of only supporting require.
> 
> Do you mean support for `import a from './a.json'`? This should implement a JSONLoader, but nodejs not support this syntax

It does not but does through import assertions. I think we can allow it in LLRT as well.
 Also left another comment in the PR regarding evaluation of JSON.

llrt_core/src/vm.rs
```diff
@@ -605,6 +605,12 @@ fn init(ctx: &Ctx<'_>, module_names: HashSet<&'static str>) -> Result<()> {
                 join_path(vec![import_directory, specifier])
             };
 
+            if import_name.ends_with(".json") {
+                let source = std::fs::read_to_string(import_name)?;
+                let v: Value = ctx.eval(format!("({source})"))?;
```
richarddavison, line 610: This will be evaluated as JavaScript. Please use `json_parse` instead.

richarddavison (changes requested):

ahaoboy:
> It does not but does through import assertions. I think we can allow it in LLRT as well.


This may be more troublesome than expected due to the confusing behavior of cjs and esm

If you use the json file as a js file that exports a js object by default
```json
{ "a": 1 }
```
```js
export default { "a": 1 } 
```
Then what we get is an object like this
```
const a = require('./a.json')

// { default: { a: 1} }
```

richarddavison:
> If you use the json file as a js file that exports a js object by default

What do you mean by this? :)

ahaoboy:
> What do you mean by this? :)

```rust
use rquickjs::{loader::Loader, Ctx, Error, Module, Result};

#[derive(Debug)]
pub struct JSONLoader;

impl Loader for JSONLoader {
    fn load<'js>(&mut self, ctx: &Ctx<'js>, path: &str) -> Result<Module<'js>> {
        if !path.ends_with(".json") {
            return Err(Error::new_loading(path));
        }
        let source = std::fs::read_to_string(path)?;
        Module::declare(ctx.clone(), path, format!("export default ({source})"))
    }
}

```

richarddavison:
Right I see what you mean. It might be better to have two separate implementations since there is no way to return an already created object from a module.

I would for the require function use `json_parse` and for the json loader use:

  `Module::declare(ctx.clone(), path, ["export default ", &source].concat())`

richarddavison (approved):

PR #583, opened 2024-09-10
Allow any value for clearTimeout

richarddavison:
### Description of changes

Accept any parameter for setTimeout. None numeric values will just be ignored.

### Checklist

### DIFF
--- a/llrt_modules/src/modules/timers.rs
+++ b/llrt_modules/src/modules/timers.rs
@@ -16,7 +16,7 @@ use once_cell::sync::Lazy;
 use rquickjs::{
     module::{Declarations, Exports, ModuleDef},
     prelude::{Func, Opt},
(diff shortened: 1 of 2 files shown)


llrt_modules/src/modules/timers.rs
```diff
@@ -130,16 +130,19 @@ fn get_timer_state<'a>(
     unsafe { rt_timers.unwrap_unchecked() }
 }
 
-fn clear_timeout_interval(ctx: Ctx<'_>, id: usize) -> Result<()> {
```
Sytten, line 133: You can use the `Coerced` wrapper for that

richarddavison: Hum, would that would return a NaN and go through the lookup unnecessary no?

Sytten: I guess it might

PR #585, opened 2024-09-11
Fix JSON.stringify of objects with function values and fix incorrect toJSON indentaion

richarddavison:
### Description of changes

keys containing functions or constructors where not ignored. Also indentation was incorrectly rendered when toJSON returned a primitive instead of an object. This PR fixes that.

### Checklist


ahaoboy:
Have llrt done any benchmarks? I'm curious about how much better simd is than the default qjs implementation.

PR #588, opened 2024-09-14
Optimize path handling

richarddavison:
### Description of changes

Optimize the path module.

### Checklist

### DIFF
--- a/llrt_modules/src/modules/child_process.rs
+++ b/llrt_modules/src/modules/child_process.rs
@@ -79,6 +79,7 @@ generate_signal_from_str_fn!(
     libc::SIGTERM
 );
 
+#[allow(unused_variables)]
 fn prepare_shell_args(
     shell: &str,
     windows_verbatim_arguments: &mut bool,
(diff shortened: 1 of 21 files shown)


llrt_modules/src/modules/child_process.rs
```diff
@@ -79,30 +79,44 @@ generate_signal_from_str_fn!(
     libc::SIGTERM
 );
 
+#[allow(unused_variables)]
 fn prepare_shell_args(
```
Sytten, line 83: The code in this function is much harder to read I feel...
And I am not sure it is more efficient with the push in the string since it might alloc twice vs a concat that will alloc once

llrt_modules/src/modules/process.rs
```diff
@@ -17,10 +17,34 @@ use rquickjs::{
 pub use crate::sysinfo::{get_arch, get_platform};
 use crate::{time, ModuleInfo, VERSION};
 
-fn cwd(ctx: Ctx<'_>) -> Result<rquickjs::String<'_>> {
-    env::current_dir()
-        .or_throw(&ctx)
-        .and_then(|path| rquickjs::String::from_str(ctx, path.to_string_lossy().as_ref()))
+#[cfg(windows)]
+fn to_slash_lossy(path_buf: PathBuf) -> String {
```
Sytten, line 21: Might want to use dunce to remove UNC paths

llrt_modules/src/modules/child_process.rs
richarddavison: It _might_ only allocate once but the second push will have capacity since cap is doubled if first needs to allocate. However we can reserve upfront since we’ll know that we’ll push the arguments plus I can push_str them rather than join + push 👌

PR #598, opened 2024-09-20
Move cargo-patch to workspace root

NoelJacob:
### Issue # (if available)

Cargo patch only works in root workspace. This is shown as warning during all builds of llrt_core.

### Description of changes

Moved cargo-patch patching from `build.rs` in `llrt_core` to patching via `cargo patch` command in Makefile in workspace root. This seems to be the only way for automated cargo-patch in this workspace.

In Makefile, I added `patch` after every `js` was called and `cargo install cargo-patch` after every cargo toolchain install.

### Checklist

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -7,6 +7,13 @@ inherits = "release"
 strip = false
 debug = true
 
+[workspace.metadata.patch.rquickjs-core]
+version = "*"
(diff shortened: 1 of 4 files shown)


richarddavison:
Hi @NoelJacob, thanks for the PR. It actually works in non-root workspaces, it just warns. We want to trigger the patch from a build script so `cargo` commands will work.

PR #601, opened 2024-09-22
feat: Support for new bundle type `hola`

### LINKED ISSUE #599: New bundle type for AWS-friendly web applications
We thought we could create a new Web Application Stack around LLRT that is compliant with standard Web APIs (in other words, not too dependent on the Node.js API), highly scalable and compatible with AWS.

We would like to call it the `hola` web application stack. Incidentally, `hola (pronounced /ˈola/)` means 'Hi' or hello' in Spanish :)

The components are as follows :
- H : Hono
- O : Drizzle ORM
- L : AWS Lambda on LLRT
- A : Amazon Aurora Serverless v2(with AWS Data API)

**1. Hono** <[Website](https://hono.dev/)> <[Github](https://github.com/honojs/hono)>
- It has a UX similar to express, but uses only the standard WebAPI. It consists of a small core and a lot of middleware, and development is very active.
- In addition, there is an adapter called `hono/aws-lambda` that provides a mechanism to wrap the Request/Response to interact with the API Gateway and ALB (https://hono.dev/docs/getting-started/aws-lambda).

**2. Drizzle ORM** <[Website](https://orm.drizzle.team/)> <[Github](https://github.com/drizzle-team/drizzle-orm)>
- Like Prisma, it supports a variety of databases, but most importantly, it works with `@aws-sdk/client-rds-data` and supports the AWS Data API (https://orm.drizzle.team/docs/get-started-postgresql#aws-data-api).

To add, it is worth noting that these products are not dependent on other packages.

I have already been able to get my lambda to work by bundling the additional modules needed by the following sources into the `full-sdk` (when actually creating the new bundle type, the number of `@aws-sdk`'s should be narrowed down to the minimum necessary).

Again, in the example below, I only need a single `index.mjs` to create a web application that is highly compatible with AWS managed services.

```javascript
// index.mjs
import { Hono } from 'hono'
import { handle } from 'hono/aws-lambda'

import { sql } from 'drizzle-orm'
import { drizzle } from 'drizzle-orm/aws-data-api/pg'
import { RDSDataClient } from '@aws-sdk/client-rds-data'

const rdsClient = new RDSDataClient({})

const db = drizzle(rdsClient, {
  database: process.env['DATABASE'],
  secretArn: process.env['SECRET_ARN'],
  resourceArn: process.env['RESOURCE_ARN'],
})

const app = new Hono()

app.post('/path/to/resource', async (c) => {
  const result = await db.execute(sql`SELECT tablename FROM pg_tables limit 10`)
  return c.json(result.rows)
})

export const handler = handle(app)
```

Execution result (sorry for the Japanese display screenshot):
image

We propose a new bundle type of `hola` to introduce a different and new use case to the LLRT world. I would be happy if you are interested. :)

nabetti1720:
### Issue # (if available)

Closed #599 

### Description of changes

We have created a new bundle type that is highly compatible with AWS Managed Services and can be easily described as one of the web application's use cases for LLRT.

### Constraint

- Some modules of `Hono` do not work, but this is because they assume the Stream API and the SubtleCrypto API in LLRT. However, we believe that they will eventually be supported by LLRT, so we have left them as they are.
- Also, the target of esbuild is downgraded to `es2020` at build time only if `hola` is selected as the bundle type. This is because part of the Hono's core uses `get` in the property name, and `quickjs` cannot interpret this correctly, so we work around it by transpiling with `es2020`. It is working fine now.

### Checklist


richarddavison:
This looks like a very cool project and is an interesting idea, but is a bit out of scope for LLRT. You could instead open up a new `hola` repository 🙂

PR #602, opened 2024-09-22
add snmalloc feature and msvc support

ahaoboy:
### Description of changes

msvc + snmalloc has a slight performance advantage over gnu, but since msvc has no equivalent implementation of gettimeofday, the lastModified attribute of the file may be slightly different from Date.now

The increase in size is due to the use of ```RUSTFLAGS: -C target-feature=+crt-static```, which reduces dll dependencies and can run on old Windows devices

| Engine | llrt-msvc | llrt-gnu |
| --- | --- | --- |
| Executable size | 12M | 10M |
| Richards | 985 | 961 |
| DeltaBlue | 1076 | 870 |
| Crypto | 604 | 1156 |
| RayTrace | 1745 | 1382 |
| EarleyBoyer | 2731 | 2145 |
| RegExp | 407 | 382 |
| Splay | 2912 | 2102 |
| NavierStokes | 1617 | 2311 |
| Score | 1247 | 1232 |

```
 ldd ./llrt-msvc.exe
        ntdll.dll => /c/WINDOWS/SYSTEM32/ntdll.dll (0x7ff9a1900000)
        KERNEL32.DLL => /c/WINDOWS/System32/KERNEL32.DLL (0x7ff99f810000)
        KERNELBASE.dll => /c/WINDOWS/System32/KERNELBASE.dll (0x7ff99ed10000)
        bcryptprimitives.dll => /c/WINDOWS/System32/bcryptprimitives.dll (0x7ff99f4c0000)
        WS2_32.dll => /c/WINDOWS/System32/WS2_32.dll (0x7ff9a05c0000)
        RPCRT4.dll => /c/WINDOWS/System32/RPCRT4.dll (0x7ff99fe60000)
        bcrypt.dll => /c/WINDOWS/SYSTEM32/bcrypt.dll (0x7ff99e590000)
        CRYPTBASE.dll => /c/WINDOWS/SYSTEM32/CRYPTBASE.dll (0x7ff99dd00000)

```

```
 ldd ./llrt-gnu.exe
        ntdll.dll => /c/WINDOWS/SYSTEM32/ntdll.dll (0x7ff9a1900000)
        KERNEL32.DLL => /c/WINDOWS/System32/KERNEL32.DLL (0x7ff99f810000)
        KERNELBASE.dll => /c/WINDOWS/System32/KERNELBASE.dll (0x7ff99ed10000)
        advapi32.dll => /c/WINDOWS/System32/advapi32.dll (0x7ff99ff80000)
        msvcrt.dll => /c/WINDOWS/System32/msvcrt.dll (0x7ff9a0050000)
        sechost.dll => /c/WINDOWS/System32/sechost.dll (0x7ff9a0480000)
        RPCRT4.dll => /c/WINDOWS/System32/RPCRT4.dll (0x7ff99fe60000)
        ws2_32.dll => /c/WINDOWS/System32/ws2_32.dll (0x7ff9a05c0000)
        bcryptprimitives.dll => /c/WINDOWS/System32/bcryptprimitives.dll (0x7ff99f4c0000)
        bcrypt.dll => /c/WINDOWS/SYSTEM32/bcrypt.dll (0x7ff99e590000)
        CRYPTBASE.DLL => /c/WINDOWS/SYSTEM32/CRYPTBASE.DLL (0x7ff99dd00000)
```

### DIFF
--- a/.github/workflows/build-artifacts.yml
+++ b/.github/workflows/build-artifacts.yml
@@ -11,8 +11,7 @@ jobs:
             platform: windows
             arch: x86_64
             release: x64
-            toolchain: nightly-x86_64-pc-windows-gnu
-            RUSTFLAGS: -C target-feature=+crt-static
+            toolchain: nightly-x86_64-pc-windows-msvc
           - os: ubuntu-latest
             platform: linux
             arch: x86_64
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -17,10 +17,11 @@ on:
       toolchain:
         required: false
         type: string
-        default: "nightly"
+        default: nightly
 
 jobs:
   build:
+    name: build-${{ inputs.platform }}-${{ inputs.arch }}
     runs-on: ${{ inputs.os }}
     steps:
       - name: Checkout
@@ -69,6 +70,9 @@ jobs:
             zstd
             patch
             unzip
+          shell: pwsh
+          run: |
+            echo "RUSTFLAGS='-C target-feature=+crt-static'" | Out-File -FilePath $env:GITHUB_ENV -Append
 
       - name: Install JavaScript dependencies
         run: |
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -118,8 +118,6 @@ jobs:
       - name: Build Windows binaries
         if: inputs.release && inputs.platform == 'windows'
         shell: msys2 {0}
-        env:
-          RUSTFLAGS: "-C target-feature=+crt-static"
         run: |
           # HACK: Add scoop/shims to PATH
           export PATH="$PATH:/c/Users/$USER/scoop/shims"
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -70,9 +70,6 @@ jobs:
             zstd
             patch
             unzip
-          shell: pwsh
-          run: |
-            echo "RUSTFLAGS='-C target-feature=+crt-static'" | Out-File -FilePath $env:GITHUB_ENV -Append
 
       - name: Install JavaScript dependencies
         run: |
@@ -121,6 +118,8 @@ jobs:
       - name: Build Windows binaries
         if: inputs.release && inputs.platform == 'windows'
         shell: msys2 {0}
+        env:
+          RUSTFLAGS: "-C target-feature=+crt-static"
         run: |
           # HACK: Add scoop/shims to PATH
           export PATH="$PATH:/c/Users/$USER/scoop/shims"
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -118,6 +118,8 @@ jobs:
       - name: Build Windows binaries
         if: inputs.release && inputs.platform == 'windows'
         shell: msys2 {0}
+        env:
+          RUSTFLAGS: "-C target-feature=+crt-static"
         run: |
           # HACK: Add scoop/shims to PATH
           export PATH="$PATH:/c/Users/$USER/scoop/shims"
--- a/Makefile
+++ b/Makefile
@@ -2,7 +2,7 @@ TARGET_linux_x86_64 = x86_64-unknown-linux-musl
 TARGET_linux_arm64 = aarch64-unknown-linux-musl
 TARGET_darwin_x86_64 = x86_64-apple-darwin
 TARGET_darwin_arm64 = aarch64-apple-darwin
-TARGET_windows_x86_64 = x86_64-pc-windows-gnu
+TARGET_windows_x86_64 = x86_64-pc-windows-msvc
 TARGET_windows_arm64 = aarch64-is-not-yet-supported
 RUST_VERSION = nightly
 TOOLCHAIN = +$(RUST_VERSION)
@@ -35,7 +35,7 @@ endif
 ZSTD_LIB_CC_ARGS = -s -O3 -flto
 ZSTD_LIB_ARGS = -j lib-nomt UNAME=Linux ZSTD_LIB_COMPRESSION=0 ZSTD_LIB_DICTBUILDER=0 AR="zig ar"
 ifeq ($(DETECTED_OS),windows)
-ZSTD_LIB_CC_x64 = CC="zig cc -target x86_64-windows-gnu $(ZSTD_LIB_CC_ARGS)"
+ZSTD_LIB_CC_x64 = CC="zig cc -target x86_64-windows-msvc $(ZSTD_LIB_CC_ARGS)"
 else
 ZSTD_LIB_CC_arm64 = CC="zig cc -target aarch64-linux-musl $(ZSTD_LIB_CC_ARGS)"
 ZSTD_LIB_CC_x64 = CC="zig cc -target x86_64-linux-musl $(ZSTD_LIB_CC_ARGS)"
--- a/tests/unit/http.test.ts
+++ b/tests/unit/http.test.ts
@@ -884,8 +884,10 @@ describe("File class", () => {
     expect(file.type).toBe("text/plain");
   });
 
-  it("has last modified date", () => {
+  it("has last modified date", async () => {
     const file = new File(["file content"], "example.txt");
+    // FIXME: msvc has no equivalent implementation of gettimeofday
+    await new Promise(r => setTimeout(r, 1000))
     const now = new Date();
     expect(file.lastModified).toBeLessThanOrEqual(now.getTime());
   });
--- a/tests/unit/http.test.ts
+++ b/tests/unit/http.test.ts
@@ -886,8 +886,10 @@ describe("File class", () => {
 
   it("has last modified date", async () => {
     const file = new File(["file content"], "example.txt");
-    // FIXME: msvc has no equivalent implementation of gettimeofday
-    await new Promise(r => setTimeout(r, 1000))
+    if (IS_WIN) {
+      // FIXME: msvc has no equivalent implementation of gettimeofday
+      await new Promise(r => setTimeout(r, 1000))
+    }
     const now = new Date();
     expect(file.lastModified).toBeLessThanOrEqual(now.getTime());
   });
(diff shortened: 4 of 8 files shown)


nabetti1720:
Minor comments.
- In LLRT, all per-target rustflags seem to be managed in .cargo/config.toml.
- Only Crypto and NavierStokes are underperforming, what could be the reason? It would be very nice if this could be improved.
- If this PR is adopted, it would be easy to support `arm64` with `aarch64-pc-windows-msvc` target.

ahaoboy:
> Minor comments.
> 
> * In LLRT, all per-target rustflags seem to be managed in .cargo/config.toml.

I moved msvc rustflags  to it too.

> * Only `Crypto` and `NavierStokes` are underperforming, what could be the reason? For example, the effect of different target-cpu. It would be very nice if this could be improved.

In fact, the difference is not big, because they all use qjs

https://github.com/ahaoboy/js-engine-benchmark


> * If this PR is adopted, it would be easy to support `arm64` with `aarch64-pc-windows-msvc` in the future.

Not familiar with this target, windows-gnu doesn't seem to support aarch64

nabetti1720:
> Not familiar with this target, windows-gnu doesn't seem to support aarch64

Sorry for the confusion. It is “msvc”, not “gnu”, that I thought could produce LLRT binaries in the future. `aarch64-pc-windows-msvc` is currently in [Tier 2 with Host Tools](https://doc.rust-lang.org/nightly/rustc/platform-support.html#tier-2-with-host-tools), the same as `aarch64-unknown-linux-musl`.

Of course, it is not too late to support this aarch64/windows artifact if someone wants to.

ahaoboy:
The flag added using the .cargo/config.toml does not work? ​​It's a bit strange
```
[target.x86_64-pc-windows-msvc]
rustflags = ["-Ctarget-feature=+crt-static"]
```

richarddavison:
@ahaoboy RUSTFLAGS will override rustflags in .cargo/config. Can you please fix?

ahaoboy:
> @ahaoboy RUSTFLAGS will override rustflags in .cargo/config. Can you please fix?

I tested four configurations:

### scope env
 ```      
env:
    RUSTFLAGS: -C target-feature=+crt-static
run: |
```

### global env
```
jobs:
  build:
    runs-on: ${{ inputs.os }}
    env:
      RUSTFLAGS: "-C target-feature=+crt-static"
```

### config
```
[target.x86_64-pc-windows-msvc]
rustflags = ["-C", "target-feature=+crt-static"]
```

this doesn't work... 
```
[target.x86_64-pc-windows-msvc]
rustflags = ["-Ctarget-feature=+crt-static"]
```

Only the global env works as expected, others will depend on vcruntime140_1.dll, which is not installed by default.

**Comment by richarddavison** [comment]
@ahaoboy you also changed from nightly to stable toolchain. That won't work as we depend on nightly compiler. I'm adding some comments in the PR

File: .github/workflows/build-artifacts.yml [hunk 1]
```diff
@@ -11,28 +11,28 @@ jobs:
             platform: windows
             arch: x86_64
             release: x64
-            toolchain: nightly-x86_64-pc-windows-gnu
-            RUSTFLAGS: -C target-feature=+crt-static
+            toolchain: stable-x86_64-pc-windows-msvc
+            RUSTFLAGS: "-C target-feature=+crt-static"
```
richarddavison, line 15: Don't use rust flags, and keep nightly toolchain

.github/workflows/build.yml
```diff
@@ -17,11 +17,18 @@ on:
       toolchain:
         required: false
         type: string
-        default: "nightly"
+        default: stable
+      RUSTFLAGS:
+        required: false
+        type: string
+        default: ""
```
richarddavison, line 24: Remove

.github/workflows/build.yml
```diff
@@ -17,11 +17,18 @@ on:
...
 
 jobs:
   build:
+    name: build-${{ inputs.platform }}-${{ inputs.arch }}
     runs-on: ${{ inputs.os }}
+    # env:
+    #   RUSTFLAGS: ${{ inputs.RUSTFLAGS }}
```
richarddavison, line 31: Remove

.github/workflows/build.yml
```diff
@@ -17,11 +17,18 @@ on:
...
-        default: "nightly"
+        default: stable
+      RUSTFLAGS:
+        required: false
+        type: string
+        default: ""
 
 jobs:
   build:
+    name: build-${{ inputs.platform }}-${{ inputs.arch }}
```
richarddavison, line 28: Nice

Makefile
```diff
@@ -2,9 +2,9 @@ TARGET_linux_x86_64 = x86_64-unknown-linux-musl
 TARGET_linux_arm64 = aarch64-unknown-linux-musl
 TARGET_darwin_x86_64 = x86_64-apple-darwin
 TARGET_darwin_arm64 = aarch64-apple-darwin
-TARGET_windows_x86_64 = x86_64-pc-windows-gnu
+TARGET_windows_x86_64 = x86_64-pc-windows-msvc
 TARGET_windows_arm64 = aarch64-is-not-yet-supported
-RUST_VERSION = nightly
+RUST_VERSION = stable
```
richarddavison, line 7: nightly

Makefile
```diff
@@ -185,7 +185,7 @@ test-e2e: js
 
 test-ci: export JS_MINIFY = 0
 test-ci: clean-js | toolchain js
-	cargo $(TOOLCHAIN) -Z panic-abort-tests test --target $(CURRENT_TARGET)
```
richarddavison, line 188: We need this back

tests/unit/http.test.ts
```diff
@@ -884,8 +884,10 @@ describe("File class", () => {
     expect(file.type).toBe("text/plain");
   });
 
-  it("has last modified date", () => {
+  it("has last modified date", async () => {
     const file = new File(["file content"], "example.txt");
+    // FIXME: msvc has no equivalent implementation of gettimeofday
+    await new Promise(r => setTimeout(r, 1000))
     const now = new Date();
     expect(file.lastModified).toBeLessThanOrEqual(now.getTime());
```
richarddavison, line 892: Do this only for windows

richarddavison (changes requested):

ahaoboy:
> @ahaoboy you also changed from nightly to stable toolchain. That won't work as we depend on nightly compiler. I'm adding some comments in the PR

Sorry for pushing the test code up before.
After testing, it is necessary to add rustflags in config.toml to ensure the local build is successful. In CI, only add environment variables for Windows build to ensure CI build is successful.

.github/workflows/build.yml
```diff
@@ -117,6 +118,8 @@ jobs:
       - name: Build Windows binaries
         if: inputs.release && inputs.platform == 'windows'
         shell: msys2 {0}
+        env:
+          RUSTFLAGS: "-C target-feature=+crt-static"
```
richarddavison, line 122: no needed since we have cargo cfg

richarddavison (changes requested):
Almost there! Just a small RUSTFLAGS that's still left

.github/workflows/build.yml
ahaoboy: https://github.com/awslabs/llrt/actions/runs/11089019083

If remove this env, it doesn't work, although the number of dlls is less, it still contains VCRUNTIME140_1.dll
I don't know how to fix this, so I just added it to ci

```
 ldd ./llrt.exe
        ntdll.dll => /c/WINDOWS/SYSTEM32/ntdll.dll (0x7ff9a1900000)
        KERNEL32.DLL => /c/WINDOWS/System32/KERNEL32.DLL (0x7ff99f810000)
        KERNELBASE.dll => /c/WINDOWS/System32/KERNELBASE.dll (0x7ff99ed10000)
        apphelp.dll => /c/WINDOWS/SYSTEM32/apphelp.dll (0x7ff99b630000)
        bcryptprimitives.dll => /c/WINDOWS/System32/bcryptprimitives.dll (0x7ff99f4c0000)
        WS2_32.dll => /c/WINDOWS/System32/WS2_32.dll (0x7ff9a05c0000)
        RPCRT4.dll => /c/WINDOWS/System32/RPCRT4.dll (0x7ff99fe60000)
        ucrtbase.dll => /c/WINDOWS/System32/ucrtbase.dll (0x7ff99f2c0000)
        bcrypt.dll => /c/WINDOWS/SYSTEM32/bcrypt.dll (0x7ff99e590000)
        VCRUNTIME140.dll => /c/WINDOWS/SYSTEM32/VCRUNTIME140.dll (0x7ff993df0000)
        CRYPTBASE.dll => /c/WINDOWS/SYSTEM32/CRYPTBASE.dll (0x7ff99dd00000)
        VCRUNTIME140_1.dll => /c/WINDOWS/SYSTEM32/VCRUNTIME140_1.dll (0x7ff97c3d0000)
        MSVCP140.dll => /c/WINDOWS/SYSTEM32/MSVCP140.dll (0x7ff97cbf0000)
```

ahaoboy: Add RUSTFLAGS
https://github.com/awslabs/llrt/actions/runs/11089325578

richarddavison: Thank you for the clarification! Taking a step back, I’m curious about the potential benefits of compiling for MSVC instead of GNU. From what I can see, the total score of the benchmark appears to be the same. One challenge with using MSVC is that it limits our ability to cross-compile for Windows from a Linux host. On the Linux side, we’re statically including the C runtime because we're targeting musl, which isn’t widely available on most distros. Additionally, musl is more lightweight and can offer better performance, especially when combined with a custom allocator like mimalloc or snmalloc. Could you test if snmalloc on GNU offers any benefits, otherwise I think we should stick to GNU :)

ahaoboy: snmalloc does not support mingw, the only reason to use msvc is to reduce the dll dependency, so that the executable program can run on some old or lite version of windows, the performance gain is very limited, tjs also has this problem when using msys environment and releases the dll and executable together. 
I also prefer gnu, maybe add msvc support when llrt needs to handle this problem

https://github.com/saghul/txiki.js/blob/9076cb366a9707990a8c615b3c8178219f29e86e/.github/workflows/release.yml#L37

richarddavison: Agree. I'll close the PR for now and we can revisit once there is a need! However I want to thank you for your contributions!

Issue #610, opened 2024-09-26
TypeError: cannot read property 'HTTP' of undefined


nabetti1720:
The error message here is not very important. As we have raised in other issues, there still seems to be a way of writing modules that require does not work on LLRT.

```javascript
const a = require('@aws-sdk/client-rds-data')
```

```
% llrt ./src/index.js
TypeError: cannot read property 'HTTP' of undefined
  at <anonymous> (/Users/shinya/Workspaces/hono-test/node_modules/@smithy/util-endpoints/dist-cjs/index.js:200:36)
      at <anonymous> (/Users/shinya/Workspaces/hono-test/node_modules/@aws-sdk/util-endpoints/dist-cjs/index.js:62:29)
      at <anonymous> (/Users/shinya/Workspaces/hono-test/node_modules/@aws-sdk/middleware-user-agent/dist-cjs/index.js:41:29)
      at <anonymous> (/Users/shinya/Workspaces/hono-test/node_modules/@aws-sdk/client-rds-data/dist-cjs/index.js:63:36)
      at <anonymous> (/Users/shinya/Workspaces/hono-test/src/index.js:6:11)
      at <anonymous> (eval_script:1:5)
```

By the way, if you load @smithy/util-endpoints, you can see the difference in behavior from Node.js as follows.

@smithy/util-endpoint (with llrt-dawrin-arm64-no-sdk):
```javascript
const a = require('@smithy/util-endpoints')
console.log(a)
```
```
% llrt ./src/reproduction.js
{}

% node ./src/reproduction.js
{
  EndpointCache: [Getter],
  EndpointError: [Getter],
  customEndpointFunctions: [Getter],
  isIpAddress: [Getter],
  isValidHostLabel: [Getter],
  resolveEndpoint: [Getter]
}
```

nabetti1720:
Even stranger behavior included the following.

1. When modules that have no dependencies on each other are loaded.
```javascript
var a = require("tinybench");
//const a = require('@smithy/types')
const b = require('@smithy/util-endpoints')
console.log(b)
```
```
% llrt ./src/reproduction.js
{}

% node ./src/reproduction.js
{
  EndpointCache: [Getter],
  EndpointError: [Getter],
  customEndpointFunctions: [Getter],
  isIpAddress: [Getter],
  isValidHostLabel: [Getter],
  resolveEndpoint: [Getter]
}
```

2. If a module with prior dependencies is loaded (but the variable to be assigned is irrelevant)

```javascript
[...]

PR #613, opened 2024-09-30
feat: Loading node_modules by import

### LINKED ISSUE #589: Loading node_modules with import/require
We know that bundling is better for performance, but for experimental or less performance-intensive use cases, it would be great if node_modules loading could be supported.

```javascript
// index4.js
import { Event } from "event-target-shim";

globalThis.Event = Event

import { Bench } from 'tinybench';

const bench = new Bench({ time: 100 });

import mod from '../package.json';

bench
  .add('json', async () => {
    JSON.stringify(mod)
  })
await bench.warmup();
await bench.run();
console.log(bench.table());
```

```shell
% llrt ./src/index4.js
ReferenceError: Error resolving module 'event-target-shim' from './src/index4.js'
```

nabetti1720:
### Issue # (if available)

Closed #589 

### Description of changes

- The import instruction now supports import of node_modules.
- For Resolver/Loader, it was divided into several parts, but they were integrated into CostomResolver/CustomLoader.　However, due to the priority given to integration, optimization has not yet been achieved.
- Currently, the module search uses a standard (i.e., mostly Node.js-friendly) search path, which results in an error if it runs into a Node.js-specific build-in module load inside the target module. Browser settings should be a priority, but it is difficult to address all cases at this time because of the various ways of defining them. We hope to gradually address this issue in the future.

### Checklist


nabetti1720:
It is still in rough shape, but I am presenting it in draft form for your input :)

llrt_core/src/vm.rs
```diff
@@ -79,11 +76,11 @@ fn print(value: String, stdout: Opt<bool>) {
 }
 
 #[derive(Debug)]
-pub struct BinaryResolver {
+pub struct CustomResolver {
```
richarddavison, line 79: Much better!

llrt_core/src/vm.rs
```diff
@@ -102,10 +99,11 @@ impl BinaryResolver {
 }
 
 #[allow(clippy::manual_strip)]
-impl Resolver for BinaryResolver {
-    fn resolve(&mut self, _ctx: &Ctx, base: &str, name: &str) -> Result<String> {
-        trace!("Try resolve \"{}\" from \"{}\"", name, base);
+impl Resolver for CustomResolver {
+    fn resolve(&mut self, ctx: &Ctx, base: &str, name: &str) -> Result<String> {
+        trace!("Try resolve '{}' from {}", name, base);
```
richarddavison, line 104: Nice

richarddavison (changes requested):
Some comments

llrt_core/src/vm.rs
```diff
@@ -748,17 +777,64 @@ fn get_module_path_in_node_modules(ctx: &Ctx<'_>, specifier: &str) -> Result<Str
...
+    } else {
+        get_module_path_via_cjs(&package_json, name)?
+    };
 
     Ok([&current_dir, "/node_modules/", scope, "/", module_path].concat())
 }
 
-fn get_module_path<'a>(package_json: &'a BorrowedValue<'a>, module_name: &str) -> Result<&'a str> {
-    if let BorrowedValue::Object(map) = package_json {
+fn get_module_path_via_esm<'a>(json: &'a BorrowedValue<'a>, module_name: &str) -> Result<&'a str> {
```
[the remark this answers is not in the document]
nabetti1720: I think we can probably do it.

llrt_core/src/vm.rs
```diff
@@ -151,29 +149,33 @@ impl Resolver for BinaryResolver {
...
+            || name.ends_with(".js")
+            || name.ends_with(".mjs")
+            || name.ends_with(".cjs")
+            || name.ends_with(".json"))
+        {
+            if let Ok(node_modules_path) = get_module_path_in_node_modules(ctx, name, true) {
+                trace!("--> node_modules path: {}", node_modules_path);
+                return Ok(node_modules_path);
+            }
+        }
```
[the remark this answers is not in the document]
nabetti1720: Sorry for the lack of clarity on Diff in GitHub. It is implemented in CustomResolver, not CustomLoader.

llrt_core/src/vm.rs
```diff
@@ -197,6 +199,20 @@ where
...
+        let name = if name.contains("node_modules") && Path::new(name).is_dir() {
+            [".js", ".mjs", ".cjs"]
+                .iter()
+                .map(|ext| [name, ext].concat())
+                .find(|path| Path::new(path).exists())
+                .unwrap_or_else(|| [name, "/index.js"].concat())
+        } else {
+            name.to_string()
+        };
+        let name = name.as_str();
```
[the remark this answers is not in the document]
nabetti1720: I also wanted to implement the workaround logic for node_modules on the Resolver side, rather than adding it to these places. However, I had no choice but to add it because I saw a behavior inside the node_modules package where only the loader is executed without going through Resolver when Javascript imports are run further from there.

I may have implemented this incorrectly, so I'll do some more testing.

llrt_core/src/vm.rs
```diff
@@ -151,29 +149,33 @@ impl Resolver for BinaryResolver {
...
-            })
-            .ok_or_else(|| Error::new_resolving(base, name))?;
+        let path = self.paths.iter().find_map(|path| {
+            let path = path.join(normalized_path);
+            let bin_path = CustomResolver::get_bin_path(&path);
+            if bin_path.exists() {
+                return Some(bin_path);
+            }
+            get_js_path(path.to_str().unwrap())
+        });
```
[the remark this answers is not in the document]
nabetti1720: This was originally thought to be the part of the BinaryResolver implementation that does not affect the resolution of mode_modules. (In other words, it cannot be resolved here, so it reaches the logic of node_modules path resolution.)

I didn't pursue this part too deeply, so I'll think about it some more by checking the contents of each variable while inserting `println!`

llrt_core/src/vm.rs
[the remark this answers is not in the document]
richarddavison: Ah right, my mistake!

llrt_core/src/vm.rs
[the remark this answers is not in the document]
nabetti1720: The reason was that the original BinaryResolver implementation returned the value as resolved when normalized_path existed as a directory.

By clearly determining when normalized_path exists as a file and when it exists as a directory, we were able to resolve the path with CostomResolver and make the Loader clean again.

llrt_core/src/vm.rs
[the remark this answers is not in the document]
nabetti1720: The integration is now complete.

llrt_core/src/vm.rs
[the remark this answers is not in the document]
nabetti1720: By “node module resolution” you are referring to the resolution of submodules imported by relative paths within a Node.js package, rather than the resolution of the Node.js package itself? Since `get_js_path()` was the same functionality that we had been forced to implement in Loader, we removed the implementation we had added to Loader and used `get_js_path()`.

We believe that the means of resolving paths from Node.js package names must still be resolved using `get_module_path_in_node_modules()`.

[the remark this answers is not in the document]
nabetti1720: I cannot decide if CustomResolver::get_bin_path() is unnecessary or not. Here, I assume that for path files that do not contain an extension, if one with the extension “lrt” exists, it is returned.

Do I need to read bytecode-compiled files on the OS?

llrt_core/src/vm.rs
```diff
@@ -67,6 +68,25 @@ pub(crate) static COMPRESSION_DICT: &[u8] =
...
+        if node_modules_path.exists() && node_modules_path.is_dir() {
+            return node_modules_path.to_str().unwrap().to_string();
+        }
+        if !base_dir.pop() {
+            break;
+        }
+    }
+    "/node_modules".to_string()
+});
+
```
richarddavison, line 89: Here, we're just checking if node_modules exists in a base dir or move up the parent directory.
This can't be static as we need each path passed to Custom resolver if any of them contains "node_modules". I think in order to get this right, we must line by line copy the implementation pseudocode mentioned here:
https://nodejs.org/api/modules.html#all-together
Note here that we should probably cache NODE_MODULES_PATHS() functions, return value based on START.
GLOBAL_FOLDERS here would be the paths added to CustomResolver

richarddavison (changes requested):

llrt_core/src/vm.rs
nabetti1720: Thank you for pointing this out. This was a major correction, but we believe it is an improvement.

nabetti1720:
Is this relevant to this correction?

image

llrt_core/src/vm.rs
richarddavison: Sorry, what I meant was replace the entire resolution implementation to pseudo algorithm I linked. The only difference from node is that we check for ".lrt" in addition to ".js" ."cjs" and ".mjs also core modules are located in bytecode_cache

richarddavison: It think they share a lot of similarities (for step 1-5) but worth refactor our current code so its a 1:1 implementation. What I don't understand is 4. When do you ever use "#"? I have never seen it.

nabetti1720: Certainly, a 1:1 implementation would be easier to compare/validate with pseudo-code, and the code would be more transparent.

Also, I have never seen a module that starts with a '#'. This is why I say that only what is possible should be implemented.

richarddavison:
> Is this relevant to this correction?
> 
> image

Yes, some string conversion that fails. The error is somewhat misleading. I'm working on a completely rewritten test runner that uses a pool of worker processes. That way if tests like these fail, it's much easier to locate.

nabetti1720:
We have begun working on a 1:1 implementation of the Node.js path resolution pseudo code.
https://nodejs.org/api/modules.html#all-together

Some of the code may be redundant, but we will first try to reproduce as much of the pseudocode as possible. I think it is currently about 50% complete, but all LLRT tests including require/import have passed on macos. And as a byproduct, the #610 issue has been resolved.

Please wait a little longer as we want to take the time to work on this...

richarddavison:
> > 1. If the "type" field is "module", load X.js as an ECMAScript module. STOP.
> > 2. If the "type" field is "commonjs", load X.js as an CommonJS module. STOP.
> 
> I have this pseudo code, but is this control possible with Resolver? It doesn't seem like it can be done since Load is the Loader's responsibility.

In our case we don't have to care about if it's a CJS or MJS module since they are interoperable and loaded the same way.

nabetti1720:
Hi @richarddavison , I have a few points to report.

- I think I've got the CJS resolver mostly implemented. I tried to get started on the ESM, but looking at the pseudo code, it looks to me like you are trying to implement something similar in a different way. LLRT works without much distinction between CJS/ESM. Is it possible to use a method where CJS is given priority when called by `require` and ESM is given priority when called by `import`?
- I could not find where the detailed specifications on how to search package.json are written. In particular, I could not find the further destination of `exports->[require | import]`. So, I have tried to make use of the previous implementation.
- Windows cargo test does not pass. I think it is probably an inconsistency regarding path in require/import as before, but I can't get to the cause because the details of the error are unknown.

There may be more things you don't like about it, but please check it out first. :)

richarddd (changes requested):

llrt_core/src/custom_resolver.rs
```diff
@@ -0,0 +1,570 @@
...
+    //    "." + X.slice("name".length), `package.json` "exports", ["node", "require"])
+    //    <a href="esm.md#resolver-algorithm-specification">defined in the ESM resolver</a>.
+    // 6. RESOLVE_ESM_MATCH(MATCH)
+    let mut package_json = fs::read(&package_json_path).unwrap_or_default();
+    let package_json = simd_json::to_borrowed_value(&mut package_json).or_throw(ctx)?;
+
+    if let Ok(path) = package_exports_resolve(&package_json, name, is_esm) {
+        trace!("|  load_package_self(2.c): {}", path);
+        return Ok(Some(path.to_string()));
+    }
```
richarddavison, line 456: Avoid parsing and loading the same json file 3 times. Parse json once as simd::borrowed and pass to get_main_field, get_exports_field etc

richarddavison (changes requested):

llrt_core/src/custom_resolver.rs
```diff
@@ -0,0 +1,570 @@
...
+fn load_as_file(ctx: &Ctx<'_>, x: &str) -> Result<Option<String>> {
+    trace!("|  load_as_file(x): {}", x);
+
+    // 1. If X is a file, load X as its file extension format. STOP
+    if Path::new(&x).is_file() {
+        trace!("|  load_as_file(1): {}", x);
+        return Ok(Some(x.to_string()));
+    }
+
+    // 2. If X.js is a file,
```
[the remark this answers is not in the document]
nabetti1720: The “.node” process was implemented in a position to be replaced by the “.node” process, but it will be added here.

llrt_core/src/custom_resolver.rs
nabetti1720: We have responded to your request.

llrt_core/src/custom_resolver.rs
```diff
@@ -0,0 +1,570 @@
...
+            // b. STOP
+            trace!("+- Resolved by `BYTECODE_CACHE`: {}\n", check_x);
+            return Ok(check_x.to_string());
+        }
+    }
+
+    // 2. If X begins with '/'
+    let y = if is_absolute(x) {
+        // a. set Y to be the file system root
+        let file_system_root = match env::consts::OS {
```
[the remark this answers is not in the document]
nabetti1720: I could not find the reference in the existing implementation. Could you please tell me where to find it?

llrt_core/src/custom_resolver.rs
```diff
@@ -0,0 +1,570 @@
...
+            _ => Path::new("/"),
+        };
+        file_system_root.to_str().unwrap()
+    } else {
+        y
+    };
+
+    // Normalize path Y to generate dirname(Y)
+    let dirname_y = if Path::new(y).is_dir() {
+        path::resolve(Rest(vec![y.to_string()]))
```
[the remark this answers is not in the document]
nabetti1720: Modify to use path::resolve_path().

llrt_core/src/custom_resolver.rs
```diff
@@ -0,0 +1,570 @@
...
+            trace!("|  load_node_modules(2.c): {}", path);
+            return Some(path);
+        }
+    }
+
+    None
+}
+
+// NODE_MODULES_PATHS(START)
+fn node_modules_paths(start: &str) -> Vec<String> {
```
richarddavison, line 313: Can we do something like this? (untested)
```rust
fn node_modules_paths(start: &str) -> Vec<PathBuf> {
    let path = Path::new(start);
    let mut dirs = Vec::new();
    let mut current = Some(path);

    // Iterate through parent directories
    while let Some(dir) = current {
        if dir.file_name().map_or(false, |name| name != "node_modules") {
            let mut node_modules = dir.to_path_buf();
            node_modules.push("node_modules");
            dirs.push(node_modules);
        }
        current = dir.parent();
    }

    // Add global folders
    if let Some(home) = home::home_dir() {
        dirs.push(home.join(".node_modules"));
        dirs.push(home.join(".node_libraries"));
    }

    dirs
}
```

nabetti1720: Does the return value have to be `Vec<PathBuf>`? If possible, I would like to return it as `Vec<String>` as before. Other than that point, this implementation is fine.

nabetti1720: Does the return value have to be Vec<PathBuf>? If possible, I would like to return it as Vec<String> as before. Other than that point, this implementation is fine.

richarddd: Oh right, that should be it. Or even better, `Vec<Rc<str>>` or `Vec<Box<str>>`. Try to avoid String if we're returning immutable strings.

richarddavison: That's right, or even return` Vec<Rc<str>>` or `Vec<Box<str>>.` Try to avoid String if we're returning immutable string values.

nabetti1720: Please wait while I work on this issue some more.

nabetti1720:
> * Windows cargo test does not pass. I think it is probably an inconsistency regarding path in require/import as before, but I can't get to the cause because the details of the error are unknown. Windows is difficult...

This issue continues to plague me. It would be nice to see a more detailed description of the error.

richarddavison:
> This issue continues to plague me. It would be nice to see a more detailed description of the error.

Consider removing -Zpanic-abort-tests in CI, that would add more helpful info!

richarddavison (changes requested):
Amazing work! Just found some small adjustments and we should be pretty ready to merge :)

nabetti1720:
> > This issue continues to plague me. It would be nice to see a more detailed description of the error.
> 
> Consider removing -Zpanic-abort-tests in CI, that would add more helpful info!

When I remove `-Z panic-abort-tests` from the cargo test options, the tests don't pass as well as before. Is it possible to turn off this option only for specific tests?
```
error[E0080]: evaluation of constant value failed
Error:   --> /home/runner/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/panic.rs:63:9
   |
63 |         $crate::panicking::panic_fmt($crate::const_format_args!($($t)+));
   |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the evaluated program panicked at 'panic_immediate_abort requires -C panic=abort', /home/runner/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/panicking.rs:35:15
   |
  ::: /home/runner/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/panicking.rs:35:15
   |
35 | const _: () = assert!(cfg!(panic = "abort"), "panic_immediate_abort requires -C panic=abort");
   |               ------------------------------------------------------------------------------- in this macro invocation
   |
   = note: this error originates in the macro `$crate::panic::panic_2021` which comes from the expansion of the macro `assert` (in Nightly builds, run with -Z macro-backtrace for more info)
```

llrt_core/src/custom_resolver.rs
```diff
@@ -0,0 +1,528 @@
...
+            Path::new(&[drive, ":\\"].concat()).to_path_buf()
+        } else {
+            Path::new("C:\\").to_path_buf()
+        }
+    } else {
+        Path::new("C:\\").to_path_buf()
+    };
+
+    root.into_os_string().into_string().unwrap()
+}
```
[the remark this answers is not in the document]
nabetti1720: The proposed implementation caused an error, so we have made some improvements.

richarddavison:
> > > This issue continues to plague me. It would be nice to see a more detailed description of the error.
> > 
> > 
> > Consider removing -Zpanic-abort-tests in CI, that would add more helpful info!
> 
> When I remove `-Z panic-abort-tests` from the cargo test options, the tests don't pass as well as before. Is it possible to turn off this option only for specific tests?
> 
> ```
> error[E0080]: evaluation of constant value failed
> Error:   --> /home/runner/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/panic.rs:63:9
>    |
> 63 |         $crate::panicking::panic_fmt($crate::const_format_args!($($t)+));
>    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the evaluated program panicked at 'panic_immediate_abort requires -C panic=abort', /home/runner/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/panicking.rs:35:15
>    |
>   ::: /home/runner/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/panicking.rs:35:15
>    |
> 35 | const _: () = assert!(cfg!(panic = "abort"), "panic_immediate_abort requires -C panic=abort");
>    |               ------------------------------------------------------------------------------- in this macro invocation
>    |
>    = note: this error originates in the macro `$crate::panic::panic_2021` which comes from the expansion of the macro `assert` (in Nightly builds, run with -Z macro-backtrace for more info)
> ```

No, you're right, we need panic abort since we have it enabled as a feature in cargo.toml.
I think we can change test-ci target to this:

```makefile
test-ci: export JS_MINIFY = 0
test-ci: export RUST_BACKTRACE = 1
test-ci: clean-js | toolchain js 
	cargo $(TOOLCHAIN) -Z panic-abort-tests test --target $(CURRENT_TARGET)  -- --nocapture --show-output 
	cargo $(TOOLCHAIN) run -r --target $(CURRENT_TARGET) -- test -d bundle/js/__tests__/unit
```

nabetti1720:
> No, you're right, we need panic abort since we have it enabled as a feature in cargo.toml. I think we can change test-c target to this:
> 
> ```makefile
> test-ci: export JS_MINIFY = 0
> test-ci: export RUST_BACKTRACE = 1
> test-ci: clean-js | toolchain js 
> 	cargo $(TOOLCHAIN) -Z panic-abort-tests test --target $(CURRENT_TARGET)  -- --nocapture --show-output 
> 	cargo $(TOOLCHAIN) run -r --target $(CURRENT_TARGET) -- test -d bundle/js/__tests__/unit
> ```

Unfortunately, no detailed information was output.

```
test runtime_client::tests::runtime ... FAILED
test number::test::test_base_conversions ... ok
failures:
---- runtime_client::tests::runtime stdout ----
---- runtime_client::tests::runtime stderr ----
note: got unexpected return code -1073741795
failures:
    runtime_client::tests::runtime
test result: FAILED. 26 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 1.14s
error: test failed, to rerun pass `-p llrt_core --lib`
make: *** [Makefile:189: test-ci] Error 101
Error: Process completed with exit code 2.
```

richarddavison:
> > No, you're right, we need panic abort since we have it enabled as a feature in cargo.toml. I think we can change test-c target to this:
> > ```makefile
> > test-ci: export JS_MINIFY = 0
> > test-ci: export RUST_BACKTRACE = 1
> > test-ci: clean-js | toolchain js 
> > 	cargo $(TOOLCHAIN) -Z panic-abort-tests test --target $(CURRENT_TARGET)  -- --nocapture --show-output 
> > 	cargo $(TOOLCHAIN) run -r --target $(CURRENT_TARGET) -- test -d bundle/js/__tests__/unit
> > ```
> 
> Unfortunately, no detailed information was output.
> 
> ```
> test runtime_client::tests::runtime ... FAILED
> test number::test::test_base_conversions ... ok
> failures:
> ---- runtime_client::tests::runtime stdout ----
> ---- runtime_client::tests::runtime stderr ----
> note: got unexpected return code -1073741795
> failures:
>     runtime_client::tests::runtime
> test result: FAILED. 26 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 1.14s
> error: test failed, to rerun pass `-p llrt_core --lib`
> make: *** [Makefile:189: test-ci] Error 101
> Error: Process completed with exit code 2.
> ```

I recon its panic abort that disables the backtraces. We can enable log tracing and see what's going on:
Add this to the makefile target
`test-ci: export LLRT_LOG = trace`

nabetti1720:
> I recon its panic abort that disables the backtraces. We can enable log tracing and see what's going on: Add this to the makefile target `test-ci: export LLRT_LOG = trace`

I tried the following two options in `test-ci`, but it still did not output detailed information.

- export RUST_LOG = trace
- export RUST_LOG = llrt=trace

nabetti1720:
> EDIT: I skipped `runtime_client::tests::runtime` and got some resolve errors in my Javascript tests. This may be a clue, so I will look into it more.

Hi @richarddavison , In the new Resolver, it was not loading correctly when require/import was specified by windows absolute path. Please confirm that we have fixed it and both cargo test/js test pass. Other errors are still occurring in the latest ci, but we are aware that they are not related to this fix.

PR #623, opened 2024-10-11
fix: Load error when no path is specified

### LINKED ISSUE #622: `llrt compile` output binary runtime error
The binary that results from `llrt compile <file.js>` no long runs and returns the error on the last line below.

```
llrt compile hello_console.js
llrt hello_console.lrt
ReferenceError: Error resolving module 'hello_console.lrt'
```

nabetti1720:
### Issue # (if available)

Fixed #622

### Description of changes

- Fixed a loading error when no path is specified.
- Test cases were also added to check for absolute path/relative path/no path specified cases.

### Checklist

### DIFF
--- a/hello.js
+++ b/hello.js
@@ -1,0 +1,2 @@
+export const hello = "hello world!";
+console.log(hello);
--- a/hello.js
+++ b/hello.js
@@ -1,2 +1,0 @@
-export const hello = "hello world!";
-console.log(hello);
--- a/index.mjs
+++ b/index.mjs
@@ -1,1 +1,2 @@
+export const hello = "hello world!";
 console.log("\x1B[32m✔\x1B[0m", "a");
--- a/index.mjs
+++ b/index.mjs
@@ -1,2 +1,1 @@
-export const hello = "hello world!";
 console.log("\x1B[32m✔\x1B[0m", "a");
--- a/fixtures/package.json
+++ b/fixtures/package.json
@@ -1,0 +1,3 @@
+{
+  "private": true
+}
--- a/llrt_core/src/custom_resolver.rs
+++ b/llrt_core/src/custom_resolver.rs
@@ -131,6 +131,12 @@ pub fn require_resolve(ctx: &Ctx<'_>, x: &str, y: &str, is_esm: bool) -> Result<
         return Ok(path.to_string());
     }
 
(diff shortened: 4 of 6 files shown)


hello.js
richarddavison, line 1: We already have this file in `./fixtures` directory

richarddavison (changes requested):

hello.js
nabetti1720: Yes, I did. The test case for relative paths is . /fixtures, I believe you can refer to them in /fixtures. But when testing a file load without a path specification, I really needed to put the file in the current directory.

nabetti1720: Modify what already exists to be used effectively in the test case.

nabetti1720: The following error occurred in the latest ci. Does this have anything to do with this correction?

```
TypeError: Conversion from string failed: invalid utf-8 sequence of 1 bytes from index 0
  
thread 'modules::http::fetch::tests::test_fetch_function' panicked at llrt_core\src\vm.rs:359:17:
Exception(
Exception { message: Some("Conversion from string failed: invalid utf-8 sequence of 1 bytes from index 0"), file: None, line: Some(-1), column: Some(-1), stack: Some("") })

Testing ../fixtures/handler.handler
test modules::http::fetch::tests::test_fetch_function ... FAILED
```

richarddavison: Hi, no it does not. There is some issue in windows and back traces aren't properly catching this. I'm trying to find a better solution.

richarddavison: It's a little tricky to reproduce since it only occurs "sometimes" on windows builds. I suspect this has to do with encoding raw data to json in here https://github.com/awslabs/llrt/blob/45af2207f021365a12b922fae3c7e251d32fbdfd/llrt_utils/src/encoding.rs#L138

richarddavison (changes requested):
Almost there!

hello.js
nabetti1720: I see. In any case, it is too esoteric for me to handle, so I will leave it alone. :)

index.mjs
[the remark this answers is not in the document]
nabetti1720: As for how the paths of the js and json files are determined in load_as_file(), the content is equivalent, just the order of confirmation is different. So we have simplified the absolute path loading test a bit.

PR #625, opened 2024-10-14
docs(README.md): Improve readability

dzvon:
The footnotes are displayed in a single line, which is hard to
read. This commit adds a line break to improve readability.


richarddavison (approved):
Thanks, this is better

PR #626, opened 2024-10-14
fix: Prevent unnecessary merging of absolute paths

nabetti1720:
### Description of changes

- Suppress unnecessary merging of absolute paths (works fine as is, but the path representation is weird when outputting stack traces).
- We also noticed regression in makefile, so we fixed it as well.

### Checklist


richarddavison (approved):
Fantastic, thank you! I also found the issue causing "random" build errors on Windows. It's a very tricky issue but it comes from inaccurate mapping of JS_ATOM_NULL from generated Rust bindings that differs from QJS definitions. This means that JS_ATOM_NULL in Rust is NOT equal to JS_ATOM_NULL in QuickJS and we return a random string pointer when trying to resolve the base name. Since we ignore the base the error only happens when the dangling pointer returns invalid utf8 chars which is random.

PR #634, opened 2024-10-18
feat: Browser priority for loading Node.js packages

nabetti1720:
### Description of changes

- Node.js packages usually use the Node.js API, but some have a module for the Browser.
- In the case of LLRT, as can be seen from the fact that the bundle target of `@aws-sdk` is Browser, if there is a module for Browser, it is more likely to work if you give priority to that one. At least modules like `axios` will now work.
- There may be other ways to specify a different Browser, but there doesn't seem to be a clear rule, so we will address them as soon as we find them.

node_modules/axios/package.json:
```json
  "exports": {
    ".": {
      "browser": {
        "require": "./dist/browser/axios.cjs",
        "default": "./index.js"
      },
      "default": {
        "require": "./dist/node/axios.cjs",
        "default": "./index.js"
      }
    },
  },
```

### Checklist


richarddavison (approved):
Nice!!!

PR #637, opened 2024-10-20
feat: Initial implementation of Event

nabetti1720:
### Description of changes

- Initial implementation of `Event` to advance WinterCG compliance.
https://dom.spec.whatwg.org/#event
- Since we do not know how much is really needed on the server side, we will start with the minimum necessary implementation. However, packages such as `tinybench` work well with this implementation.

```javascript
// bench.js

// import { Event } from "event-target-shim";
// globalThis.Event = Event

import { Bench } from 'tinybench';
import mod from './package.json';

const bench = new Bench({ time: 100 });
bench
  .add('json', async () => {
    JSON.stringify(mod)
  })
await bench.warmup();
await bench.run();
console.log(bench.table());
```

```
$ llrt bench.js 
[ {
    Task Name: 'json',
    ops/sec: '205713',
    Average Time (ns): 4861.136447598518,
    Margin: '±0.41%',
    Samples: 20572
  } ]
  ```

### Checklist

### DIFF
--- a/llrt_modules/src/modules/events/event.rs
+++ b/llrt_modules/src/modules/events/event.rs
@@ -1,26 +1,22 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
-use std::marker::PhantomData;
-
 use rquickjs::{prelude::Opt, Result, Value};
 
 use llrt_utils::object::ObjectExt;
 
 #[rquickjs::class]
 #[derive(rquickjs::class::Trace)]
-pub struct Event<'js> {
+pub struct Event {
     event_type: String,
     bubbles: bool,
     cancelable: bool,
     composed: bool,
-    #[qjs(skip_trace)]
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: llrt_modules/src/modules/events/event.rs [hunk 1]
```diff
@@ -0,0 +1,66 @@
...
+
+#[rquickjs::class]
+#[derive(rquickjs::class::Trace)]
+pub struct Event<'js> {
+    event_type: String,
+    bubbles: bool,
+    cancelable: bool,
+    composed: bool,
+    #[qjs(skip_trace)]
+    marker: PhantomData<&'js ()>,
```
richarddavison, line 17: Why do we need the lifetime here?

richarddavison (changes requested):

llrt_modules/src/modules/events/event.rs
nabetti1720: It was completely unnecessary at this time and has been removed.

richarddavison (approved):

PR #640, opened 2024-10-22
Refactor crates

Sytten:
### Description of changes

- Split `llrt_modules` into multiple crates, this is more maintainable and easier to track dependencies
- Move `http` module to it's own crate (that also implied creating `llrt_abort` and `llrt_json`
- Added a clonable `BodyVariant` to avoid freezing the current thread (and also it no works with single threaded runtimes

### Checklist


llrt_utils/build.rs
```diff
@@ -0,0 +1,12 @@
+// WARN: Do not modify this function without changing the other build.rs files in the other crates.
+fn set_nightly_cfg() {
```
richarddavison, line 2: We can move this to a shared file or create and reuse.

modules/llrt_http/src/lib.rs
```diff
@@ -0,0 +1,160 @@
...
+mod file;
+mod headers;
+mod incoming;
+mod request;
+mod response;
+mod security;
+
+const DEFAULT_CONNECTION_POOL_IDLE_TIMEOUT: Duration = Duration::from_secs(15);
+
+static CONNECTION_POOL_IDLE_TIMEOUT: OnceLock<Duration> = OnceLock::new();
```
richarddavison, line 35: We can avoid locking by using an AtomicUsize and then build Duration from ms

modules/llrt_http/src/response.rs
```diff
@@ -141,34 +143,48 @@ impl<'js> Response<'js> {
...
+                "gzip" => GzDecoder::new(&bytes[..]).read_to_end(&mut data)?,
+                "deflate" => ZlibDecoder::new(&bytes[..]).read_to_end(&mut data)?,
+                _ => return Err(Exception::throw_message(ctx, "Unsupported encoding")),
+            };
+            Ok(data)
+        } else {
+            Ok(bytes.to_vec())
+        }
+    }
+
```
richarddavison, line 174: Out of scope for this PR but we should be able to stream to the decompression decoders rather than from memory

richarddavison (changes requested):
Some minor comments :)

modules/llrt_http/src/response.rs
Sytten: Agreed, also I think my cloned body might be short lived if we can land the streams.

llrt_utils/build.rs
Sytten: Sure

.github/workflows/build-modules.yml
```diff
@@ -33,24 +33,10 @@ jobs:
...
-                    "path"
-                    "process"
-                    "perf-hooks"
-                    "timers"
-                    "url"
-                    "zlib")
-          for feature in "${features[@]}"; do
+          readarray crates < <(cargo metadata --no-deps --format-version 1 --quiet | jq -r '.packages[] | select(.manifest_path | contains("modules/")) | .name')
+          for crate in "${crates[@]}"; do
             echo "Compiling feature: $feature"
```
nabetti1720, line 38: Isn't `$crate` correct, not `$feature`?

Sytten: Good catch

.github/workflows/build-modules.yml
```diff
@@ -33,24 +33,10 @@ jobs:
...
-                    "timers"
-                    "url"
-                    "zlib")
-          for feature in "${features[@]}"; do
-            echo "Compiling feature: $feature"
-            cargo build -p llrt_modules --no-default-features --features $feature
+          readarray crates < <(cargo metadata --no-deps --format-version 1 --quiet | jq -r '.packages[] | select(.manifest_path | contains("modules/")) | .name')
+          for crate in "${crates[@]}"; do
+            echo "Compiling feature: $crate"
+            cargo build -p $crate
```
richarddavison, line 39: Isn't this used in `compile.sh`script?

modules/llrt_http/src/incoming.rs
```diff
@@ -0,0 +1,354 @@
...
+                    let mut response = response.borrow_mut();
+                    let mut response2 = response.clone(ctx.clone()).unwrap();
+
+                    let response_text = response.text(ctx.clone()).await.unwrap();
+                    assert_eq!(response.status(), 200);
+                    assert_eq!(response_text.as_bytes(), welcome_message);
+
+                    let response2_err = response2.text(ctx.clone()).await.catch(&ctx).unwrap_err();
+                    assert_eq!(response2.status(), 200);
+                    assert!(matches!(response2_err, CaughtError::Exception(e) if e.message().unwrap() == ERROR_MSG_BROADCAST_LAGGED));
```
richarddavison, line 345: Why do we lagg behind here? This would be possible no? 
```javascript
const a = await fetch("https://www.amazon.com")
const b = a.clone();
console.log(await a.text(), await b.text())
```

modules/llrt_http/src/lib.rs
richarddavison: That's true!

richarddavison (changes requested):

.github/workflows/build-modules.yml
Sytten: I will check

modules/llrt_http/src/incoming.rs
Sytten: So lag is due to our implementation of broadcast. Broadcast uses a bounded channel, I set it to 16. Your example should work, but it is not always guaranteed to, it depends on how many frames are in the fetch.

The official implementation mentions that it would buffer indefinitely but also says that it is not ideal for readers to lag behind.

This implementation is a middle ground, I expect we will rip it off once we have proper streams.

.github/workflows/build-modules.yml
Sytten: Thats a file I created for local testing, dont need it

richarddavison (approved):
LGTM! Nice work 👍

PR #647, opened 2024-10-25
feat: Implementation of buffer.constants

nabetti1720:
### Description of changes

This is an implementation of `buffer.constants` used by `@opensearch-project/opensearch` and others.

- `constants.MAX_LENGTH` :  Conforms to the Node.js specification.
https://nodejs.org/api/buffer.html#bufferconstantsmax_length
- `constants.MAX_STRING_LENGTH` :  Since we did not know what to define, we chose a minimum value for the major runtimes. If you have a more suitable value, I would appreciate your advice.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/length#description

### Checklist

### DIFF
--- a/modules/llrt_buffer/src/buffer.rs
+++ b/modules/llrt_buffer/src/buffer.rs
@@ -334,10 +334,7 @@ impl ModuleDef for BufferModule {
         let buf: Constructor = globals.get(stringify!(Buffer))?;
 
         let constants = Object::new(ctx.clone())?;
-        #[cfg(target_pointer_width = "32")]
-        constants.set("MAX_LENGTH", (1u32 << 30) - 1);
-        #[cfg(target_pointer_width = "64")]
-        constants.set("MAX_LENGTH", (1u64 << 53) - 1)?;
+        constants.set("MAX_LENGTH", usize::MAX)?;
         constants.set("MAX_STRING_LENGTH", (1u32 << 28) - 16)?;
 
         export_default(ctx, exports, |default| {
--- a/modules/llrt_buffer/src/buffer.rs
+++ b/modules/llrt_buffer/src/buffer.rs
@@ -334,8 +334,8 @@ impl ModuleDef for BufferModule {
         let buf: Constructor = globals.get(stringify!(Buffer))?;
 
         let constants = Object::new(ctx.clone())?;
-        constants.set("MAX_LENGTH", usize::MAX)?;
-        constants.set("MAX_STRING_LENGTH", (1u32 << 28) - 16)?;
+        constants.set("MAX_LENGTH", u32::MAX)?; // For QuickJS
+        constants.set("MAX_STRING_LENGTH", (1 << 30) - 1)?; // For QuickJS
 
         export_default(ctx, exports, |default| {
             default.set(stringify!(Buffer), buf)?;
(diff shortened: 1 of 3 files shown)


modules/llrt_buffer/src/buffer.rs
```diff
@@ -332,10 +333,18 @@ impl ModuleDef for BufferModule {
         let globals = ctx.globals();
         let buf: Constructor = globals.get(stringify!(Buffer))?;
 
+        let constants = Object::new(ctx.clone())?;
+        #[cfg(target_pointer_width = "32")]
+        constants.set("MAX_LENGTH", (1u32 << 30) - 1);
+        #[cfg(target_pointer_width = "64")]
+        constants.set("MAX_LENGTH", (1u64 << 53) - 1)?;
```
Sytten, line 340: I think you can do usize::MAX and get rid of that?

nabetti1720: Thanks for pointing that out. And I will comment here that I like very much the modularization you have implemented. :)

modules/llrt_buffer/src/buffer.rs
```diff
@@ -332,10 +333,15 @@ impl ModuleDef for BufferModule {
         let globals = ctx.globals();
         let buf: Constructor = globals.get(stringify!(Buffer))?;
 
+        let constants = Object::new(ctx.clone())?;
+        constants.set("MAX_LENGTH", usize::MAX)?;
```
richarddavison, line 337: Max length is not `usize:MAX`. I've looked this up and it is very confusing.
On my system using Node.js 22 I have 8TB as `MAX_LENGTH` = 9007199254740991 =  0x1FFFFFFFFFFFFF which seems to be derived from here from v8
https://github.com/v8/v8/blob/4c13cae8b15a7ae40007a38f8f2af7c2fdb1aa60/src/common/globals.h#L1859

However I can't allocate a Buffer (nor a typed array) that big:
```javascript
Buffer.alloc(require("buffer").constants.MAX_LENGTH)
```
```shell
Uncaught RangeError: Array buffer allocation failed
    at new ArrayBuffer (<anonymous>)
```

modules/llrt_buffer/src/buffer.rs
```diff
@@ -332,10 +333,15 @@ impl ModuleDef for BufferModule {
...
+        constants.set("MAX_STRING_LENGTH", (1u32 << 28) - 16)?;
```
richarddavison, line 338: In my system its: `(2<<28)-24`

richarddavison (changes requested):

modules/llrt_buffer/src/buffer.rs
nabetti1720: I understood these constants to be the maximum possible values in the specification when there is no upper limit on computing resources.

If I want these to return the maximum possible values that can be taken in the execution environment, is there a better way to do this?

richarddavison: QJS supports online u32::MAX so we can stick with this
```c
 if (len > INT32_MAX) {
        JS_ThrowRangeError(ctx, "invalid array buffer length");
        goto fail;
    }
```

modules/llrt_buffer/src/buffer.rs
richarddavison: For QJS it's so we should stick with this
```c
#define JS_STRING_LEN_MAX ((1 << 30) - 1)
```

nabetti1720:
Thanks @richarddavison and @Sytten. I believe this has been corrected.

richarddavison (approved):

PR #653, opened 2024-10-26
feat: OS Module Improvements

nabetti1720:
### Description of changes

Added most of the methods of the OS module.

- os.cpus()
- os.devNull
- os.endianness()
- os.freemem()
- os.getPriority()
- os.homedir()
- os.hostname()
- os.loadavg()
- os.machine()
- os.networkInterfaces()
- os.setPriority()
- os.totalmem()
- os.uptime()
- os.userInfo()

### Checklist

### DIFF
--- a/modules/llrt_os/src/lib.rs
+++ b/modules/llrt_os/src/lib.rs
@@ -18,7 +18,7 @@ use rquickjs::{
     prelude::Func,
     Ctx, Exception, Object, Result,
 };
-use sysinfo::{MemoryRefreshKind, Networks, System};
+use sysinfo::{CpuRefreshKind, MemoryRefreshKind, Networks, RefreshKind, System};
 
 #[cfg(unix)]
 use self::unix::{get_type, get_version, DEV_NULL, EOL};
@@ -30,7 +30,14 @@ mod unix;
 #[cfg(windows)]
 mod windows;
 
-static SYSTEM: Lazy<Arc<Mutex<System>>> = Lazy::new(|| Arc::new(Mutex::new(System::new_all())));
+static SYSTEM: Lazy<Arc<Mutex<System>>> = Lazy::new(|| {
+    Arc::new(Mutex::new(System::new_with_specifics(
+        RefreshKind::new()
+            .with_cpu(CpuRefreshKind::new().with_cpu_usage().with_frequency())
+            .with_memory(MemoryRefreshKind::new().with_ram()),
+    )))
+});
+
 static NETWORKS: Lazy<Arc<Mutex<Networks>>> =
     Lazy::new(|| Arc::new(Mutex::new(Networks::new_with_refreshed_list())));
 
--- a/modules/llrt_os/src/lib.rs
+++ b/modules/llrt_os/src/lib.rs
@@ -5,6 +5,7 @@ use std::{
     env,
     net::{Ipv4Addr, Ipv6Addr},
     sync::{Arc, Mutex},
+    thread,
 };
 
 use llrt_utils::{
@@ -47,8 +48,10 @@ fn get_available_parallelism() -> usize {
 
 fn get_cpus(ctx: Ctx<'_>) -> Result<Vec<Object>> {
     let mut vec: Vec<Object> = Vec::new();
+    let mut system = SYSTEM.lock().unwrap();
 
-    let system = SYSTEM.lock().unwrap();
+    thread::sleep(sysinfo::MINIMUM_CPU_UPDATE_INTERVAL);
+    system.refresh_cpu_all();
 
     for cpu in system.cpus() {
         let obj = Object::new(ctx.clone())?;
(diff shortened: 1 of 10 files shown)


richarddavison (changes requested):
Thanks for the PR 🙌I think some changes are required for performance reasons: Read recommendations here:
https://docs.rs/sysinfo/latest/sysinfo/#good-practice--performance-tips
Also disable threading as it will introduce contagion with tokio

nabetti1720:
> Thanks for the PR 🙌I think some changes are required for performance reasons: Read recommendations here: https://docs.rs/sysinfo/latest/sysinfo/#good-practice--performance-tips Also disable threading as it will introduce contagion with tokio

- I think I've narrowed down the sysinfo function flags so I can avoid multithreading problems.
- To improve performance, we would like to “define it once and use it again” as described on the web site, but this is not working. In this use case, I believe that Systems and Networks are the target, is the approach of using lazy_static correct?

richarddavison (changes requested):
Thanks! Still some questions :)

modules/llrt_os/src/unix.rs
```diff
@@ -1,49 +1,113 @@
...
+        CStr::from_ptr(info.sysname.as_ptr())
+            .to_string_lossy()
+            .into_owned()
+    };
+    // SAFETY: `info.release` is a valid NUL-terminated pointer.
+    let _ = unsafe {
+        CStr::from_ptr(info.release.as_ptr())
+            .to_string_lossy()
+            .into_owned()
+    };
```
Sytten, line 105: We dont seem to be using this

modules/llrt_os/src/lib.rs
```diff
@@ -1,47 +1,291 @@
...
+
+    Ok((is_internal, scope_id))
+}
+
+fn get_release(ctx: Ctx<'_>) -> Result<String> {
+    match System::kernel_version() {
+        Some(val) => Ok(val),
+        None => Err(Exception::throw_reference(&ctx, "System::kernel_version")),
+    }
+}
```
Sytten, line 244: I get this is a safe interface but since already call uname I am not sure what we gain by using this

Sytten (changes requested):
Small changes.
I think I will fork this module since I don't want to expose those new APIs to my users and not a fan of the new dependencies, but that's a me problem.

modules/llrt_os/src/lib.rs
```diff
@@ -1,47 +1,291 @@
...
+
+fn get_available_parallelism() -> usize {
+    num_cpus::get()
+}
+
+fn get_cpus(ctx: Ctx<'_>) -> Result<Vec<Object>> {
+    let mut vec: Vec<Object> = Vec::new();
+    let mut system = SYSTEM.lock().unwrap();
+
+    thread::sleep(sysinfo::MINIMUM_CPU_UPDATE_INTERVAL);
```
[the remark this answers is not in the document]
Sytten: Might be a good idea to create a `Cache` structure to store the last time the cache was refreshed and put a config on that refresh period (similar to what we do with http version).

```rust
static REFRESH_PERIOD: AtomicU64 = AtomicU64::new();

struct Cached<T> {
  value: T,
  updated_at: u64
}

impl<T> Cached<T> {
   fn should_refresh(&self) -> bool {
       self.updated >= now + REFRESH_PERIOD
   }
}
```

modules/llrt_os/src/lib.rs
```diff
@@ -1,47 +1,291 @@
...
+static SYSTEM: Lazy<Arc<Mutex<System>>> = Lazy::new(|| {
+    Arc::new(Mutex::new(System::new_with_specifics(
+        RefreshKind::new()
+            .with_cpu(CpuRefreshKind::new().with_cpu_usage().with_frequency())
+            .with_memory(MemoryRefreshKind::new().with_ram()),
+    )))
+});
+
+static NETWORKS: Lazy<Arc<Mutex<Networks>>> =
+    Lazy::new(|| Arc::new(Mutex::new(Networks::new_with_refreshed_list())));
```
Sytten, line 45: There is medium to high a chance that at some point the code that update those mutexes will panic because we call a lot of C code in it. If that happens those mutexes will be poisoned for all future executions even if that particular quickjs is killed. That is why I usually don't use the std mutexes and I prefer the parking_lot mutexes.
@richarddavison What are your thoughts on that

Sytten:
I would have a small ask/idea, could we create feature flags for the various categories of APIs.
Something like `statistics`, `network`, etc. All enabled by default obviously.
The goal is that in restricted use-cases you might want to limit the APIs offered to the client code.
I get that the user of the module could code it's own module if they want full control, but that would be a middle ground.

gc-victor:
> I would have a small ask/idea, could we create feature flags for the various categories of APIs. Something like `statistics`, `network`, etc. All enabled by default obviously. The goal is that in restricted use-cases you might want to limit the APIs offered to the client code. I get that the user of the module could code it's own module if they want full control, but that would be a middle ground. Thoughts @richarddavison?

This could be a good reference: https://github.com/rscarson/rustyscript?tab=readme-ov-file#crate-features

Sytten:
Thats even more precise then that for the os module. It is more akin to http1, http2, compression-rust, compression-c I added in the http module

nabetti1720:
Thank you all for your ideas.

https://github.com/awslabs/llrt/blob/1b967873ad244c827716f2a98f98b868a0a72959/modules/llrt_http/Cargo.toml#L14-L21

I was rather thinking that the following flags would be preferable, with reference to the above.

- Allow feature flags in sysinfo to be transparently set.
- However, the default is no flag specification, and it should be an opt-in method of addition.
- In LLRT, all feature flags should be specified.

modules/llrt_os/src/lib.rs
nabetti1720: Indeed, parking_lot::Mutex looks good, so I installed it.  Please check it out. :)

nabetti1720:
We have implemented feature flags for statistics and network. I think I can further optimize by creating flags such as infomation and collecting all the remaining sysinfo calls. Please let me know what you all think.

modules/llrt_os/src/lib.rs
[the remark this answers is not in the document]
nabetti1720: Sorry, it seemed that if we only wanted to get brand and frequency, there was no need to store differential information. Therefore, I decided to remove sleep and refresh.

https://github.com/GuillaumeGomez/sysinfo/blob/e9ac63b59096f7cee8230043036ffb1dbbe6af09/src/common/system.rs#L2406-L2437

modules/llrt_os/src/lib.rs
richarddavison: @Sytten I don't see a lot of c code in the sysinfo create. It basically reads unix files that contains system info mostly in pure rust. Nevertheless I see some nice benefits of using parking lot mutex instead of std one:

> - No poisoning, the lock is released normally on panic.
> - Only requires 1 byte of space, whereas the standard library boxes the Mutex due to platform limitations.
> - Can be statically constructed.
> - Does not require any drop glue when dropped.
> - Inline fast path for the uncontended case.
> - Efficient handling of micro-contention using adaptive spinning.
> - Allows raw locking & unlocking without a guard.
> - Supports eventual fairness so that the mutex is fair on average.
> - Optionally allows making the mutex fair by calling MutexGuard::unlock_fair.

Similarly we have this for RwLock from parking_lot. If we add this we should use this crate exclusively for all RwLocks and Mutexes. But this is out of scope for this PR. So @nabetti1720 please use std mutex and (if you want) create a new PR replacing all Mutexes and RwLocks :)

nabetti1720:
> We have implemented feature flags for `statistics` and `network`. I think I can further optimize by creating flags such as `information` and collecting all the remaining `sysinfo` calls. Furthermore, for `[get|set]Priority`, it may be better to have a separate flag such as `priority`.
> 
> Or should feature flags be created on a more granular basis? Please let me know what you all think.

Since there seems to be no particular request, I would like to conclude the implementation of feature flags for now. And I think we have done what we should have done with this PR.

modules/llrt_os/src/lib.rs
nabetti1720: Sorry, I was a bit premature. I have cut back to `std::sync::Mutex` and would appreciate confirmation. :)

richarddavison (approved):

PR #663, opened 2024-11-01
Update deps

richarddavison:
### Description of changes

Update and remove deprecated deps.

### Checklist


Sytten:
General comment but I would really like us to be more relaxed in the dependencies of the modules.
The general rule for libraries are:
1. If it is not v1, then update the minor when it comes out. If possible keep a range of version (`>=0.12, <=0.13`)
2. If v1, take the oldest minor version you want to support (usually the one when you started the lib) and don't update it unless you need a new feature from the lib
3. The patch version is only specified if you really need a specific patch

modules/llrt_os/Cargo.toml
```diff
@@ -31,7 +31,7 @@ users = { version = "0.11", features = [
 
 [target.'cfg(windows)'.dependencies]
 whoami = { version = "1", default-features = false }
-windows-registry = "0.2"
+windows-registry = "0.3"
```
Sytten, line 34: This is a good update

modules/llrt_http/Cargo.toml
```diff
@@ -44,17 +44,17 @@ once_cell = "1"
...
-rustls = { version = "0.23.11", default-features = false, features = [
+rustls = { version = "0.23.16", default-features = false, features = [
   "tls12",
   "ring",
 ] }
 ryu = "1.0.18"
 tokio = "1"
 tracing = "0.1"
-webpki-roots = "0.26.3"
+webpki-roots = "0.26.6"
```
Sytten, line 54: This is not and we should probably switch to "0.26"

modules/llrt_http/Cargo.toml
```diff
@@ -44,17 +44,17 @@ once_cell = "1"
...
 
 [dev-dependencies]
 llrt_compression = { version = "0.3.0-beta", path = "../../libs/llrt_compression" }
 llrt_test = { path = "../../libs/llrt_test" }
 tokio = { version = "1", features = ["full"] }
-wiremock = "0.6.0"
+wiremock = "0.6.2"
```
Sytten, line 60: This is fine since it's a dev dep

modules/llrt_http/Cargo.toml
```diff
@@ -21,16 +21,16 @@ compression-c = ["llrt_compression/all-c"]
 compression-rust = ["llrt_compression/all-rust"]
 
 [dependencies]
-bytes = "1.6.1"
+bytes = "1.8.0"
 either = "1"
 http-body-util = "0.1.2"
-hyper = { version = "1.4.1", features = ["client"] }
-hyper-rustls = { version = "0.27.2", default-features = false, features = [
+hyper = { version = "1.5.0", features = ["client"] }
```
Sytten, line 27: `"1"` should be sufficient here

richarddavison:
> General comment but I would really like us to be more relaxed in the dependencies of the modules. The general rule for libraries are:
> 
> 1. If it is not v1, then update the minor when it comes out. If possible keep a range of version (`>=0.12, <=0.13`)
> 2. If v1, take the oldest minor version you want to support (usually the one when you started the lib) and don't update it unless you need a new feature from the lib
> 3. The patch version is only specified if you really need a specific patch
> 
> The idea is that if all libs are agressive in pushing newer versions, you force the consumer to update their version without really a reason for it.

Agree. I'll cleanup deps to be more relaxed

Issue #641, opened 2024-10-22
Can't import CJS modules via ESM import statement


nabetti1720:
While verifying the loading of the Node.js package, we encountered another strange event.

```javascript
// logger.js
import { Logger } from '@aws-lambda-powertools/logger'
const logger = new Logger()
logger.info('Hello world')
```
```
% llrt logger.js 
SyntaxError: Could not find export 'default' in module '/Users/shinya/Workspaces/llrt-test/node_modules/lodash.merge/./index.js'
  at <anonymous> (eval_script:1:5)

% node logger.js 
{"level":"INFO","message":"Hello world","sampling_rate":0,"service":"service_undefined","timestamp":"2024-10-22T13:47:00.450Z"}

% bun logger.js 
{"level":"INFO","message":"Hello world","sampling_rate":0,"service":"service_undefined","timestamp":"2024-10-22T13:50:09.638Z"}
```

While investigating reproduction methods, we found that the same event reproduced itself in the following application structure.

```javascript
// hello.js
var hello = function world() {
  return "hello world"
}
module.exports = hello;
```
```javascript
// main.js
import hello from './hello.js'
console.log(hello())
```

```
% llrt main.js
SyntaxError: Could not find export 'default' in module '/Users/shinya/Workspaces/llrt-test/./hello.js'
  at <anonymous> (eval_script:1:5)

% node main.js  
file:///Users/shinya/Workspaces/llrt-test/main.js:1
import hello from './hello.js'
       ^^^^^
SyntaxError: The requested module './hello.js' does not provide an export named 'default'
    at ModuleJob._instantiate (node:internal/modules/esm/module_job:171:21)
    at async ModuleJob.run (node:internal/modules/esm/module_job:254:5)
    at async onImport.tracePromise.__proto__ (node:internal/modules/esm/loader:483:26)
    at async asyncRunEntryPointWithESMLoader (node:internal/modules/run_main:117:5)

Node.js v22.9.0

% bun main.js 
hello world
```

The same event occurs with Node.js in this way. However, the behavior changes when `hello.js` is changed to `hello.cjs`.

```javascript
// main.js
import hello from './hello.cjs' // The content is exactly the same as hello.js
console.log(hello())
```

```
% llrt main.js 
SyntaxError: Could not find export 'default' in module '/Users/shinya/Workspaces/llrt-test/./hello.cjs'
  at <anonymous> (eval_script:1:5)

% node main.js 
hello world

% bun main.js 
hello world
```

[...]

PR #664, opened 2024-11-02
fix: CJS module import via ESM import

nabetti1720:
### Issue # (if available)

#641

### Description of changes

- Allow import as a CJS module even if the extension is `js` or `mjs`.

### Checklist

### DIFF
--- a/llrt_core/src/module_loader/loader.rs
+++ b/llrt_core/src/module_loader/loader.rs
@@ -148,7 +148,7 @@ impl CustomLoader {
 
                 return Ok((Module::declare(ctx, path, json)?, None));
             }
-            if name.ends_with(".cjs") {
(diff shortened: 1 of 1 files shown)


richarddavison:
This will load all (non-embedded) modules as CJS which is wrong. Right now it works because ES modules in QuickJS are not loaded asynchronously. But this might change in the future causing a slowdown.

nabetti1720:
> This will load all (non-embedded) modules as CJS which is wrong. Right now it works because ES modules in QuickJS are not loaded asynchronously. But this might change in the future causing a slowdown.

Thanks for your comment. Unfortunately, I don't have an idea to fix it, so I will close the PR.

In addition, the issue we cited is already closed, making it difficult to track what else needs to be done. I would like to start a new issue.

PR #665, opened 2024-11-02
feat: Implementation of process.versions.node

nabetti1720:
### Description of changes

- Some Node.js packages use `process.versions.node` as a decision condition.
For example : https://www.npmjs.com/package/cls-hooked?activeTab=readme
- Having a provisional version in `process.versions.node` takes compatibility a little further.
- `22.11.0` is the latest LTS version of Node.js at the time of implementation. If there are no inconveniences, there is no need to change it in the future.

### Checklist

### DIFF
--- a/modules/llrt_process/src/lib.rs
+++ b/modules/llrt_process/src/lib.rs
@@ -67,10 +67,8 @@ pub fn init(ctx: &Ctx<'_>) -> Result<()> {
     let process = Object::new(ctx.clone())?;
     let process_versions = Object::new(ctx.clone())?;
     process_versions.set("llrt", VERSION)?;
-    // Node.js version - We need to set up a version that meets all the conditions listed below.
-    // [Node.js package] required version : reason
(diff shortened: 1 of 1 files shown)


nabetti1720:
I believe this error is unrelated to this correction.

image

richarddavison:
Thanks for the PR. However, I'm not to sure about this. This might assume some Node APIs are available when they're not. Bun and Deno does this but they aim to be compatible with Node so different use case. What we can do is to provide a "n/a" value or similar to somewhat increase compatibility.

nabetti1720:
> Thanks for the PR. However, I'm not to sure about this. This might assume some Node APIs are available when they're not. Bun and Deno does this but they aim to be compatible with Node so different use case. What we can do is to provide a "n/a" value or similar to somewhat increase compatibility.

Thank you for your comment.
Since `process.versions.node` is supposed to be a semver, `n/a` did not solve the problem. I think it would be possible to direct them to the fallback routine by making the version `0.0.0`, but what do you think? I think we can at least state that we do not support all Node.js APIs from now on.

richarddavison (approved):

Issue #628, opened 2024-10-14
Problems with 'aws-xray-sdk-core'


micro-jumbo:
Hi,

I have this small lambda function: 
```typescript
import {ListBucketsCommand, S3Client} from "@aws-sdk/client-s3";

export const handle = async () => {
    const s3Client = new S3Client();
    const rest = await s3Client.send(new ListBucketsCommand());
    rest.Buckets?.forEach(bucket => {
        console.log(bucket.Name);
    });
    console.log('Hello world');
}
```
which works great.

But when I want to capture calls to S3 with X-Ray traces using the `aws-xray-sdk-core` library:
```typescript
import {captureAWSv3Client} from 'aws-xray-sdk-core';
...
    const s3Client = captureAWSv3Client(new S3Client());
....
```

it starts failing with error:
```
2024-10-14T20:11:17.555Z	n/a	FATAL	ReferenceError: Error resolving module '/var/task/assert' from '/var/task/index.mjs'
  at <anonymous> (/var/task/index.mjs:1:4249)
      at <anonymous> (/var/task/index.mjs:1:458)
      at <anonymous> (/var/task/index.mjs:6:16634)
      at <anonymous> (/var/task/index.mjs:1:458)
      at <anonymous> (/var/task/index.mjs:6:89512)
      at <anonymous> (/var/task/index.mjs:1:458)
      at <anonymous> (/var/task/index.mjs:6:91308)
      at <anonymous> (/var/task/index.mjs:1:458)
      at <anonymous> (/var/task/index.mjs:6:91325)
INIT_REPORT Init Duration: 62.39 ms	Phase: init	Status: error	Error Type: Runtime.ExitError
2024-10-14T20:11:18.415Z	n/a	FATAL	ReferenceError: Error resolving module '/var/task/assert' from '/var/task/index.mjs'
  at <anonymous> (/var/task/index.mjs:1:4249)
      at <anonymous> (/var/task/index.mjs:1:458)
      at <anonymous> (/var/task/index.mjs:6:16634)
      at <anonymous> (/var/task/index.mjs:1:458)
      at <anonymous> (/var/task/index.mjs:6:89512)
      at <anonymous> (/var/task/index.mjs:1:458)
      at <anonymous> (/var/task/index.mjs:6:91308)
      at <anonymous> (/var/task/index.mjs:1:458)
      at <anonymous> (/var/task/index.mjs:6:91325)
[...]

PR #668, opened 2024-11-07
feat: Initial implementation of assert module

nabetti1720:
### Issue # (if available)

#628

### Description of changes

The following Node.js APIs, which are also used in the libraries on which `aws-xray-sdk-core` depends, have been implemented. Note that we do not intend to support all APIs from the beginning.

- assert.ok()

### Checklist

### DIFF
--- a/modules/llrt_assert/src/lib.rs
+++ b/modules/llrt_assert/src/lib.rs
@@ -31,13 +31,14 @@ fn assert(ctx: Ctx, value: Value, message: Opt<Value>) -> Result<()> {
     }
 
     if let Some(obj) = message.0 {
-        if let Some(msg) = obj.as_string() {
-            let msg = msg.to_string().unwrap();
-            return Err(Exception::throw_message(&ctx, &msg));
-        }
-        if let Some(err) = obj.as_exception() {
-            return Err(err.clone().throw());
-        }
+        match obj.type_of() {
+            Type::String => {
+                let msg = obj.as_string().unwrap().to_string().unwrap();
+                return Err(Exception::throw_message(&ctx, &msg));
+            },
+            Type::Exception => return Err(obj.as_exception().cloned().unwrap().throw()),
+            _ => {},
+        };
     }
 
     Err(Exception::throw_message(
--- a/modules/llrt_assert/src/lib.rs
+++ b/modules/llrt_assert/src/lib.rs
@@ -4,21 +4,30 @@ use llrt_utils::module::{export_default, ModuleInfo};
 use rquickjs::{
     module::{Declarations, Exports, ModuleDef},
     prelude::{Func, Opt},
-    Ctx, Exception, Result, Value,
+    Ctx, Exception, Result, Type, Value,
 };
 
-fn ok(ctx: Ctx, value: Value, message: Opt<Value>) -> Result<()> {
-    if value.as_bool().unwrap_or(false) {
-        return Ok(());
-    }
-    if value.as_number().unwrap_or(0.0) != 0.0 {
-        return Ok(());
-    }
-    if value.is_string() && !value.as_string().unwrap().to_string().unwrap().is_empty() {
-        return Ok(());
-    }
-    if value.is_array() || value.is_object() {
-        return Ok(());
+fn assert(ctx: Ctx, value: Value, message: Opt<Value>) -> Result<()> {
+    match value.type_of() {
+        Type::Bool => {
+            if value.as_bool().unwrap() {
+                return Ok(());
+            }
+        },
+        Type::Float | Type::Int => {
+            if value.as_number().unwrap() != 0.0 {
+                return Ok(());
+            }
+        },
+        Type::String => {
+            if !value.as_string().unwrap().to_string().unwrap().is_empty() {
+                return Ok(());
+            }
+        },
+        Type::Array | Type::Object => {
+            return Ok(());
+        },
+        _ => {},
     }
 
     if let Some(obj) = message.0 {
@@ -49,7 +58,7 @@ impl ModuleDef for AssertModule {
 
     fn evaluate<'js>(ctx: &Ctx<'js>, exports: &Exports<'js>) -> Result<()> {
         export_default(ctx, exports, |default| {
-            default.set("ok", Func::from(ok))?;
+            default.set("ok", Func::from(assert))?;
 
             Ok(())
         })
--- a/modules/llrt_assert/src/lib.rs
+++ b/modules/llrt_assert/src/lib.rs
@@ -1,0 +1,66 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+use llrt_utils::module::{export_default, ModuleInfo};
+use rquickjs::{
+    module::{Declarations, Exports, ModuleDef},
+    prelude::{Func, Opt},
+    Ctx, Exception, Result, Value,
+};
+
+fn ok(ctx: Ctx, value: Value, message: Opt<Value>) -> Result<()> {
+    if value.as_bool().unwrap_or(false) {
+        return Ok(());
+    }
+    if value.as_number().unwrap_or(0.0) != 0.0 {
+        return Ok(());
+    }
+    if value.is_string() && !value.as_string().unwrap().to_string().unwrap().is_empty() {
+        return Ok(());
+    }
+    if value.is_array() || value.is_object() {
+        return Ok(());
+    }
+
+    if let Some(obj) = message.0 {
+        if let Some(msg) = obj.as_string() {
+            let msg = msg.to_string().unwrap();
+            return Err(Exception::throw_message(&ctx, &msg));
+        }
+        if let Some(err) = obj.as_exception() {
+            return Err(err.clone().throw());
+        }
+    }
+
+    Err(Exception::throw_message(
+        &ctx,
+        "The expression was evaluated to a falsy value",
+    ))
+}
+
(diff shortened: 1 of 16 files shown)


modules/llrt_assert/src/lib.rs
```diff
@@ -0,0 +1,66 @@
...
+        if let Some(err) = obj.as_exception() {
+            return Err(err.clone().throw());
+        }
+    }
+
+    Err(Exception::throw_message(
+        &ctx,
+        "The expression was evaluated to a falsy value",
+    ))
+}
```
richarddavison, line 38: A more performant way would be to match on `value.type_of();` That way we don't have to cast and recheck for all types.

modules/llrt_assert/src/lib.rs
```diff
@@ -0,0 +1,66 @@
...
+
+pub struct AssertModule;
+
+impl ModuleDef for AssertModule {
+    fn declare(declare: &Declarations) -> Result<()> {
+        declare.declare("ok")?;
+
+        declare.declare("default")?;
+        Ok(())
```
richarddavison, line 47: I think `assert` is also a function. So should basically export a function that also has additional properties, similar to this javascript code:
```javascript
function assert(...){
//body
}
assert.ok = assert
```
etc

richarddavison (changes requested):
Thanks for the PR. I think a basic assert module is good to have!

modules/llrt_assert/src/lib.rs
nabetti1720: I agree. However, I am having a little trouble with the implementation.

In Node.js, we cannot call `assert` until we require the assert module. I understand that this means that it is not defined in globalThis.

```javascript
// index16.cjs
// const assert = require('assert');
assert(true);
assert.ok(true);
```

```
% node index16.cjs
/Users/shinya/Workspaces/llrt-test/index16.cjs:2
assert(true);
^

ReferenceError: assert is not defined
    at Object.<anonymous> (/Users/shinya/Workspaces/llrt-test/index16.cjs:2:1)
    at Module._compile (node:internal/modules/cjs/loader:1546:14)
    at Module._extensions..js (node:internal/modules/cjs/loader:1691:10)
    at Module.load (node:internal/modules/cjs/loader:1317:32)
    at Module._load (node:internal/modules/cjs/loader:1127:12)
    at TracingChannel.traceSync (node:diagnostics_channel:315:14)
    at wrapModuleLoad (node:internal/modules/cjs/loader:217:24)
    at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:166:5)
    at node:internal/main/run_main_module:30:49

Node.js v22.9.0
```
Is there a good implementation of how to define it so that it acts like a function when called on the variable name itself, but also has an “ok” method? I could not find a good example from the LLRT module implementation.

richarddavison: Functions are just objects. You can add whatever property you want to them.
So you can basically do:

You should probably don't use the `export_default` util here but rather to something like:
```rust
fn evaluate<'js>(ctx: &Ctx<'js>, exports: &Exports<'js>) -> Result<()> {
  
    let ok_function = Function::new(ctx.clone(),ok)?; 
    ok_function.set("ok", ok_function.clone())?;

    exports.export("ok", ok_function.clone())?;
    exports.export("default", ok_function)?;
 }

```

nabetti1720: I have tried, but it does not seem to be working as expected...

```javascript
// index16.js
const assert = require('assert');
console.log(assert);
assert.ok(true);
assert.default(true);
assert(true);
```

```
% llrt index16.cjs
{
  default: [function: (anonymous)],
  ok: [function: (anonymous)]
}
TypeError: not a function
  at <anonymous> (__cjs:/Users/shinya/Workspaces/llrt-test/index16.cjs:5:1)
```

richarddd: It does work when you import as ESM right? This might be a CJS limitation as we have assert now as a module and importing modules from require wraps them in default. This might be wrong. The tricky part is that if we require ESM modules, what should we return? One solution would be to return the default object but with other properties added to that. However that falls short if we export a non object. For instance, what are you expected to return as an object from the following file when requiring it:
```javascript 
export default 1
export const foo = 2
```
I would expect:
```javascript
{
  foo: 2,
  default: 1
}
```
Do you see the problem?

nabetti1720: Thank you for your input. I tried running the following reproduction code, but the results did not change.

```rust
// AssertModule definition section
impl ModuleDef for AssertModule {
    fn declare(declare: &Declarations) -> Result<()> {
        declare.declare("ok")?;
        declare.declare("default")?;
        Ok(())
    }

    fn evaluate<'js>(ctx: &Ctx<'js>, exports: &Exports<'js>) -> Result<()> {
        let ok_function = Function::new(ctx.clone(), ok)?;
        ok_function.set("ok", ok_function.clone())?;

        exports.export("ok", ok_function.clone())?;
        exports.export("default", ok_function)?;
        Ok(())
    }
}
```

```javascript
// index16.js
import * as assert from 'assert';
console.log(assert);
assert.ok(true);
assert.default(true);
assert(true);
```

```
% llrt index16.js
{
  default: [function: (anonymous)],
  ok: [function: (anonymous)]
}
TypeError: not a function
  at <anonymous> (/Users/shinya/Workspaces/llrt-test/index16.js:5:1)
```

I must confess that I do not fully understand the problem and find it difficult for me to solve it.

We believe that once assert.ok has been implemented, the initial goal of PR has been achieved. Can this issue be treated as a separate issue after the PR has been merged?

richarddd: You should do this:
```javascript 
import assert from 'assert';
console.log(assert);
assert.ok(true);
assert.default(true);
assert(true);
```

nabetti1720: I'm sorry. It worked. The last commit will correctly handle the following reproduced code.

```javascript
// index16.js
import assert from 'assert';
console.log(assert);
assert.ok(true);
assert(true);
```

```
% llrt index16.js
[function: (anonymous)]
```

richarddavison: @nabetti1720 please rebase

nabetti1720:
The following verification is still the same before and after the merge of #673 results, with no problems.

We also received the following error, which we recognize is unrelated to this correction.
image

modules/llrt_assert/src/lib.rs
```diff
@@ -0,0 +1,77 @@
...
+        },
+        Type::String => {
+            if !value.as_string().unwrap().to_string().unwrap().is_empty() {
+                return Ok(());
+            }
+        },
+        Type::Array | Type::Object => {
+            return Ok(());
+        },
+        _ => {},
```
richarddavison, line 30: Some missing cases. What about `constructor`,`function`, `bigint`, `symbol`, `Exception`.

modules/llrt_assert/src/lib.rs
```diff
@@ -0,0 +1,77 @@
...
+            },
+            Type::Exception => return Err(obj.as_exception().cloned().unwrap().throw()),
+            _ => {},
+        };
+    }
+
+    Err(Exception::throw_message(
+        &ctx,
+        "The expression was evaluated to a falsy value",
+    ))
```
richarddavison, line 47: Ideally we should ha an `AssertionError` but that can be added later. It might be good to show wha the value was by coercing it to string

richarddavison (changes requested):
Some comments

nabetti1720:
@richarddavison ,  Following this `fix: CJS export default (#673)` , the following modules can no longer be loaded.

Specifically, the `fix: CJS export default (#673)` will cause an IO error when loading modules, and even if you fix it, an empty object will be returned.

```javascript
// index17.js
const a = require('call-bind/callBound');
console.log(a)
```

Specifically, the "Fix default CJS exports” will cause an IO error when loading modules, and even if you fix it, an empty object will be returned.

Previous state (Commit#0a7509de):
```
% llrt index17.js
[function: callBind]
```

main branch (Commit#6dcd7784):
```
% llrt index17.js  
Error: IO Error: No such file or directory (os error 2)
  at <anonymous> (/Users/shinya/Workspaces/llrt-test/index17.js:1:11)
```
Commit#2d0e3381:
```
% llrt index17.js
{}
```

richarddavison:
> Specifically, the "Fix default CJS exports” will cause an IO error when loading modules, and even if you fix it, an empty object will be returned.
> 
Fixed by https://github.com/awslabs/llrt/pull/674/files

nabetti1720:
> > Specifically, the "Fix default CJS exports” will cause an IO error when loading modules, and even if you fix it, an empty object will be returned.
> 
> Fixed by https://github.com/awslabs/llrt/pull/674/files

Thank you very much. After rebese, this correction has been reverted.

tests/unit/assert.test.ts
```diff
@@ -0,0 +1,80 @@
...
+    try {
+      assert.ok("");
+    } catch (err) {
+      expect(err.message).toEqual(errMsg);
+    }
+    try {
+      assert.ok(null);
+    } catch (err) {
+      expect(err.message).toEqual(errMsg);
+    }
```
richarddavison, line 40: You should use expect to detect throws and not try catch:
expect(t).toThrow("message");

richarddavison (changes requested):

richarddavison:
Thank you. LGTM. Please rebase of main and i'll merge!

richarddavison (approved):

PR #672, opened 2024-11-08
fix: Improved package.json resolution logic

### LINKED ISSUE #671: IO error occurs if package.json resolution results in a directory
While verifying #670, I found different issues.

```javascript
// reproduction.js
const a = require('pg');
```
```
% llrt reproduction.js
Error: IO Error: Is a directory (os error 21)
  at <anonymous> (/Users/shinya/Workspaces/llrt-test/reproduction.js:1:11)
```

I checked the package.json of the `pg` module and found a consideration leak.

package.json:
```json
{
  "main": "./lib",
}
```

In the case of such a description, we had to refer to `index.js` directly under that directory.

nabetti1720:
### Issue # (if available)

Fixed #671

### Description of changes

- Fix package.json so that “index” is used as the name of the file directly under it when the resolved result is a directory.

### Checklist

### DIFF
--- a/llrt_core/src/module_loader/resolver.rs
+++ b/llrt_core/src/module_loader/resolver.rs
@@ -611,14 +611,22 @@ fn is_exports_field_exists<'a>(package_json: &'a BorrowedValue<'a>) -> bool {
 }
 
 fn correct_extensions<'a>(x: String) -> Cow<'a, str> {
-    if Path::new(&x).is_file() {
-        return x.into();
+    if let Ok(metadata) = fs::metadata(&x) {
+        if metadata.is_file() {
+            return x.into();
+        }
+        if metadata.is_dir() {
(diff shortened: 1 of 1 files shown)


llrt_core/src/module_loader/resolver.rs
```diff
@@ -614,8 +614,11 @@ fn correct_extensions<'a>(x: String) -> Cow<'a, str> {
     if Path::new(&x).is_file() {
         return x.into();
     }
+
+    let index = if Path::new(&x).is_dir() { "/index" } else { "" };
+
```
richarddavison, line 619: Create Path once and pull in metadata. is_dir and is_file each call metadata which adds extra cost. Better to reuse metadata and match on metadata.is_dir/is_file.

richarddavison (changes requested):

richarddavison (approved):

PR #675, opened 2024-11-13
modules: add tty

ahaoboy:
### Description of changes

Add tty module and use libc to implement tty.isatty function

### Checklist

### DIFF
--- a/modules/llrt_tty/src/lib.rs
+++ b/modules/llrt_tty/src/lib.rs
@@ -1,0 +1,80 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+use llrt_utils::module::{export_default, ModuleInfo};
+use rquickjs::{
+    module::{Declarations, Exports, ModuleDef},
+    prelude::Func,
+    Ctx, Result,
+};
+
+fn isatty(fd: i32) -> bool {
+    unsafe { libc::isatty(fd) != 0 }
+}
+
+pub struct TtyModule;
+
+impl ModuleDef for TtyModule {
+    fn declare(declare: &Declarations<'_>) -> Result<()> {
+        declare.declare("isatty")?;
+        declare.declare("default")?;
+        Ok(())
+    }
+
+    fn evaluate<'js>(ctx: &Ctx<'js>, exports: &Exports<'js>) -> Result<()> {
+        export_default(ctx, exports, |default| {
+            default.set("isatty", Func::from(isatty))?;
+            Ok(())
+        })
+    }
+}
+
+impl From<TtyModule> for ModuleInfo<TtyModule> {
+    fn from(val: TtyModule) -> Self {
+        ModuleInfo {
+            name: "tty",
+            module: val,
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use crate::TtyModule;
+    use llrt_test::{call_test, test_async_with, ModuleEvaluator};
+    use std::io::{stderr, stdin, stdout, IsTerminal};
+
+    #[tokio::test]
+    async fn test_isatty() {
+        test_async_with(|ctx| {
+            Box::pin(async move {
+                ModuleEvaluator::eval_rust::<TtyModule>(ctx.clone(), "tty")
+                    .await
+                    .unwrap();
+
+                let module = ModuleEvaluator::eval_js(
+                    ctx.clone(),
+                    "test",
+                    r#"
+                        import { isatty } from 'tty';
+
+                        export async function test() {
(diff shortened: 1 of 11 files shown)


richarddavison:
@ahaoboy tests are failing because you need to mark it as external in esbuild config inside` build.mjs`

tests/unit/tty.test.ts
```diff
@@ -0,0 +1,18 @@
...
+      isatty(2),
+      isatty(3),
+      isatty(4),
+    ]).toEqual([
+      false,
+      false,
+      false,
+      false,
+    ]);
+  });
```
nabetti1720, line 17: Do I get the same results if I run it on the laptop at hand?

ahaoboy: The results of running the command line are correct. I suspect that llrt test uses redirection or other unknown behaviors, which causes all false to be returned in the test.

```
import { isatty } from 'tty'

console.log(
  isatty(1),
  isatty(2),
  isatty(3),
  isatty(4),
)
```

```
llrt ./a.js

true true false false
```

nabetti1720: I figured that would probably be the case. This is a little troubling because I always run the test suite on the laptop at hand to make sure it is error free before shipping...

ahaoboy: Now only the return value type is checked

richarddavison: I think a way to properly test this is to add a rust test for this (we can skip js test) where we launch a child process to dermine if file descriptor is tty. This has the be adjusted for posix and windows

richarddavison (changes requested):
Thanks for this. Please add tests in rust (see comment)

tests/unit/tty.test.ts
ahaoboy: Can you explain it in more detail? The difficulty in testing is that isatty always returns false in github action
https://github.com/ahaoboy/tty-test/actions/runs/11886599988/job/33118007342

ahaoboy: The easiest way I've found to test this is to distinguish whether the program is running in GitHub action or not.

ahaoboy:
https://github.com/awslabs/llrt/actions/runs/11888714370/job/33123869729

Need to fix the resolve error on Windows first

nabetti1720:
@ahaoboy, #690 has been merged into the mainline and the require issue should be resolved. Please rebase. :)

tests/unit/tty.test.ts
richarddavison: You can run "test -t" from a rust child process:
```rust
let status = Command::new("test")
        .arg("-t")
        .arg("1") // File descriptor 1 is stdout,
        .status(); 
```

modules/llrt_tty/src/lib.rs
```diff
@@ -0,0 +1,77 @@
...
+                        import { isatty } from 'tty';
+
+                        export async function test() {
+                            return new Array(5).fill(0).map((_, i) => +isatty(i)).join('')
+                        }
+                    "#,
+                )
+                .await
+                .unwrap();
+                let expect = if std::env::var("GITHUB_ACTIONS").unwrap_or_default() == "true" {
```
richarddavison, line 66: We can't hard code GITHUB_ACTIONS to test this. Better would be to run the `test -t` command as a subcommand and test this on posix only for now. See comment

richarddavison (changes requested):

modules/llrt_tty/src/lib.rs
ahaoboy: I'm not sure if the two are equivalent, but using is_terminal removes libc

```
fn isatty(fd: i32) -> bool {
    unsafe { libc::isatty(fd) != 0 }
}
```

```
fn isatty(fd: i32) -> bool {
    match fd {
        0 => stdin().is_terminal(),
        1 => stdout().is_terminal(),
        2 => stderr().is_terminal(),
        _ => false,
    }
}
```

tests/unit/tty.test.ts
richarddavison, line 1: Remove this test since we test in rust

richarddavison (changes requested):

tests/unit/tty.test.ts
ahaoboy: Do you mean using is_terminal to test the isatty function in rust?

```
    #[test]
    fn test_isatty() {
        let expect = [
            stdin().is_terminal(),
            stdout().is_terminal(),
            stderr().is_terminal(),
            false,
            false,
        ];
        assert_eq!(
            expect,
            [isatty(0), isatty(1), isatty(2), isatty(3), isatty(4)]
        );
    }
```

richarddavison (approved):

PR #678, opened 2024-11-15
ci: upload artifacts on pull_request

ahaoboy:
### Description of changes

upload artifacts on pull_request to compare performance and size changes caused by PR

https://boajs.dev/benchmarks

It is also possible to implement observation diagrams similar to boa

### Checklist

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -131,7 +131,7 @@ jobs:
         if: inputs.release
         uses: actions/upload-artifact@v4
         with:
-          name: artifacts-${{ inputs.platform }}-${{ inputs.arch }}
+          name: llrt-${{ inputs.platform }}-${{ inputs.arch }}
           path: |
             *.zip
             llrt-container-arm64*
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -3,6 +3,7 @@ on:
   push:
     tags:
       - "v*.*.*"
+  pull_request:
 
 jobs:
   build:
@@ -40,6 +41,7 @@ jobs:
       release: ${{ matrix.release }}
       toolchain: ${{ matrix.toolchain }}
   release:
+    if: startsWith(github.ref, 'refs/tags/')
     permissions:
       contents: write
       discussions: write


ahaoboy:
https://github.com/awslabs/llrt/releases/download/v0.3.0-beta/llrt-linux-x64-full-sdk.zip

https://github.com/awslabs/llrt/actions/runs/11850925586

qjs-ng win~

```
 ./llrt ./dist/run.js
Richards: 787
DeltaBlue: 758
Crypto: 902
RayTrace: 1210
EarleyBoyer: 2281
RegExp: 179
Splay: 1618
NavierStokes: 1467
----
Score: 944

./llrt-ng ./dist/run.js
Richards: 797
DeltaBlue: 815
Crypto: 938
RayTrace: 1226
EarleyBoyer: 2255
RegExp: 177
Splay: 1728
NavierStokes: 1551
----
Score: 973
```

richarddavison:
Hi @ahaoboy! We had this before and it's a good suggestion but running the release flow is not the way to do it as it causes significant increase in PR checks. A better way would be to add a "make bench" that runs benchmarks from the `benchmarks` directory
```shell
cargo $(TOOLCHAIN) run -r --target $(CURRENT_TARGET) -- benchmark/v8-v7
cargo $(TOOLCHAIN) run -r --target $(CURRENT_TARGET) -- benchmark/microbench.js
```

Microbench can be found in quicks-ng repo and needs some slight adjustments for LLRT.

ahaoboy:
> Hi @ahaoboy! We had this before and it's a good suggestion but running the release flow is not the way to do it as it causes significant increase in PR checks.

This pr only uploads artifacts, no benchmark, but it will indeed run one more action

richarddavison (commented):
If finding it hard to see a reason why artifacts should be valuable in PR pipeline?

ahaoboy:
> If finding it hard to see a reason why artifacts should be valuable in PR pipeline?

The release cycle is very long, developers who want to use the latest version can only compile it themselves, it is easier to download it directly

richarddavison:
> The release cycle is very long, developers who want to use the latest version can only compile it themselves, it is easier to download it directly

Right, but this seems like an X/Y problem. We should rather shorten the release cycle instead of adding CI artifacts for PRs :) However, we can add nightly versions. But then we should only upload artifacts from main `branch` and not every PR and suffix them with `-nighty` version or similar.

ahaoboy:
> we can add nightly versions. But then we should only upload artifacts from main `branch` and not every PR and suffix them with `-nighty` version or similar.

Agreed, compiling aarch64-unknown-linux-musl on windows is very difficult, it would be great if there were nightly builds

ahaoboy:
> Feel free to change this PR to apply this proposal or open up a new one!


I will try to add it in my fork branch, and if it works, I will open new PR

PR #682, opened 2024-11-16
refactor: context, encoding, json and numbers crate

nabetti1720:
### Description of changes

This PR reviewed the crate to improve future maintainability.

- llrt_json: Until now this was managed in `modules`, but since it is not a module, it will be managed in `libs`.
- llrt_numbers: It will be separated from `core` and managed in `libs`.
- llrt_context / llrt_encoding: Previously this was controlled by a feature flag in llrt_utils, but now it's an independent module for easier use.
- benches: This was also in `core`, but we have increased its independence by managing the module you want to benchmark.

As for `llrt_stream`, it is best to manage it in `libs`, since it is not currently in a format that can be loaded as a module. However, in the future, it will be attached to globalThis, so we are leaving it as it is. :)

Also, `llrt_utils` still has two functions, `fs` and `bytearray_buffer`, but I'm putting them on hold for now because I had trouble naming them. (llrt_fsutil / llrt_bytearray_buffer?) We do not plan to include it in this PR.

### Checklist


Sytten (approved):
This is going in the right direction and it was my original vision too. I am glad the community is onboard. We should continue splitting the llrt_utils into separate libraries

richarddavison (approved):

PR #683, opened 2024-11-17
fix: Improved behavior when loading CJS modules from import statements

### LINKED ISSUE #667: Need improvement on calling as CJS module from ESM
Calling the CJS module from the ESM has been improved by the resolution of #641, but we need further improvements.

See also.
https://github.com/awslabs/llrt/issues/641#issuecomment-2450988284

In addition, the following PRs would be helpful as a path that should not be taken.
https://github.com/awslabs/llrt/pull/664#issuecomment-2453113856

--
This is by design. We need to be able to distinguish cjs from regular mjs files. I’ve considered options for trying to automatically detect this or simply treat all require imports as CJS. What’s your take?

_Originally posted by @richarddavison in https://github.com/awslabs/llrt/issues/641#issuecomment-2450550500_

nabetti1720:
### Issue # (if available)

Fixed #667
Related #641 

### Description of changes

Fixed an issue where the error `SyntaxError: Could not find export 'default' from '...'` occurred when loading a CJS module from an import statement.

- When resolving a Node.js Package, we now also determine whether the package is CJS.
- After that, if we determine that it is a CJS module called by import statement, we now use prefix `__cjsm:` to achieve this behavior.

The following use case, which was an issue in #641, will now work:

```javascript
// powertoolsLogger.js
import { Logger } from '@aws-lambda-powertools/logger'
const logger = new Logger()
logger.info('Hello world')
```

```
% llrt powertoolsLogger.js 
{"level":"INFO","message":"Hello world","sampling_rate":0,"service":"service_undefined","timestamp":"2024-11-17T01:52:13.459Z"}
```

### Checklist

### DIFF
--- a/llrt_core/src/module_loader/resolver.rs
+++ b/llrt_core/src/module_loader/resolver.rs
@@ -510,11 +510,16 @@ fn load_package_exports<'a>(
         return Ok(sub_module.into());
     }
 
-    let module_path = package_exports_resolve(&package_json, name, is_esm)?;
+    let (module_path, is_cjs) = package_exports_resolve(&package_json, name, is_esm)?;
 
-    Ok(correct_extensions(
-        [dir, "/", scope, "/", module_path].concat(),
-    ))
+    let module_path = correct_extensions([dir, "/", scope, "/", module_path].concat());
+    let prefix = if is_cjs && is_esm {
+        CJS_LOADER_PREFIX
+    } else {
+        ""
+    };
+
+    Ok([prefix, &module_path].concat().into())
 }
 
 // LOAD_PACKAGE_SELF(X, DIR)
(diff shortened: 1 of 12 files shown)


llrt_core/src/module_loader/resolver.rs
```diff
@@ -548,7 +553,7 @@ fn load_package_self(ctx: &Ctx<'_>, x: &str, dir: &str, is_esm: bool) -> Result<
     //    "." + X.slice("name".length), `package.json` "exports", ["node", "require"])
     //    <a href="esm.md#resolver-algorithm-specification">defined in the ESM resolver</a>.
     // 6. RESOLVE_ESM_MATCH(MATCH)
-    if let Ok(path) = package_exports_resolve(&package_json, &name, is_esm) {
+    if let Ok((path, _)) = package_exports_resolve(&package_json, &name, is_esm) {
```
nabetti1720, line 556: We don't yet know if we need the same implementation here. Even if we did, the path is still relative, so adding a prefix at this point would be a mistake.

nabetti1720:
It seems that a similar issue is now being reported in QuickJS-NG (numerous warnings occur when building in Visual Studio 2022)...

image

richarddavison:
I think it would be valuable to add some tests for require to handle edge cases. We can create a mock `node_modules` folder in `fixtures` and add fake packages in there (not real packages)

nabetti1720:
> I think it would be valuable to add some tests for require to handle edge cases. We can create a mock `node_modules` folder in `fixtures` and add fake packages in there (not real packages)

I've written more specific test code for some of the require patterns that have been supported so far. This PR covers the patterns that can confirm this use case and the patterns where regressions have recently occurred.

richarddavison (approved):

PR #689, opened 2024-11-18
Improve resolve on windows

ahaoboy:
### Description of changes

The following test fails on Windows due to the path separator

```
  _require(`${CWD}/fixtures/test_modules/test-uuid.js`);
```

### Checklist

### DIFF
--- a/llrt_core/src/module_loader/resolver.rs
+++ b/llrt_core/src/module_loader/resolver.rs
@@ -61,8 +61,8 @@ static FILESYSTEM_ROOT: Lazy<Box<str>> = Lazy::new(|| {
 
 #[derive(Debug, Default)]
 pub struct CustomResolver;
-
 #[allow(clippy::manual_strip)]
+#[allow(clippy::bind_instead_of_map)]
 impl Resolver for CustomResolver {
     fn resolve(&mut self, ctx: &Ctx, base: &str, name: &str) -> Result<String> {
         if name.starts_with(CJS_IMPORT_PREFIX) {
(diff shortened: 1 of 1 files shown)


llrt_core/src/module_loader/resolver.rs
```diff
@@ -73,7 +73,13 @@ impl Resolver for CustomResolver {
...
         trace!("Try resolve '{}' from '{}'", name, base);
 
-        require_resolve(ctx, name, base, true).map(|name| name.into_owned())
+        require_resolve(ctx, name, base, true).map(|name| {
+            if cfg!(not(windows)) {
+                name.into_owned()
+            } else {
+                name.replace("\\", "/")
+            }
+        })
```
richarddavison, line 82: We have a "to_abs_path" used by `require_resolve` in various places. Use that method inside of require resolve where this case isn't covered!

richarddavison (changes requested):

ahaoboy:
> The results returned by `require_resolve()` (build-in module name, bytecode cache name, absolute path including search results from package.json, presence or absence of prefix used on the loader side, etc.) vary, so I think it's difficult to judge here.
> 
> I'll take a look at this if you don't mind. :)

The logic is indeed a bit complicated, and I'm not sure whether it should be handled here. If there is a better way, you can directly submit a new PR

nabetti1720:
@ahaoboy , This issue was addressed in #690. We have confirmed that the Require tests pass at least in windows-build.

image

nabetti1720:
@ahaoboy, #690 has been merged into mainline, so this PR is no longer needed. :)

PR #680, opened 2024-11-15
feat: Support imports field in package.json

nabetti1720:
### Issue # (if available)

Closed #677

### Description of changes

This PR addresses basic loading of the imports field in package.json.

import:
```javascript
//  index19.js
import chalk from 'chalk';
console.log(chalk.blue('Hello world!'));
```
```
% llrt index19.js
Hello world!
```

require:
```javascript
//  index19.js
const chalk = require('chalk');
console.log(chalk.blue('Hello world!'));
```
```
% llrt index19.js
Hello world!
```

### Checklist

- [x] Created unit tests in `tests/unit` and/or in Rust for my feature if needed
- [x] Ran `make fix` to format JS and apply Clippy auto fixes
- [x] Made sure my code didn't add any additional warnings: `make check`
- [x] Added relevant type info in `types/` directory
- [x] Updated documentation if needed ([API.md](API.md)/[README.md](README.md)/Other)

_By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice._

PR #690, opened 2024-11-18
feat: Resolver control by environment variable "LLRT_PLATFORM"

nabetti1720:
### Issue # (if available)

https://github.com/awslabs/llrt/pull/680#issuecomment-2482365305

### Description of changes

This PR adds the ability to toggle platform preferences in the Node.js package resolver.

```javascript
// index19.js
import chalk from 'chalk';
console.log(chalk.blue('Hello world!'));
```

Use "browser":
```
% export LLRT_PLATFORM=browser
% llrt index19.js
Hello world!
```

Use "node":
```
% export LLRT_PLATFORM=node
% llrt index19.js
ReferenceError: Error resolving module 'node:tty' from '/Users/shinya/Workspaces/llrt-test/node_modules/chalk/./source/vendor/supports-color/index.js'
```
NOTE: This behavior is currently as expected, as the tty module has not yet been merged into mainline.

Non use:
```
% unset LLRT_PLATFORM
% llrt index19.js
Hello world!
```

### Checklist

### DIFF
--- a/llrt_core/src/module_loader/mod.rs
+++ b/llrt_core/src/module_loader/mod.rs
@@ -1,3 +1,11 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+use std::env;
+
+use once_cell::sync::Lazy;
+
+use crate::environment;
+
 pub mod loader;
 pub mod resolver;
 
@@ -5,3 +13,13 @@ pub mod resolver;
 pub const CJS_IMPORT_PREFIX: &str = "__cjs:";
 // added to force CJS imports in loader
 pub const CJS_LOADER_PREFIX: &str = "__cjsm:";
+
+pub static LLRT_PLATFORM: Lazy<String> = Lazy::new(|| {
+    let platform =
+        env::var(environment::ENV_LLRT_PLATFORM).unwrap_or_else(|_| "browser".to_string());
+    if platform == "node" {
+        "node".to_string()
+    } else {
+        "browser".to_string()
+    }
+});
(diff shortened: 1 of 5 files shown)


llrt_core/src/module_loader/mod.rs
```diff
@@ -1,7 +1,25 @@
...
+
 pub mod loader;
 pub mod resolver;
 
 // added when .cjs files are imported
 pub const CJS_IMPORT_PREFIX: &str = "__cjs:";
 // added to force CJS imports in loader
 pub const CJS_LOADER_PREFIX: &str = "__cjsm:";
+
+pub static LLRT_PLATFORM: Lazy<String> = Lazy::new(|| {
```
richarddavison, line 17:
```rust
pub static LLRT_PLATFORM: Lazy<String> = Lazy::new(|| {
    env::var(environment::ENV_LLRT_PLATFORM)
        .ok()
        .filter(|platform| platform == "node")
        .unwrap_or_else(|| "browser".to_string())
});
```

richarddavison (changes requested):
Good idea! Just a minor suggestion

llrt_core/src/module_loader/mod.rs
nabetti1720: Fixed it!

nabetti1720:
@richarddavison , We have also fixed the issue where Windows paths were not being processed correctly, so please check it out.

richarddavison (changes requested):
We can avoid a few unwraps. I wonder if we should do this on all paths on after calling `require_resolve` instead of on a per function basis. Also make that function private to make sure it always return an abs path.

nabetti1720:
> We can avoid a few unwraps. I wonder if we should do this on all paths on after calling `require_resolve` instead of on a per function basis. Also make that function private to make sure it always return an abs path.

Thinking about it, in this use case, we only need to worry about the package.json resolution process, and since that function can return a Result type, we can avoid using unwrap().

richarddavison (approved):

PR #693, opened 2024-11-19
support repl

ahaoboy:
### Description of changes

Adding a simple repl mode

https://nodejs.org/en/learn/command-line/how-to-use-the-nodejs-repl

https://github.com/DelSkayn/rquickjs/blob/master/examples/rquickjs-cli/src/main.rs

### Checklist

### DIFF
--- a/llrt_core/src/vm.rs
+++ b/llrt_core/src/vm.rs
@@ -44,7 +44,7 @@ use crate::{
         resolver::{require_resolve, CustomResolver},
         CJS_IMPORT_PREFIX,
     },
-    modules::{console, crypto::SYSTEM_RANDOM},
+    modules::{console, crypto::SYSTEM_RANDOM, repl::run_repl},
     security,
     utils::clone::structured_clone,
 };
@@ -184,6 +184,11 @@ impl Vm {
 
         self.run(source, strict, global).await;
     }
+
+    pub async fn run_repl(&self) {
+        run_repl(&self.ctx).await;
+    }
+
     pub async fn run<S: Into<Vec<u8>> + Send>(&self, source: S, strict: bool, global: bool) {
         self.run_with(|ctx| {
             let mut options = EvalOptions::default();
--- a/llrt/src/main.rs
+++ b/llrt/src/main.rs
@@ -192,12 +192,7 @@ async fn start_cli(vm: &Vm) {
             }
         }
     } else {
-        let index = if let Ok(dir) = std::env::current_dir() {
-            [dir.to_string_lossy().as_ref(), "/index"].concat()
-        } else {
-            "index".into()
-        };
-        vm.run_file(index, true, false).await;
+        vm.run_repl().await;
     }
 }
 
--- a/llrt_core/src/modules/mod.rs
+++ b/llrt_core/src/modules/mod.rs
@@ -8,4 +8,5 @@ pub use llrt_modules::{
 pub mod console;
 pub mod llrt;
 pub mod module;
+pub mod repl;
 pub mod util;
--- a/llrt_core/src/modules/repl.rs
+++ b/llrt_core/src/modules/repl.rs
@@ -1,0 +1,59 @@
+use llrt_modules::VERSION;
+use llrt_utils::error::ErrorExtensions;
+use rquickjs::{prelude::Rest, AsyncContext, CatchResultExt, Ctx, Result, Value};
+
+use crate::modules::console::{self};
+
+fn process_input(ctx: &Ctx<'_>, input: &str) -> String {
+    match ctx
+        .eval::<Value, _>(input.as_bytes())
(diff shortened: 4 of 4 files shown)


llrt_core/src/vm.rs
```diff
@@ -184,6 +184,28 @@ impl Vm {
 
         self.run(source, strict, global).await;
     }
+
+    pub async fn run_repl(&self) {
+        self.ctx
+            .with(|ctx| -> Result<()> {
+                let global = ctx.globals();
+                let console: Object = global.get("console")?;
+                let js_log: Function = console.get("log")?;
```
richarddavison, line 193: We can call these methods directly from rust and print. Check the console module.

llrt_core/src/vm.rs
```diff
@@ -184,6 +184,28 @@ impl Vm {
...
+                loop {
+                    let mut input = String::new();
```
richarddavison, line 195: We don't need to create a new string very time. We can create one outside the loop and clear it.

llrt_core/src/vm.rs
```diff
@@ -184,6 +184,28 @@ impl Vm {
...
+                    print!("> ");
```
richarddavison, line 196: Maybe we also should print the version header? (see version method).

llrt_core/src/vm.rs
```diff
@@ -184,6 +184,28 @@ impl Vm {
...
+                    std::io::Write::flush(&mut std::io::stdout())?;
+                    std::io::stdin().read_line(&mut input)?;
+                    ctx.eval::<Value, _>(input.as_bytes())
+                        .and_then(|ret| js_log.call::<(Value<'_>,), ()>((ret,)))
+                        .catch(&ctx)
+                        .unwrap_or_else(|err| println!("{err}"));
```
richarddavison, line 202: I think the right approach here would be to evaluate the code and return value and pass that value to format method and write to std out

richarddavison (changes requested):
Cool idea, I like it! Some minor comment comments!

llrt_core/src/vm.rs
```diff
@@ -184,6 +185,44 @@ impl Vm {
...
+        self.ctx
+            .with(|ctx| -> Result<()> {
+                let mut input = String::new();
+                println!("Welcome to llrt v{}", VERSION);
+                loop {
+                    print!("> ");
+                    std::io::Write::flush(&mut std::io::stdout())?;
+                    std::io::stdin().read_line(&mut input)?;
+                    match ctx.eval::<Value, _>(input.as_bytes()) {
+                        Ok(v) => console::log_std_out(
```
richarddavison, line 199: Call `pub fn format` in console instead to get the formated string back. Then do `println` or `eprintln` (for exception) with that string.

richarddavison (changes requested):
Almost there! Just a minor change

llrt_core/src/vm.rs
```diff
@@ -184,6 +185,48 @@ impl Vm {
...
+    pub async fn run_repl(&self) {
+        self.ctx
+            .with(|ctx| -> Result<()> {
+                let mut input = String::new();
+                println!("Welcome to llrt v{}", VERSION);
+                loop {
+                    print!("> ");
+                    std::io::Write::flush(&mut std::io::stdout())?;
+                    std::io::stdin().read_line(&mut input)?;
+                    match ctx.eval::<Value, _>(input.as_bytes()) {
```
richarddavison, line 198: This still needs a bit of work. Something like this (pseudo code):
```rust
match ctx.eval::<Value, _>(input.as_bytes()).map(|v|console::format(&ctx, Rest(vec![v])).catch(&ctx){
  Ok(s) => println!("{s}"),
  Err(caught_err) => {
        match caught_err.into_value(ctx).map(|v|console.format(&ctx, Rest(vec![v]))){
           Ok(s) => eprintln!("Error: {s}"),
           Err(_) => eprintln!("Error: {:?}", caught_err)
        }
  }
}
```

richarddavison (changes requested):
Some suggestions to simplify. Would also be great if we can extract the repl inner logic to an own file and pass input and output to it. That way we can make it testable from rust.

ahaoboy:
> That way we can make it testable from rust.

I honestly have no idea how to test the repl, simulate keyboard input?

richarddavison:
> > That way we can make it testable from rust.
> 
> I honestly have no idea how to test the repl, simulate keyboard input?

No this can be simpler. Extract the processing of the line and just test that. We don't need to test capturing of std input.
In order to this we should probably move the entire repl capability from vm.rs and into a separate file (repl.rs). In there we can have like:
```rust
fn process_line(&line) -> std::result::Result<String,String>{
 //existing line processing logic but return strings for error and console rather than printing them stderr or stdout
}
```

And then we can test `process_line` with different input valid and invalid :) Hope it makes sense.

richarddavison (approved):
Nice thank you!

PR #698, opened 2024-11-23
feat(crypto): Initial support for SubtleCrypto

### LINKED ISSUE #184: SubtleCrypto
https://w3c.github.io/webcrypto/#dfn-SubtleCrypto
https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto?retiredLocale=sv-SE

```[tasklist]
### Related issues
- [ ] https://github.com/awslabs/llrt/issues/125
- [ ] https://github.com/awslabs/llrt/issues/208
```

nabetti1720:
### Issue # (if available)

Closed #184

### Description of changes

With this PR, we're making a small step forward with SubtleCrypto.
We're not yet fully web standards compliant, but we wanted to share our progress so far.

At the moment, the following functions are not implemented, but since it will be complicated and time-consuming from here, I would like to make it the next PR.

- exportKey() w/o `raw` mode
- importKey()
- unwrapKey()
- wrapKey()

Runtime compatibility (Results on a laptop):
image

### Checklist


Sytten:
Nice!!!
Two comments:
- We will need CryptoKey / CryptoKeyPair classes
- Most methods need to be async as let spec and we will want to use tokio spawn_blocking I think.

richarddavison:
Super nice! Yeah, async needs to use  `ctx.spawn_exit()`. For example see `random_fill` vs `random_fill_sync` in crypto module.

Sytten:
@richarddavison Not in this case I think, spawn_exit is only for running background tasks. Here we want to asyncify a blocking task, so you have to use the tokio thread pool for that. Same as when we do file operations.

nabetti1720:
> @richarddavison Not in this case I think, spawn_exit is only for running background tasks. Here we want to asyncify a blocking task, so you have to use the tokio thread pool for that. Same as when we do file operations.

There is still debate as to which parts of the implementation should be asynchronous, but for the time being, I think it will be meaningful to simply add "async" to a function in order to conform to the specifications.

For example, the following blob function is marked with "async" just to comply with the specification, even though no asynchronous processing is actually called.

https://github.com/awslabs/llrt/blob/aa742eb56778d5d10c20c7d414b849a6c58245d3/modules/llrt_http/src/blob.rs#L91-L102

nabetti1720:
> * We will need CryptoKey / CryptoKeyPair classes

I know it's necessary, but I'm not sure how to implement it yet.

For example, when you run the following code on bun, gentlekey is returned as a CryptoKey type, but when you look at it in console.log, it is an ordinary object.

```javascript
// subtleGenerateKey.js
async function exportCryptoKey(key) {
  const exported = await crypto.subtle.exportKey("raw", key);
  return new Uint8Array(exported);
}

const key = await crypto.subtle
  .generateKey(
    {
      name: "AES-GCM",
      length: 256,
    },
    true,
    ["encrypt", "decrypt"],
  );

console.log(key);
console.log(await exportCryptoKey(key));
console.log(await exportCryptoKey(key));
```
```
% bun subtleGenerateKey.js
CryptoKey {
  type: "secret",
  extractable: true,
  algorithm: {
    name: "AES-GCM",
    length: 256,
  },
  usages: [ "decrypt", "encrypt" ],
}
Uint8Array(32) [ 228, 104, 149, 176, 21, 129, 162, 92, 75, 1, 147, 65, 121, 50, 242, 28, 144, 65, 103, 203, 240, 167, 105, 60, 37, 247, 158, 156, 204, 127, 94, 16 ]
Uint8Array(32) [ 228, 104, 149, 176, 21, 129, 162, 92, 75, 1, 147, 65, 121, 50, 242, 28, 144, 65, 103, 203, 240, 167, 105, 60, 37, 247, 158, 156, 204, 127, 94, 16 ]
```

Nevertheless, since the same value is returned even if you run exportKey multiple times, this object must have hidden information (seed).

https://w3c.github.io/webcrypto/#cryptokey-interface-internal-slots

modules/llrt_crypto/Cargo.toml
```diff
@@ -30,6 +30,18 @@ uuid = { version = "1", default-features = false, features = [
...
+sha2 = "0.10"
+aes = "0.8"
+aes-gcm = "0.10"
+cbc = { version = "0.1", features = ["std"] }
+num-traits = "0.2"
+rsa = { version = "0.9", features = ["std", "sha2"], default-features = false }
+p256 = { version = "0.13", features = ["ecdh"] }
+p384 = "0.13"
+ctr = "0.9"
+
```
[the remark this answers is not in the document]
Sytten: The only thing is that ring is C code (requires a C compiler) where the others are rust code.

modules/llrt_crypto/src/subtle/decrypt.rs
```diff
@@ -0,0 +1,77 @@
...
+            )),
+        },
+        Algorithm::RsaOaep(label) => {
+            let private_key = RsaPrivateKey::from_pkcs1_der(&key).or_throw(ctx)?;
+            let padding = match label {
+                Some(buf) => {
+                    Oaep::new_with_label::<Sha256, String>(String::from_utf8(buf.to_vec())?)
+                },
+                None => Oaep::new::<Sha256>(),
+            };
```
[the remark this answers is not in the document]
nabetti1720: Since "?" cannot be used in map, it looks like you will need to nest `map(...).unwrap_or_else(...)`.

```rust
let padding = label
    .as_ref()
    .map(|buf| {
        String::from_utf8(buf.to_vec())
            .map(|s| Oaep::new_with_label::<Sha256, String>(s))
            .unwrap_or_else(|_| Oaep::new::<Sha256>())
    })
    .unwrap_or_else(|| Oaep::new::<Sha256>());
```
Is it worth doing?

modules/llrt_crypto/Cargo.toml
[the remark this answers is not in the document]
nabetti1720: Hmm... I'm thinking of replacing some of the ring parts, but even though the names are the same, the methods are different, so it's not going to work.

It looks like I'll need to spend some time on this...

modules/llrt_crypto/src/subtle/decrypt.rs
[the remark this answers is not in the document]
richarddavison: I think we can even do this. We don't need UTF8 validations. In fact, we probably don't need UTF8 validations for any of the subtle crypto string conversions.
```rust
let padding = label.map_or(
    Oaep::new::<Sha256>(),
    |buf| Oaep::new_with_label::<Sha256>(String::from_utf8_lossy(buf).into_owned())
);
```

richarddavison (changes requested):

modules/llrt_crypto/Cargo.toml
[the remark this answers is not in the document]
nabetti1720: Take RSA for example.

```
// Cargo.toml
...
#rsa = { version = "0.9", features = ["std", "sha2"], default-features = false }
```

Then, if you replace `use rsa::...` with `use ring::rsa::...`, you will get an error saying that the import cannot be resolved.

image

If the command doesn't exist, do we need to look for an alias? If it still doesn't exist, do we need to implement it?

modules/llrt_crypto/src/subtle/mod.rs
```diff
@@ -0,0 +1,404 @@
...
+pub async fn subtle_decrypt<'js>(
+    ctx: Ctx<'js>,
+    algorithm: Value<'js>,
+    key: ObjectBytes<'js>,
+    data: ObjectBytes<'js>,
+) -> Result<ArrayBuffer<'js>> {
+    let algorithm = extract_algorithm_object(&ctx, &algorithm)?;
+
+    let bytes = decrypt(&ctx, &algorithm, key.as_bytes(), data.as_bytes())?;
+    ArrayBuffer::new(ctx, bytes.as_slice())
```
richarddavison, line 118: Do not pass slice as this will be an extra copy. Pass the vec directly to `ArrayBuffer`

modules/llrt_crypto/src/subtle/mod.rs
```diff
@@ -0,0 +1,404 @@
...
+pub async fn subtle_derive_bits<'js>(
+    ctx: Ctx<'js>,
+    algorithm: Value<'js>,
+    base_key: ObjectBytes<'js>,
+    length: u32,
+) -> Result<ArrayBuffer<'js>> {
+    let derive_algorithm = extract_derive_algorithm(&ctx, &algorithm)?;
+
+    let bytes = derive_bits(&ctx, &derive_algorithm, base_key.as_bytes(), length)?;
+    ArrayBuffer::new(ctx, bytes.as_slice())
```
richarddavison, line 130: Do not pass slice as this will be an extra copy. Pass the vec directly to `ArrayBuffer`

modules/llrt_crypto/src/subtle/mod.rs
```diff
@@ -0,0 +1,404 @@
...
+pub async fn subtle_encrypt<'js>(
+    ctx: Ctx<'js>,
+    algorithm: Value<'js>,
+    key: ObjectBytes<'js>,
+    data: ObjectBytes<'js>,
+) -> Result<ArrayBuffer<'js>> {
+    let algorithm = extract_algorithm_object(&ctx, &algorithm)?;
+
+    let bytes = encrypt(&ctx, &algorithm, key.as_bytes(), data.as_bytes())?;
+    ArrayBuffer::new(ctx, bytes.as_slice())
```
richarddavison, line 161: Do not pass slice as this will be an extra copy. Pass the vec directly to `ArrayBuffer`

modules/llrt_crypto/src/subtle/mod.rs
```diff
@@ -0,0 +1,404 @@
...
+pub async fn subtle_generate_key<'js>(
+    ctx: Ctx<'js>,
+    algorithm: Value<'js>,
+    _extractable: bool,
+    _key_usages: Array<'js>,
+) -> Result<ArrayBuffer<'js>> {
+    let key_gen_algorithm = extract_generate_key_algorithm(&ctx, &algorithm)?;
+
+    let bytes = generate_key(&ctx, &key_gen_algorithm)?;
+    ArrayBuffer::new(ctx, bytes.as_slice())
```
richarddavison, line 173: Do not pass slice as this will be an extra copy. Pass the vec directly to `ArrayBuffer`

modules/llrt_crypto/src/subtle/mod.rs
```diff
@@ -0,0 +1,404 @@
...
+pub async fn subtle_sign<'js>(
+    ctx: Ctx<'js>,
+    algorithm: Value<'js>,
+    key: ObjectBytes<'js>,
+    data: ObjectBytes<'js>,
+) -> Result<ArrayBuffer<'js>> {
+    let algorithm = extract_sign_verify_algorithm(&ctx, &algorithm)?;
+
+    let bytes = sign(&ctx, &algorithm, key.as_bytes(), data.as_bytes())?;
+    ArrayBuffer::new(ctx, bytes.as_slice())
```
richarddavison, line 185: Do not pass slice as this will be an extra copy. Pass the vec directly to `ArrayBuffer`

richarddavison (changes requested):

modules/llrt_crypto/Cargo.toml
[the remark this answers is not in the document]
richarddavison: For RSA I don't think ring can decrypt. So we probably need to keep it as well. Rings API is totally different then what you already have so you have to change it when switching to ring (where applicable)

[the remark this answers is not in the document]
nabetti1720: As for the other components, they are completely different from those built into the ring, so it would be difficult to replace them as is.

Even if it were possible, I would like to consider it in a separate PR.

nabetti1720:
> > * We will need CryptoKey / CryptoKeyPair classes
> 
> I know it's necessary, but I'm not sure how to implement it yet.
> 
> Nevertheless, since the same value is returned even if you run exportKey multiple times, this object must have hidden information (seed).
> 
> https://w3c.github.io/webcrypto/#cryptokey-interface-internal-slots

@richarddavison and @Sytten , Any good ideas on how to implement this? Is it possible to retain values ​​in LLRT that are not displayed in console.log?

Sytten:
I would create a class to wrap the vec at the very least so the user cant assume it is an array. Then we can start implementing the props the spec requires.

richarddavison:
> @richarddavison and @Sytten , Any good ideas on how to implement this? Is it possible to retain values ​​in LLRT that are not displayed in console.log?

Yes, create a Class for CryptoKey and CryptoKeyPair and keep hidden fields in rust only. Then expose only getters to what you need to provide as read only:

```rust
#[rquickjs::class]
#[derive(rquickjs::JsLifetime)]
struct CryptoKey<'js>{
   algorithm: Object<'js>,
   usages: Array<'js>,
   ...
}

impl<'js> CryptoKey<'js>{

    #[qjs(get)]
    pub fn algorithm(&self) -> Object<'js> {
        self.algorithm.clone()
    }
}
```

modules/llrt_crypto/src/subtle/decrypt.rs
```diff
@@ -0,0 +1,57 @@
...
+                Oaep::new_with_label::<Sha256, _>(&String::from_utf8_lossy(buf))
+            });
+
+            private_key.decrypt(padding, data).or_throw(ctx)
+        },
+        _ => Err(Exception::throw_message(ctx, "Algorithm not supported")),
+    }
+}
+
+fn decrypt_aes_ctr_gen<B>(ctx: &Ctx<'_>, key: &[u8], counter: &[u8], data: &[u8]) -> Result<Vec<u8>>
```
richarddavison, line 47: nit: maybe rename generic type to T rather than B?

modules/llrt_crypto/src/subtle/encrypt.rs
```diff
@@ -0,0 +1,59 @@
...
+            });
+            let mut rng = OsRng;
+
+            Ok(public_key.encrypt(&mut rng, padding, data).or_throw(ctx)?)
+        },
+        _ => Err(Exception::throw_message(ctx, "Algorithm not supported")),
+    }
+}
+
+fn encrypt_aes_ctr_gen<B>(ctx: &Ctx<'_>, key: &[u8], counter: &[u8], data: &[u8]) -> Result<Vec<u8>>
```
richarddavison, line 49: nit: change <B> to <T>

modules/llrt_crypto/src/subtle/mod.rs
```diff
@@ -0,0 +1,404 @@
...
+                .get_optional("iterations")?
+                .ok_or_else(|| Exception::throw_message(ctx, "PBKDF2 must have iterations"))?;
+
+            Ok(DeriveAlgorithm::Pbkdf2 {
+                hash,
+                salt,
+                iterations,
+            })
+        },
+        _ => Err(Exception::throw_message(ctx, "Algorithm not supported")),
```
richarddavison, line 348: Maybe give name of the algorithm they passed?

modules/llrt_crypto/Cargo.toml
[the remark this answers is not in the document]
nabetti1720: If you don't use the p256/p384 crates I don't understand how the implementation below replaces it. Are these features implemented in ring?

modules/llrt_crypto/src/subtle/derive_bits.rs:
- p256::SecretKey::from_pkcs8_der()
- p256::SecretKey::from_pkcs8_der(public)
- p256::elliptic_curve::ecdh::diffie_hellman()
- p384::SecretKey::from_pkcs8_der()
- p384::SecretKey::from_pkcs8_der()
- p384::elliptic_curve::ecdh::diffie_hellman()

modules/llrt_crypto/src/subtle/mod.rs
```diff
@@ -34,13 +34,35 @@ pub enum Sha {
...
+impl TryFrom<&str> for Sha {
+    type Error = &'static str;
+
+    fn try_from(hash: &str) -> std::result::Result<Self, Self::Error> {
+        match hash.to_ascii_uppercase().as_str() {
+            "SHA-1" => Ok(Sha::Sha1),
+            "SHA-256" => Ok(Sha::Sha256),
+            "SHA-384" => Ok(Sha::Sha384),
+            "SHA-512" => Ok(Sha::Sha512),
+            _ => Err("hash not found"),
```
richarddavison, line 46: Which hash? Can be good to inform user what hash they try to use
Err(["\""_,"\" not available"].concat()

modules/llrt_crypto/src/subtle/mod.rs
```diff
@@ -34,13 +34,35 @@ pub enum Sha {
...
+}
+
+impl TryFrom<&str> for CryptoNamedCurve {
+    type Error = &'static str;
+
+    fn try_from(curve: &str) -> std::result::Result<Self, Self::Error> {
+        match curve.to_ascii_uppercase().as_str() {
+            "P256" => Ok(CryptoNamedCurve::P256),
+            "P384" => Ok(CryptoNamedCurve::P384),
+            _ => Err("named_curve not found"),
```
richarddavison, line 64: Same as above

richarddavison (changes requested):

modules/llrt_crypto/src/subtle/generate_key.rs
```diff
@@ -24,22 +24,49 @@ pub async fn subtle_generate_key<'js>(
...
+            "public".to_string(),
+            true,
+            algorithm,
+            key_usages, // for test
+            bytes,      // for test
+        )?;
+        let obj = Object::new(ctx)?;
+        obj.set("privateKey", private)?;
+        obj.set("publicKey", public)?;
+        Ok(obj.into())
```
richarddavison, line 62: Shouldn't this be a class? CryptoKeyPair? Or is it fine with an object?

richarddavison (changes requested):
Fantastic work so far! 🥇

modules/llrt_crypto/src/subtle/generate_key.rs
richarddavison: Yes, class with getters. We should not have writable or configurable objects.

richarddavison:
This is great. Now we can probably add the [CryptoTest](https://github.com/web-platform-tests/wpt/tree/master/WebCryptoAPI) from here https://github.com/web-platform-tests/wpt/tree/master

Sytten (changes requested):

modules/llrt_crypto/src/subtle/crypto_key.rs
```diff
@@ -0,0 +1,86 @@
...
+    pub fn usages(&self) -> Array<'js> {
+        self.usages.clone()
+    }
+}
+
+#[rquickjs::class]
+#[derive(Clone, Trace, rquickjs::JsLifetime)]
+pub struct CryptoKeyPair<'js> {
+    private_key: CryptoKey<'js>,
+    public_key: CryptoKey<'js>,
```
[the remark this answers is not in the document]
richarddavison: They will be refcounted since they implement trace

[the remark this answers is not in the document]
richarddavison: We should Store them as instances so clone is a ref count increase:
```rust
pub struct CryptoKeyPair<'js> {
    private_key: Class::<CryptoKey<'js>>,
    public_key: Class::<CryptoKey<'js>>,
}
```

modules/llrt_crypto/src/subtle/crypto_key.rs
```diff
@@ -0,0 +1,86 @@
...
+}
+
+#[rquickjs::methods(rename_all = "camelCase")]
+impl<'js> CryptoKeyPair<'js> {
+    #[qjs(constructor)]
```
richarddavison, line 65: This function is internal only and should not be exposed to JS. Move to a separate impl block that isn't using `#[rquickjs::methods]`

richarddavison (changes requested):

modules/llrt_crypto/Cargo.toml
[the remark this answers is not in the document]
richarddavison: Does this example help?
**Note that this example uses X25519, but ECDH using NIST P-256/P-384 is done exactly the same way, just substituting agreement::ECDH_P256/agreement::ECDH_P384 for agreement::X25519.**
https://docs.rs/ring/latest/ring/agreement/index.html#example

nabetti1720:
> This is great. Now we can probably add the [CryptoTest](https://github.com/web-platform-tests/wpt/tree/master/WebCryptoAPI) from here https://github.com/web-platform-tests/wpt/tree/master

Thank you. :) I'm always checking the test cases of wpt and looking for any unclear points in the implementation from the test cases. I'd like to incorporate it into the LLRT tests at some point.

nabetti1720:
What's going on? I'm suddenly getting errors with Clippy, but they don't happen on my laptop...

image

richarddavison:
> What's going on? I'm suddenly getting errors with Clippy, but they don't happen on my laptop...
> 
> image

Probably ci is using a newer version of clippy. Try upgrading.

modules/llrt_crypto/src/subtle/generate_key.rs
```diff
@@ -19,6 +23,94 @@ use super::crypto_key::{CryptoKey, CryptoKeyPair};
...
+        HashSet::from(["sign".to_string(), "verify".to_string()]),
+    );
+    map.insert(
+        "ECDH",
+        HashSet::from(["deriveKey".to_string(), "deriveBits".to_string()]),
+    );
+
+    map
+});
+
```
richarddavison, line 113: This can be heavily optimized by using static arrays instead of Lazy HashMaps. It's almost always better to use an array when dealing with few items especially if the keys are strings with short length. Also when having the same value twice we can directly work with references:

```rust
static SYMMETRIC_USAGES: &[&str] = &["encrypt", "decrypt", "wrapKey", "unwrapKey"];
static SIGNATURE_USAGES: &[&str] = &["sign", "verify"];
static EMPTY_USAGES: &[&str] = &[];
static SIGN_USAGES: &[&str] = &["sign"];
static RSA_OAEP_USAGES: &[&str] = &["decrypt", "unwrapKey"];
static ECDH_USAGES: &[&str] = &["deriveKey", "deriveBits"];

static SUPPORTED_USAGES_ARRAY: &[(&str, &[&str])] = &[
    ("AES-CTR", SYMMETRIC_USAGES),
    ("AES-CBC", SYMMETRIC_USAGES),
    ("AES-GCM", SYMMETRIC_USAGES),
    ("RSA-OAEP", SYMMETRIC_USAGES),
    ("HMAC", SIGNATURE_USAGES),
    ("RSASSA-PKCS1-v1_5", SIGNATURE_USAGES),
    ("RSA-PSS", SIGNATURE_USAGES),
    ("ECDSA", SIGNATURE_USAGES),
    ("ECDH", &["deriveKey", "deriveBits"]),
];

static MANDATORY_USAGES_ARRAY: &[(&str, &[&str])] = &[
    ("AES-CTR", EMPTY_USAGES),
    ("AES-CBC", EMPTY_USAGES),
    ("AES-GCM", EMPTY_USAGES),
    ("HMAC", EMPTY_USAGES),
    ("RSASSA-PKCS1-v1_5", SIGN_USAGES),
    ("RSA-PSS", SIGN_USAGES),
    ("ECDSA", SIGN_USAGES),
    ("RSA-OAEP", RSA_OAEP_USAGES),
    ("ECDH", ECDH_USAGES),
];
```

We can also optimize the usages check by just checking if we're using a non supported value or missing a mandatory, for example:
```rust
fn process_usages(
    mandatory_usages: &[&str],
    supported_usages: &[&str],
    key_usages_set: HashSet<String>,
) -> Result<HashSet<String>> {
    //early check for mandatory usages
    if key_usages_set.len() != mandatory_usages.len() {
        return Err(...);
    }

    //check all usages are supported first
    if !key_usages_set
        .iter()
        .all(|u| supported_usages.contains(u.as_str()))
    {
         return Err(...);
    }

    //check mandatory usages are present
    if !mandatory_usages.iter().all(|m| key_usages_set.contains(*m)) {
        return Err(...);
    }

    //all checks passed
    Ok(key_usages_set)
}

```
There is no clones, filters or duplication of data here :)

richarddavison (changes requested):

modules/llrt_crypto/src/subtle/generate_key.rs
nabetti1720: This seems like it could be significantly optimized! However, it would require separating public_usages and private_usages.

richarddavison: This was just psudo code, maybe you can apply the same concept for both at the same time? Are `public_or_secret_usages` not checked for supported but only filtered by mandatory? What's the difference between `public_or_secret_usages` and `private_usages`?

nabetti1720: I could not find any [specification](https://w3c.github.io/webcrypto/#keypair) that separates `usages` for privateKey and publicKey when generateKey returns CryptoKeyPair.

After checking the behavior of bun, I found that it follows the following rules, so I am using them as reference.

1. If `MANDATORY_USAGES` is defined, at least one of them must be specified (an error occurs if none are specified)
2. Only those defined in `SUPPORTED_USAGES` must be specified (an error occurs if anything other than `SUPPORTED_USAGES` is specified)
3. Only privateKey `usages` included in `MANDATORY_USAGES` are passed.
4. Only publicKey `usages` not included in `MANDATORY_USAGES` are passed.

```javascript
// subtleGenerateKey.js
const key = await crypto.subtle
  .generateKey(
    {
      name: "ECDSA",
      namedCurve: "P-256",
    },
    true,
    ["sign", "verify"]
  );

console.log(key);
```
```
% bun subtleGenerateKey.js 
{
  privateKey: CryptoKey {
    type: "private",
    extractable: true,
    algorithm: {
      name: "ECDSA",
      namedCurve: "P-256",
    },
    usages: [ "sign" ],
  },
  publicKey: CryptoKey {
    type: "public",
    extractable: true,
    algorithm: {
      name: "ECDSA",
      namedCurve: "P-256",
    },
    usages: [ "verify" ],
  },
}
```

modules/llrt_crypto/src/subtle/sign.rs
```diff
@@ -0,0 +1,97 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+use hmac::Mac;
+use llrt_utils::{bytes::ObjectBytes, result::ResultExt};
+use rand::rngs::OsRng;
+use ring::signature::EcdsaKeyPair;
```
richarddavison, line 6: nice

modules/llrt_crypto/src/subtle/generate_key.rs
```diff
@@ -0,0 +1,294 @@
...
+        )),
+    }
+}
+
+fn classify_and_check_usages<'js>(
+    ctx: &Ctx<'js>,
+    name: &str,
+    key_usages: &Array<'js>,
+) -> Result<(Array<'js>, Array<'js>)> {
+    fn get_usages(arr: &[(&str, &[&str])], algorithm: &str) -> Option<HashSet<String>> {
```
richarddavison, line 159: We can simplify here by not returning hashset but instead working with arrays. Remember that HashSet is great for more than 20-50+ items. Otherwise it's slower. We also can return the referenced array, no need to create a copy.
```rust
fn find_usages<'a>(table: &'a [(&str, &[&str])], algorithm: &str) -> Option<&'a [&'a str]> {
    table
        .iter()
        .find(|(name, _)| *name == algorithm)
        .map(|(_, usages)| *usages)
}
```

Then we can instead use a single for loop, something like:
```rust
let supported_usages = find_usages(SUPPORTED_USAGES_ARRAY)
let mandatory_usages = find_usages(MANDATORY_USAGES_ARRAY)
let mut private_usages: Vec<String> = Vec::with_capacity(key_usages_set.len());
let mut public_usages: Vec<String> = Vec::with_capacity(key_usages_set.len());

for usage in key_usages_set {
    if !supported_usages.contains(&usage.as_str()) {
        return Err(...);
    }
    if mandatory_usages.contains(&usage.as_str()) {
        private_usages.push(usage.into());
    }else{
        public_usages.push(usage.into());
    }
}
if !supported_usages.is_empty() && private_usages.is_empty(){
  return Err(...) //we have no mandatory usages
}
```

richarddavison (changes requested):

nabetti1720:
Although there are still some issues with parameter processing, samples including generateKey/sign/verify now work.

```javascript
// subtleVerify.js
async function generateKeyPair() {
  return await crypto.subtle.generateKey(
    {
      name: "RSASSA-PKCS1-v1_5",
      modulusLength: 2048,
      //  publicExponent: new Uint8Array([0x01, 0x00, 0x01]), // 65537
      publicExponent: [0x01, 0x00, 0x01], // 65537
      hash: "SHA-256",
    },
    true,
    ["sign", "verify"]
  );
}

async function signData(privateKey, data) {
  const encoder = new TextEncoder();
  const encodedData = encoder.encode(data);
  return await crypto.subtle.sign(
    {
      name: "RSASSA-PKCS1-v1_5",
    },
    privateKey,
    encodedData
  );
}

async function verifySignature(publicKey, signature, data) {
  const encoder = new TextEncoder();
  const encodedData = encoder.encode(data);
  const isValid = await crypto.subtle.verify(
    {
      name: "RSASSA-PKCS1-v1_5",
    },
    publicKey,
    signature,
    encodedData
  );
  return isValid;
}

async function testCryptoVerify() {
  const { publicKey, privateKey } = await generateKeyPair();
  const testData = "This is a test message.";
  const signature = await signData(privateKey, testData);
  const isValid = await verifySignature(publicKey, signature, testData);
  console.log("Signature is valid:", isValid);  // `true`

  // Invalid signature
  const fakeSignature = new Uint8Array(signature.length).fill(0);
  const isFakeValid = await verifySignature(publicKey, fakeSignature, testData);
  console.log("Fake signature is valid:", isFakeValid);  // `false`
}

testCryptoVerify().catch(console.error);
```
```
 % llrt subtleVerify.js
Signature is valid: true
Fake signature is valid: false

 % bun subtleVerify.js
Signature is valid: true
Fake signature is valid: false
```

modules/llrt_crypto/Cargo.toml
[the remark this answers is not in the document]
nabetti1720: `hmac`, `sha1`, and `sha2` crates have been removed.

modules/llrt_crypto/src/subtle/generate_key.rs
```diff
@@ -0,0 +1,276 @@
...
+        let public_key = CryptoKey::new(
+            ctx.clone(),
+            "public".to_string(),
+            true,
+            algorithm,
+            public_usages.into_js(&ctx)?.into_array().unwrap(),
+            &bytes,
+        )?;
+        let public_key = Class::instance(ctx.clone(), public_key)?;
+        CryptoKeyPair::new(ctx.clone(), private_key, public_key).into_js(&ctx)
```
richarddavison, line 97: Looking how chrome does it, we can actually remove this class and just return an object with two properties. This should not be a class. I know it was an object before but I missed this completely.
Not that it should have readonly properties and not writable properties.
```rust
let mut key_pair = Object::new(ctx)?;
key_pair.prop("privateKey", public_key)?;
key_pair.prop("publicKey", private_key)?;
Ok(key_pair);
```

modules/llrt_crypto/src/subtle/crypto_key.rs
```diff
@@ -0,0 +1,94 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+use std::rc::Rc;
+
+use rquickjs::{class::Trace, Array, Class, Ctx, JsLifetime, Result, Value};
+
+#[rquickjs::class]
+#[derive(Clone, Trace, rquickjs::JsLifetime)]
+pub struct CryptoKey<'js> {
```
richarddavison, line 9: We don't want to store the JS values in here but the rust structs directly. This is because storing the JS values allows access by reference and make them modifiable. Instead we should convert them to JS values on access.
For instance, in the algorithm getter we should convert KeyGenAlgorithm into JS but only on access. This also offers better performance as we can remove converting to and from rust as the key is likely passed to another function. Same methodology goes for all other CryptoAPI structures that should return non-modifiable data:
See https://developer.mozilla.org/en-US/docs/Web/API/Web_Crypto_API

richarddavison (changes requested):
We're getting closer 🎉

modules/llrt_crypto/src/subtle/generate_key.rs
nabetti1720: There doesn't seem to be any problem functionally, but the display of console.log is slightly inconvenient. When logging CryptoKeyPair in bun, the internal contents are displayed, but in LLRT, they are not displayed.

Is it because the specifications of the console module are different?

```javascript
const key = await crypto.subtle
  .generateKey(
    {
      name: "ECDSA",
      namedCurve: "P-256",
    },
    true,
    ["sign", "verify"]
  );

console.log("console.log(key):");
console.log(key);
console.log("console.log(key.publicKey):");
console.log(key.publicKey);
```

```
% llrt subtleGenerateKey.js
console.log(key):
{}
console.log(key.publicKey):
{
  algorithm: {
    name: 'ECDSA',
    namedCurve: 'P-256'
  },
  type: 'public',
  usages: [ 'verify' ],
  extractable: true
}

% bun subtleGenerateKey.js 
console.log(key):
{
  privateKey: CryptoKey {
    type: "private",
    extractable: true,
    algorithm: {
      name: "ECDSA",
      namedCurve: "P-256",
    },
    usages: [ "sign" ],
  },
  publicKey: CryptoKey {
    type: "public",
    extractable: true,
    algorithm: {
      name: "ECDSA",
      namedCurve: "P-256",
    },
    usages: [ "verify" ],
  },
}
console.log(key.publicKey):
CryptoKey {
  type: "public",
  extractable: true,
  algorithm: {
    name: "ECDSA",
    namedCurve: "P-256",
  },
  usages: [ "verify" ],
}
```

modules/llrt_crypto/src/subtle/mod.rs
```diff
@@ -208,7 +210,20 @@ fn extract_sha_hash(ctx: &Ctx<'_>, algorithm: &Value) -> Result<Sha> {
 }
 
 fn check_supported_usage(ctx: &Ctx<'_>, key_usages: &Array, name: &str) -> Result<()> {
-    if !key_usages.contains_key(name)? {
+    let mut is_supported: bool = false;
+
+    for value in key_usages.clone().into_iter() {
```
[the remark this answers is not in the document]
nabetti1720: It seems that the comment is about a slightly old implementation. The current implementation is as follows:

```rust
fn check_supported_usage(ctx: &Ctx<'_>, key_usages: &Array, usage: &str) -> Result<()> {
    for value in key_usages.clone().into_iter() {
        if let Some(key) = value?.as_string() {
            let key = key.to_string()?;
            if key.as_str() == usage {
                return Ok(());
            }
        }
    }
    Err(Exception::throw_type(
        ctx,
        &["CryptoKey doesn't support '", usage, "'"].concat(),
    ))
}
```

In reality, I wanted to make a one-time decision using a command like `if !key_usages.contains(usage)? {}`(not `Array.contains_key(usage)`)...

modules/llrt_crypto/src/subtle/crypto_key.rs
nabetti1720: Does this mean an implementation like the [CryptoKey internal slots](https://w3c.github.io/webcrypto/#cryptokey-interface-internal-slots) from the Web Cryptography API?

Currently, only the handle is implemented internally in this way, but the idea is to implement it so that other CryptoKey variables are also managed internally.

modules/llrt_crypto/src/subtle/generate_key.rs
nabetti1720: > I would keep the class. There is a readonly flag on the derive of class. It basically is the same as a js object but we can extend it.

@Sytten , Are you saying that it can fulfill both what you're saying and what @richarddavison  is saying?

richarddavison: For console log, you have to make the prop enumerable. Search for this and see example of how to in the project 👍

Sytten:
My point about the class CryptoKeyPair is you can make a rust class that is both readonly and the properties enumerable. IMO cleaner 

```rust
#[rquickjs::class(frozen)] 
#[derive(Clone, Trace, rquickjs::JsLifetime)] 
pub struct CryptoKeyPair<'js> {
   #[qjs(get, enumerable)]
   private_key: Class<'js, CryptoKey<'js>>,
   #[qjs(get, enumerable)]
   public_key: Class<'js, CryptoKey<'js>>, 
}
```

richarddavison:
> My point about the class CryptoKeyPair is you can make a rust class that is both readonly and the properties enumerable. IMO cleaner
> 
> ```rust
> #[rquickjs::class(frozen)] 
> #[derive(Clone, Trace, rquickjs::JsLifetime)] 
> pub struct CryptoKeyPair<'js> {
>    #[qjs(get, enumerable)]
>    private_key: Class<'js, CryptoKey<'js>>,
>    #[qjs(get, enumerable)]
>    public_key: Class<'js, CryptoKey<'js>>, 
> }
> ```

Yes but the spec uses an object

modules/llrt_crypto/src/subtle/crypto_key.rs
richarddavison: Since the algorithm is an object we pass from user space (JavaScript) we have two options. One is to convert that to an internal rust representation and store it (I think there is a proc macro for this as well) . The other is to deep clone it when access it (not preferable). The reason for this is that we should not be able to change values inside of this object after it has been passed to generate key. Remember that .clone() on a Value or Object in rquickjs returns a reference to the same object.

nabetti1720:
> Yes but the spec uses an object

Yes, the current implementation is as follows:

```rust
let key_pair = Object::new(ctx.clone())?;
key_pair.prop("privateKey", Property::from(private_key).enumerable())?;
key_pair.prop("publicKey", Property::from(public_key).enumerable())?;
key_pair.into_js(&ctx)
```

modules/llrt_crypto/src/subtle/crypto_key.rs
richarddavison: @nabetti1720 please let me know if you wan't support for this specific task. I can check out your branch and push the required changes for this change request 🥇

nabetti1720: @richarddavison, Thank you for your support!

I thought this part would be a little difficult to fix. If you have the time, could you please fix this branch? I'd love to learn from your fixes.

modules/llrt_crypto/Cargo.toml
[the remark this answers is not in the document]
nabetti1720: It is likely that the implementation will be like this, but it is difficult because it is not possible to create a `privateKey` from `base_key`.

```rust
fn derive_bits(
    ctx: &Ctx<'_>,
    algorithm: &DeriveAlgorithm,
    base_key: &[u8],
    length: u32,
) -> Result<Vec<u8>> {
    match algorithm {
        DeriveAlgorithm::Edch { curve, public } => {
            let curve = match curve {
                EllipticCurve::P256 => &ring::agreement::ECDH_P256,
                EllipticCurve::P384 => &ring::agreement::ECDH_P384,
            };
            // Originally, privateKey should be created using base_key seed.
            let private_key =
                EphemeralPrivateKey::generate(curve, &SYSTEM_RANDOM.to_owned()).or_throw(ctx)?;
            let peer_public_key = UnparsedPublicKey::new(curve, &public[..]);

            let mut shared_secret = vec![];
            agree_ephemeral(private_key, &peer_public_key, |secret| {
                // TODO:
                // shared_secret.extend_from_slice(secret);
                // let _ = Ok::<(), Error>(());
            })
            .or_throw(ctx)?;

            Ok(shared_secret)
        },
        ...
```

richarddavison:
@nabetti1720 thanks for this fantastic PR! Hold off on more implementations according to spec as I'm doing some major refactoring. We can add more implementations in other PRs as this is already quite big :)

nabetti1720:
> @nabetti1720 thanks for this fantastic PR! Hold off on more implementations according to spec as I'm doing some major refactoring. We can add more implementations in other PRs as this is already quite big :)

@richarddavison , thank you for your attention to this PR. I agree. I'd like to think about implementing the remaining features after your refactoring is complete. I think not only me but everyone else is looking forward to seeing you complete this work. :)

richarddavison:
Hi @nabetti1720. Id try keeping changes to a minimum and go little by little but I don't want to commit things in a broken state. Since what I'm doing now is refactoring the fundamental enums and more applying SOLID principles there will be a lot of modifications. In summery I'm moving things to shared enums, moving methods to enums, implementing FromJs IntoJs traits where I can and reorganizing by applying stronger separation of concern. Also there was a lot of duplication which is now removed. What's left is derive and then I'll push

richarddavison:
Ok, I pushed what I have right now. This was a major rewrite, I'll try to summarize here:

1. Divided signing and encryption params into separate enums convertible from JS
2. Remove hardcoded SHA hashes, these are now derived from the key
3. Moved a lot of logic to implement on the enums instead of in functions
4. Reduced  code duplication. Some enums have shared keys (KeyDerivation) etc
5. Added more guardrails according to spec
6. Verified that keygen + action algorithms matched.
7. ...a lot more

What's left is fixing broken tests. I'm not a 100% sure we got the verification of usages right, for import they seem to break right now. There is also a big performance hit using RSA but that will be fixed once that crate switches some dependencies.

richarddavison:
> Great refactoring! Let me know if there's anything else I should do next. I have already confirmed on my laptop that the following issues occur after resolving the usage issue:
> 
> ```
> SubtleCrypto deriveBits/deriveKey > should be processing ECDH algorithm
> Error: PKCS#8 ASN.1 error: unexpected ASN.1 DER tag: expected SEQUENCE, got OCTET STRING
>     at <anonymous> (/Users/shinya/Workspaces/llrt/bundle/js/__tests__/unit/crypto.subtle.test.js:874:53)
> 
> SubtleCrypto deriveBits/deriveKey > should be processing HKDF algorithm
> Error: PKCS#8 ASN.1 error: incorrect length for OCTET STRING
>     at <anonymous> (/Users/shinya/Workspaces/llrt/bundle/js/__tests__/unit/crypto.subtle.test.js:1051:55)
> 
> SubtleCrypto deriveBits/deriveKey > should be processing PBKDF2 algorithm
> Error: PKCS#8 ASN.1 error: unexpected ASN.1 DER tag: expected SEQUENCE, got OCTET STRING
>     at <anonymous> (/Users/shinya/Workspaces/llrt/bundle/js/__tests__/unit/crypto.subtle.test.js:1236:55)
> ```
> 
> Indeed, this seems to be an error that occurs when using something that has been exported/imported.
> 
> EDIT: This is because while the conversion process when the exportKey is in raw mode is implemented, the reverse conversion process for the importKey in raw mode is not implemented.

Thanks! I'll fix usage checks according to this table:

```
┌─────────┬─────────────────────┬─────────────────────────────────────┬─────────────────────────────────────┬─────────────────────────────────────┬─────────────────────────────────────┐
│ (index) │ Algorithm           │ Generate Key                        │ Derive Key                          │ Import Key                          │ Required Usages                     │
├─────────┼─────────────────────┼─────────────────────────────────────┼─────────────────────────────────────┼─────────────────────────────────────┼─────────────────────────────────────┤
│ 0       │ 'AES-GCM'           │ 'encrypt,decrypt,wrapKey,unwrapKey' │ '-'                                 │ 'encrypt,decrypt,wrapKey,unwrapKey' │ 'encrypt,decrypt,wrapKey,unwrapKey' │
│ 1       │ 'AES-CBC'           │ 'encrypt,decrypt,wrapKey,unwrapKey' │ '-'                                 │ 'encrypt,decrypt,wrapKey,unwrapKey' │ 'encrypt,decrypt,wrapKey,unwrapKey' │
│ 2       │ 'AES-CTR'           │ 'encrypt,decrypt,wrapKey,unwrapKey' │ '-'                                 │ 'encrypt,decrypt,wrapKey,unwrapKey' │ 'encrypt,decrypt,wrapKey,unwrapKey' │
│ 3       │ 'AES-KW'            │ 'wrapKey,unwrapKey'                 │ '-'                                 │ 'wrapKey,unwrapKey'                 │ 'wrapKey,unwrapKey'                 │
│ 4       │ 'HMAC'              │ 'sign,verify'                       │ '-'                                 │ 'sign,verify'                       │ 'sign,verify'                       │
│ 5       │ 'Ed25519'           │ 'sign,verify'                       │ '-'                                 │ '-'                                 │ 'sign'                              │
│ 6       │ 'RSA-OAEP'          │ 'encrypt,decrypt,wrapKey,unwrapKey' │ '-'                                 │ '-'                                 │ 'decrypt,unwrapKey'                 │
│ 7       │ 'RSA-PSS'           │ 'sign,verify'                       │ '-'                                 │ '-'                                 │ 'sign'                              │
│ 8       │ 'RSASSA-PKCS1-v1_5' │ 'sign,verify'                       │ '-'                                 │ '-'                                 │ 'sign'                              │
│ 9       │ 'ECDSA'             │ 'sign,verify'                       │ '-'                                 │ '-'                                 │ 'sign'                              │
│ 10      │ 'ECDH'              │ 'deriveKey,deriveBits'              │ 'encrypt,decrypt,wrapKey,unwrapKey' │ '-'                                 │ 'deriveKey,deriveBits'              │
│ 11      │ 'HKDF'              │ '-'                                 │ 'encrypt,decrypt,wrapKey,unwrapKey' │ '-'                                 │ 'encrypt,decrypt,wrapKey,unwrapKey' │
│ 12      │ 'PBKDF2'            │ '-'                                 │ 'encrypt,decrypt,wrapKey,unwrapKey' │ '-'                                 │ 'encrypt,decrypt,wrapKey,unwrapKey' │
│ 13      │ 'X25519'            │ '-'                                 │ 'encrypt,decrypt,wrapKey,unwrapKey' │ '-'                                 │ 'encrypt,decrypt,wrapKey,unwrapKey' │
└─────────┴─────────────────────┴─────────────────────────────────────┴─────────────────────────────────────┴─────────────────────────────────────┴─────────────────────────────────────┘

```

richarddavison:
@nabetti1720 we have two options to proceed here, either fix imports or disable imports completely, merge and open a new PR with this capability. This PR is already enormous 👍

nabetti1720:
> we have two options to proceed here, either fix imports or disable imports completely, merge and open a new PR with this capability. This PR is already enormous

@richarddavison , Thank you! As you say, this PR is already too big. I think it would be better to disable importKey first and merge it. After that, it would be better to open an issue that you currently know about and work towards resolving it.

I'm sorry to trouble you, but could you please merge this PR?

nabetti1720:
> @nabetti1720 please go ahead! We can just disable the export from the module and add a `it.skip(...)` or `describe.skip(...)` for the tests

@richarddavison , Instead of skipping all the tests, I kept what I could. All the tests should pass. We can finally see the goal. :)

richarddavison (approved):

modules/llrt_crypto/Cargo.toml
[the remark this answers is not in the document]
richarddavison: Let's revisit this in another PR :)

panva:
@nabetti1720 may i ask which tool presents the compatibility tables in the PR's description?

panva:
> > may i ask which tool presents the compatibility tables in the PR's description?
> 
> @richarddavison Runtime compatibility - [runtime-compat.unjs.io](https://runtime-compat.unjs.io/)
> 
> The source for this site is maintained on GitHub at: [unjs/runtime-compat](https://github.com/unjs/runtime-compat?rgh-link-date=2024-12-14T13%3A00%3A19Z)

Ah, so it's not based on WPTs. Thanks

PR #711, opened 2024-12-01
Store wakers in Vec instead of busy waiting

richarddavison:
### Description of changes

Store waiters in Vec instead of busy waiting

### Checklist

### DIFF
--- a/libs/llrt_utils/Cargo.toml
+++ b/libs/llrt_utils/Cargo.toml
@@ -18,7 +18,7 @@ rquickjs = { git = "https://github.com/DelSkayn/rquickjs.git", version = "0.8.1"
   "array-buffer",
   "macro",
 ], default-features = false }
-tokio = { version = "1" }
+tokio = { version = "1", features = ["sync"] }
 tracing = "0.1"
 
 [dev-dependencies]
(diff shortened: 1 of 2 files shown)


Sytten:
For reference we should look at the watch.
Couple of improvements they made:
- They have a shared struct with one arc, bit faster to create senders/receivers
- It is built on top of the Notify primitive (https://docs.rs/tokio/latest/tokio/sync/struct.Notify.html)
- They use an array of 8 Notify to avoid contention (BigNotify)

See https://docs.rs/tokio/latest/src/tokio/sync/watch.rs.html#1-1477

richarddavison:
> For reference we should look at the watch. Couple of improvements they made:
> 
> * They have a shared struct with one arc, bit faster to create senders/receivers

Ah ofc. Didn't think of that!

> * It is built on top of the Notify primitive (https://docs.rs/tokio/latest/tokio/sync/struct.Notify.html)
> * They use an array of 8 Notify to avoid contention (BigNotify)

That could work but is an internal class, so can't use it

> * For a write heavy load like waker registration a mutex is more performant than a RwLock

Sure, but this isn't very write heavy, right?. The executor won't poll this future again until the waker is called so registration only happens once per copy of mc_oneshot no?

Sytten:
For reference we should look at the watch.
Couple of improvements they made:
- They have a shared struct with one arc, bit faster to create senders/receivers
- It is built on top of the Notify primitive (https://docs.rs/tokio/latest/tokio/sync/struct.Notify.html)
- They use an array of 8 Notify to avoid contention (BigNotify)

See https://docs.rs/tokio/latest/src/tokio/sync/watch.rs.html#1-1477

Sytten:
We can use Notify though if we want.

Mutex is better for mixed workload.

Issue #712, opened 2024-12-01
`.toString()` of 0 returns empty string for some radix values


ainsleyrutterford:
In Node, `(0).toString(16)` returns `"0"`, whereas in LLRT it returns `""`.

Depending on the radix value used, the behaviour differs slightly, LLRT sometimes matches the Node implementation:

```js
(0).toString(2);  // Node: "0", LLRT: "0"
(0).toString(10); // Node: "0", LLRT: "0"
(0).toString(16); // Node: "0", LLRT: ""
(0).toString(13); // Node: "0", LLRT: ""
```

This causes issues when using dependencies like https://github.com/kjur/jsrsasign (which I'm just using until https://github.com/awslabs/llrt/issues/184 is implemented which will hopefully allow us to use https://github.com/auth0/node-jsonwebtoken).

ainsleyrutterford:
Looks like it might be because the `while` loop is never entered here:

https://github.com/awslabs/llrt/blob/acf47edbcfe1ec2b6e63f69fef9072f04840d857/libs/llrt_numbers/src/lib.rs#L78

Maybe the `0` case could be handled explicitly before the `while` loop.

PR #713, opened 2024-12-01
Fix `toString()` of 0 for all radix values

ainsleyrutterford:
### Issue # (if available)

https://github.com/awslabs/llrt/issues/712

### Description of changes

In Node, `(0).toString(16)` returns `"0"`, whereas in LLRT it returns `""`. This PR updates the `internal_i64_to_base_n` function so that the LLRT implementation is consistent with the Node.

### Checklist

### DIFF
--- a/libs/llrt_numbers/src/lib.rs
+++ b/libs/llrt_numbers/src/lib.rs
@@ -76,6 +76,11 @@ pub fn i64_to_base_n(number: i64, radix: u8) -> String {
 
 #[inline(always)]
 fn internal_i64_to_base_n(buf: &mut [u8], number: i64, radix: u8) -> usize {
+    if number == 0 {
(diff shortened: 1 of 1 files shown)


libs/llrt_numbers/src/lib.rs
```diff
@@ -295,6 +300,9 @@ mod test {
         }
```
ainsleyrutterford, line 300: Interestingly, I think the test on the line above would have failed if `0` was ever chosen, but I guess it hasn't been one of the million random numbers yet.

richarddavison: Maybe why we shouldn't always just test with random data :)

richarddavison (approved):

richarddavison (commented):
Thank you, nice finding!

PR #726, opened 2024-12-07
Calculate content-length if not set

floydspace:
### Issue # (if available)

The API I'm using to call with fetch requires `content-length` header to be set, it took me some hard time to figure that out by debugging llrt internals, so I decided to contribute with finding

### Description of changes

Calculate and set `content-length` header for the fetch request

### Checklist

  - _(please guide me how to cover this with tests)_

### DIFF
--- a/modules/llrt_http/src/fetch.rs
+++ b/modules/llrt_http/src/fetch.rs
@@ -6,6 +6,7 @@ use std::{collections::HashSet, time::Instant};
 use bytes::Bytes;
 use http_body_util::combinators::BoxBody;
 use http_body_util::Full;
+use hyper::body::Body;
 use hyper::{header::HeaderName, Method, Request, Uri};
 use hyper_util::client::legacy::connect::Connect;
 use hyper_util::client::legacy::Client;
@@ -202,10 +203,14 @@ fn build_request(
         req = req.header("accept", "*/*");
     }
 
-    req.body(BoxBody::new(
-        body.map(|b| b.body.clone()).unwrap_or_default(),
-    ))
-    .or_throw(ctx)
+    let body_unwrap = body.map(|b| b.body.clone()).unwrap_or_default();
+
+    if !detected_headers.contains("content-length") {
+        let body_len = body_unwrap.size_hint().exact().unwrap_or_default();
+        req = req.header("content-length", body_len.to_string());
+    }
+
+    req.body(BoxBody::new(body_unwrap)).or_throw(ctx)
 }
 
 fn is_same_origin(uri: &Uri, initial_uri: &Uri) -> bool {


modules/llrt_http/src/fetch.rs
```diff
@@ -202,10 +203,14 @@ fn build_request(
...
-    ))
-    .or_throw(ctx)
+    let body_unwrap = body.map(|b| b.body.clone()).unwrap_or_default();
+
+    if !detected_headers.contains("content-length") {
+        let body_len = body_unwrap.size_hint().exact().unwrap_or_default();
+        req = req.header("content-length", body_len.to_string());
+    }
+
+    req.body(BoxBody::new(body_unwrap)).or_throw(ctx)
```
richarddavison, line 213: Nice finding! Should be set it for all request tho? GET, DELETE are maybe questionable?

richarddavison (changes requested):
Just some minor question!

modules/llrt_http/src/fetch.rs
floydspace: looks like should be set only if body is not empty,
at least I noticed such behaviour from postman, it auto sets content length if body is not empty, regardless of method

Sytten: I am 95% sure you dont need to set a content length explicitly. This is a bad practice anyway since the logic for content length is actually complex when you have to take into account compression and streaming (TE chunked).
This is handle by hyper in this function: https://github.com/hyperium/hyper/blob/0bd4adfed2bf40dda6867260ded05542d5b921c5/src/proto/h1/role.rs#L1484
But this is only called if the body has a known length, otherwise it will default to TE chunked:
https://github.com/hyperium/hyper/blob/0bd4adfed2bf40dda6867260ded05542d5b921c5/src/proto/h1/role.rs#L1298
It is possible that something broke when we previously switched to a BoxBody, but the solution is not to set an explicit header, we should find the root cause of the issue.

Sytten: The only way for a missing CL is if the body doesnt correctly implement the size hint which you can see the logic here:
https://github.com/hyperium/hyper/blob/0bd4adfed2bf40dda6867260ded05542d5b921c5/src/proto/h1/dispatch.rs#L339-L351

Note that it is a different logic for http/2 than http/1.1, I didnt dig into it but I would be very very very surprised if they didn't handle the CL.

Sytten (changes requested):
Following comments my I doubled check by sending a request in test to an NC and we do send a content length correctly.

Screenshot 2024-12-07 at 7 31 38 PM

Here is the test code:
```rust
#[tokio::test]
    async fn test_content_length() {
        test_async_with(|ctx| {
            crate::init(&ctx).unwrap();
            Box::pin(async move {
                let globals = ctx.globals();
                let run = async {
                    let fetch: Function = globals.get("fetch")?;

                    let headers = Object::new(ctx.clone())?;
                    headers.set("content-encoding", "gzip")?;
                    headers.set("content-language", "en")?;
                    headers.set("content-location", "/documents/foo.txt")?;
                    headers.set("content-type", "text/plain")?;
                    headers.set("authorization", "Basic YWxhZGRpbjpvcGVuc2VzYW1l")?;

                    let options = Object::new(ctx.clone())?;
                    options.set("redirect", "follow")?;
                    options.set("headers", headers.clone())?;
                    options.set("body", "Hello, LLRT!".as_bytes().to_vec())?;

                    // Method: POST, Redirect Pattern: None
                    options.set("method", "POST")?;
                    let url = "http://localhost:8113/expect/200/".to_string();

                    let response_promise: Promise = fetch.call((url, options.clone()))?;
                    let response: Class<Response> = response_promise.into_future().await?;
                    let mut response = response.borrow_mut();
                    let response_text = response.text(ctx.clone()).await?;

                    Ok(())
                };
                run.await.catch(&ctx).unwrap();
            })
        })
        .await;
    }
```

floydspace:
Thanks @Sytten I will double check it, and I agree with you, what you are saying makes total sense

richarddavison:
@floydspace I guess we can close this now then?

floydspace:
Yes please, anyway the fix in this pr is incorrect, thanks

PR #746, opened 2024-12-17
feat: adding support for Buffer.isBuffer and Buffer.isEncoding

lewisgcm:
### Issue # (if available)

No issue associated with this change.

### Description of changes
This PR adds support for `Buffer.isBuffer` and `Buffer.isEncoding` (see https://nodejs.org/api/buffer.html).

### Checklist

- [Y] Created unit tests in `tests/unit` and/or in Rust for my feature if needed
- [Y] Ran `make fix` to format JS and apply Clippy auto fixes
- [Y] Made sure my code didn't add any additional warnings: `make check`
- [NA] Added relevant type info in `types/` directory
- [NA] Updated documentation if needed ([API.md](API.md)/[README.md](README.md)/Other)

### DIFF
--- a/modules/llrt_buffer/src/buffer.rs
+++ b/modules/llrt_buffer/src/buffer.rs
@@ -15,8 +15,8 @@ use rquickjs::{
     function::{Constructor, Opt},
     module::{Declarations, Exports, ModuleDef},
     prelude::{Func, This},
-    Array, ArrayBuffer, Coerced, Ctx, Exception, IntoJs, JsLifetime, Object, Result, TypedArray,
-    Value,
+    Array, ArrayBuffer, Coerced, Ctx, Exception, Function, IntoJs, JsLifetime, Object, Result,
+    TypedArray, Value,
 };
 
 #[derive(JsLifetime)]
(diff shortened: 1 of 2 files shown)


modules/llrt_buffer/src/buffer.rs
```diff
@@ -120,6 +120,25 @@ fn byte_length<'js>(ctx: Ctx<'js>, value: Value<'js>, encoding: Opt<String>) ->
     ))
 }
 
+fn is_buffer<'js>(ctx: Ctx<'js>, value: Value<'js>) -> Result<bool> {
+    let class: Function = ctx.globals().get(stringify!(Buffer))?;
```
richarddavison, line 124: Use primordials here:
```rust
let constructor = BufferPrimordials::get(ctx)?.consturctor
```

modules/llrt_buffer/src/buffer.rs
```diff
@@ -120,6 +120,25 @@ fn byte_length<'js>(ctx: Ctx<'js>, value: Value<'js>, encoding: Opt<String>) ->
...
+    if let Some(object) = value.as_object() {
+        return Ok(object.is_instance_of(class));
+    }
+
+    Ok(false)
+}
+
+fn is_encoding(value: Value) -> Result<bool> {
+    if let Some(js_string) = value.as_string() {
+        if let Ok(std_string) = js_string.to_string() {
```
richarddavison, line 134: We should fail here if we can't convert the JsString to rust string
```rust
let string = js_string.to_string()?;
```

richarddavison (changes requested):
Thanks, some minor comments!

modules/llrt_buffer/src/buffer.rs
lewisgcm: Thanks, i've addressed this now in the next commit.

modules/llrt_buffer/src/buffer.rs
lewisgcm: Good catch, addressed in next commit

richarddavison (approved):

PR #767, opened 2025-01-03
feat: Increase webcrypto compatibility

richarddavison:
### Issue # (if available)

#184 

### Description of changes

Increase compatibility. More supported algorithms and methods.

TODO:
- [x] Export JWK
- [x] Import JWK
- [x] Import Spki
- [x] Import PKCS8
- [x] Derive bits
- [x] Wrap key
- [x] Unwrap key

### Checklist


modules/llrt_crypto/Cargo.toml
```diff
@@ -37,6 +37,15 @@ ctr = "0.9"
...
 p384 = "0.13"
+x25519-dalek = { version = "2", features = [
+  "static_secrets",
+  "zeroize",
+  "getrandom",
+] }
+spki = { version = "0.7", features = ["std"] }
+pkcs8 = { version = "0.10", features = ["std"] }
+der = { version = "0.7", features = ["derive"] }
+const-oid = { version = "0.9", features = ["db"] }
```
Sytten, line 48: I am a bit worried by the number of libraries that this module pulls in. I am wondering if we should consider putting an interface in between the primitives and the code here. Kinda similar to rustls with their provider implementation. That way we could have an aws-lc-rs based implementation for example that is fips compliant or a rust based implementation with these dependencies.

modules/llrt_crypto/src/lib.rs
```diff
@@ -250,6 +251,8 @@ impl ModuleDef for CryptoModule {
                 default.set(class_name, ctor)?;
             }
 
+            let crypto: Object = ctx.globals().get("crypto")?;
```
Sytten, line 254: Lets stop doing that, it is very annoying that the user of a given module is required to call the init function just because the module eval expect a global.

richarddavison: Agree!

modules/llrt_crypto/Cargo.toml
richarddavison: We'll work on that once crypto is feature complete. "rust" and "openssl" crypto providers can be used for this. Even a provider that spawns a "openssl" or "libressl" process

modules/llrt_crypto/src/lib.rs
richarddavison: This needs to be addressed in all places. I will create a PR fixing this where we'll try getting from globals first and if not existing create a new object. This requires refactoring the object created in init to be moved to a method.

Issue #773, opened 2025-01-07
`Buffer.from(str, "base64")` fails for URL safe base64 encoded strings


ainsleyrutterford:
When executing the following:

```javascript
// buffer.test.js

const str = "eyJodHRwczovL2F2YXRhcnMuZ2l0aHVidXNlcmNvbnRlbnQuY29tL3UvODk1MTgwOTU_dj00In0";

const buffer = Buffer.from(str, "base64");

console.log(buffer.toString());
```

With Node I get:

```
$ node test/buffer.test.js
{"https://avatars.githubusercontent.com/u/89518095?v=4"}
```

With LLRT I get:

```
$ ./llrt test/buffer.test.js           
Error: Base64Error
  at <anonymous> (/.../packages/lambda/test/buffer.test.js:3:23)
```

If I replace the `_` in `str` with `/` then it decodes the base64 string successfully in both Node and LLRT.

Node decodes both standard and URL safe base64 encoded strings, but LLRT only supports the standard charset at the moment as it uses the [`forgiving_decode_to_vec`](https://docs.rs/base64-simd/latest/base64_simd/fn.forgiving_decode_to_vec.html) function [here](https://github.com/awslabs/llrt/blob/main/libs/llrt_encoding/src/lib.rs#L141) which only supports the standard charset.

ainsleyrutterford:
Could LLRT first check which charset is being decoded by looking for the `+/` or `-_` characters, and then decode the string using the corresponding charset?

That doesn't solve the problem of `forgiving_decode_to_vec` only supporting the standard charset though :/

PR #774, opened 2025-01-08
Add support for decoding URL safe base64 encoded strings

ainsleyrutterford:
### Issue # (if available)

https://github.com/awslabs/llrt/issues/773

### Description of changes

This PR updates `Buffer.from(str, "base64")` to also support decoding [URL safe base64](https://en.wikipedia.org/wiki/Base64#URL_applications) encoded strings to make the implementation more consistent with Node's.

### Checklist

### DIFF
--- a/libs/llrt_encoding/src/lib.rs
+++ b/libs/llrt_encoding/src/lib.rs
@@ -138,7 +138,16 @@ pub fn bytes_to_b64_string(bytes: &[u8]) -> String {
 }
 
 pub fn bytes_from_b64(bytes: &[u8]) -> Result<Vec<u8>, String> {
-    base64_simd::forgiving_decode_to_vec(bytes).map_err(|e| e.to_string())
+    let standard_b64_bytes: Vec<u8> = bytes
+        .iter()
+        .map(|b| match b {
+            b'-' => b'+',
+            b'_' => b'/',
+            _ => *b,
+        })
+        .collect();
+
+    base64_simd::forgiving_decode_to_vec(&standard_b64_bytes).map_err(|e| e.to_string())
 }
 
 pub fn bytes_to_b64(bytes: &[u8]) -> Vec<u8> {
--- a/tests/unit/buffer.test.ts
+++ b/tests/unit/buffer.test.ts
@@ -81,12 +81,32 @@ describe("Buffer.from", () => {
   it("should create a buffer from a string with base64 encoding", () => {
     const input = "SGVsbG8sIHdvcmxkIQ==";
     const buffer = Buffer.from(input, "base64");
-
     expect(buffer.toString()).toEqual("Hello, world!");
 
     const input2 = "SGVsbG8sIHdvcmxkIQ";
     const buffer2 = Buffer.from(input2, "base64");
     expect(buffer2.toString()).toEqual("Hello, world!");
+  });
+
+  it("should create a buffer from a string with base64 encoding that contains / or +", () => {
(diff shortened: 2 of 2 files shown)


libs/llrt_encoding/src/lib.rs
```diff
@@ -138,7 +138,16 @@ pub fn bytes_to_b64_string(bytes: &[u8]) -> String {
...
+    let standard_b64_bytes: Vec<u8> = bytes
+        .iter()
+        .map(|b| match b {
+            b'-' => b'+',
+            b'_' => b'/',
+            _ => *b,
+        })
+        .collect();
+
+    base64_simd::forgiving_decode_to_vec(&standard_b64_bytes).map_err(|e| e.to_string())
```
richarddavison, line 150: This pretty much negates the use of simd as we traverse over all bytes.
What about using  `base64_simd::URL_SAFE_NO_PAD.decode_to_vec` instead?

richarddavison (changes requested):

libs/llrt_encoding/src/lib.rs
richarddavison: If it doesn't work, while still passing tests (being forgiving) we should should identify `-` or `_` via intrinsics and then use `URL_SAFE_NO_PAD`.
For example:
`memchr::memchr2_iter(b'-',b'_',bytes)` and replace in place without recreating and allocating new bytes

ainsleyrutterford: Yep good point, I'll try those out!

ainsleyrutterford: I looked into using one of the `base64_simd::{charset}.decode_to_vec` methods directly, but I think we need to continue using `forgiving_decode_to_vec` as otherwise we can't deal with both correctly padded strings and strings [with missing padding](https://github.com/awslabs/llrt/pull/120).

> replace in place without recreating and allocating new bytes

Since the current signature of `bytes_from_b64(bytes: &[u8])` has an immutable reference of `bytes`, wouldn't that only be possible if we update the function to take a mutable reference or use unsafe code in some way? (I'm very new to Rust so my understanding could be incorrect)

If we don't want to do that, what about something like this:

```rust
pub fn bytes_from_b64(bytes: &[u8]) -> Result<Vec<u8>, String> {
    let mut url_safe_byte_positions = memchr::memchr2_iter(b'-', b'_', bytes).peekable();

    if url_safe_byte_positions.peek().is_none() {
        return base64_simd::forgiving_decode_to_vec(bytes).map_err(|e| e.to_string());
    }

    // Replace any URL safe encoding bytes with their standard charset counterparts
    let mut standard_bytes = vec![0u8; bytes.len()];
    standard_bytes.copy_from_slice(bytes);
    for pos in url_safe_byte_positions {
        standard_bytes[pos] = match bytes[pos] {
            b'-' => b'+',
            b'_' => b'/',
            _ => unreachable!(),
        };
    }

    base64_simd::forgiving_decode_to_vec(&standard_bytes).map_err(|e| e.to_string())
}
```

It does require allocating a new `Vec`, but we only iterate through the few URL safe bytes we found?

richarddavison: I think we can change the bytes utility methods to accept a Cow. I'll checkout your branch and make some changes.

richarddavison:
@ainsleyrutterford this was automatically closed as I can't push to a non main branch from a fork.

PR #779, opened 2025-01-11
fix(core): fix a bug where a package's code could not load its own package

nabetti1720:
### Description of changes

For example, when `react-dom/server.edge` was called, this issue occurred because `react-dom` was required again inside the package.

```javascript
// reproduction.js
import * as dom from "react-dom/server.edge";
console.log(dom);
```

```
% llrt reproduction.js 
Error: IO Error: No such file or directory (os error 2)
  at <anonymous> (__cjs:/Users/shinya/Workspaces/llrt-test/node_modules/react-dom/cjs/react-dom-server.edge.development.js:7330:18)
      at <anonymous> (__cjs:/Users/shinya/Workspaces/llrt-test/node_modules/react-dom/cjs/react-dom-server.edge.development.js:8975:3)
      at <anonymous> (__cjs:/Users/shinya/Workspaces/llrt-test/node_modules/react-dom/server.edge.js:9:7)

% bun reproduction.js
Module {
  default: {
    version: "19.0.0",
    renderToReadableStream: [Function],
    renderToString: [Function],
    renderToStaticMarkup: [Function],
  },
  renderToReadableStream: [Function],
  renderToStaticMarkup: [Function],
  renderToString: [Function],
  version: "19.0.0",
}
```

This PR fixes this by storing the correct path when referencing self, to avoid IO errors.

### Checklist

### DIFF
--- a/llrt_core/src/module_loader/resolver.rs
+++ b/llrt_core/src/module_loader/resolver.rs
@@ -168,7 +168,7 @@ pub fn require_resolve<'a>(
     // 5. LOAD_PACKAGE_SELF(X, dirname(Y))
     if let Ok(Some(path)) = load_package_self(ctx, x, &dirname_y, is_esm) {
         trace!("+- Resolved by `LOAD_PACKAGE_SELF`: {}\n", path);
-        return to_abs_path(path.into());
+        return Ok(path.into());
     }
 
     // 6. LOAD_NODE_MODULES(X, dirname(Y))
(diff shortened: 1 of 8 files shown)


richarddavison:
Thanks! Looks like there is some Windows is sure with path / replacement?

nabetti1720:
> Thanks! Looks like there is some Windows issue with path / replacement?

It seems that's the case, but I haven't yet found the cause, so I'll need to investigate further...

llrt_core/src/module_loader/resolver.rs
```diff
@@ -168,7 +168,7 @@ pub fn require_resolve<'a>(
     // 5. LOAD_PACKAGE_SELF(X, dirname(Y))
     if let Ok(Some(path)) = load_package_self(ctx, x, &dirname_y, is_esm) {
         trace!("+- Resolved by `LOAD_PACKAGE_SELF`: {}\n", path);
-        return to_abs_path(path.into());
```
richarddavison, line 171: I think this is the problem. Since we're not replacing the backslashes in windows they will be escaped when loaded require.

richarddavison (changes requested):

richarddavison (approved):

PR #785, opened 2025-01-15
docs: add note about systemd to wsl build instructions

s-h-a-d-o-w:
### Description of changes

I would argue that this improvement is necessary because systemd isn't enabled by default and so currently, users have to look up what to do about the error `error: cannot communicate with server: Post "http://localhost/v2/snaps/zig": dial unix /run/snapd.socket: connect: no such file or directory`.

### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -377,45 +377,45 @@ There are many cases where LLRT shows notable performance drawbacks compared wit
 
 1. Clone code and cd to directory
 ```
-    git clone git@github.com:awslabs/llrt.git
-    cd llrt
+git clone git@github.com:awslabs/llrt.git
+cd llrt
 ```
 2. Install git submodules
 ```
-    git submodule update --init --checkout
+git submodule update --init --checkout
 ```
 3. Install rust
 ```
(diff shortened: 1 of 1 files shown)


richarddavison (changes requested):
Nice addition. Thank you;
I think we should add this as a:
> [!NOTE]
> `systemd` has to be enabled for WSL2. Read more at [Microsoft Devblogs](https://devblogs.microsoft.com/commandline/systemd-support-is-now-available-in-wsl/#how-can-you-get-systemd-on-your-machine)

s-h-a-d-o-w:
Sure thing!

But... are you sure about that change? Because unless I'm missing something, that would look like this - which is why I didn't make it an actual link to begin with. 😄

note

richarddavison:
No you're right, that looks awful. Maybe add an asterix or like you said:
> ### Windows WSL2 (requires systemd to be enabled*)
And then put an explanation  below the instruction like _*Read more at... link_

s-h-a-d-o-w:
How about this? I added enumeration because otherwise... same font size right next to one another looked awkward, even italicized.

richarddavison (approved):

PR #801, opened 2025-01-28
(perf): Fetch improvements

richarddavison:
### Description of changes

Limit parallel fetches to 500 to reduce contention on network stack. This significantly improves performance when a lot of requests are executed in parallel. 
This PR also contains a DNS cache implemementation that caches DNS lookups for 300 seconds. DNS parallel lookups are also limited to 5 per host to limit excessive thread spawning.
`Response` also now also only allows for a body to be read once.

```javascript
const res = await fetch("https:///www.amazon.com");
await res.text() //OK
await res.text() //Error
```

### Checklist

### DIFF
--- a/modules/llrt_http/src/cached_dns_resolver.rs
+++ b/modules/llrt_http/src/cached_dns_resolver.rs
@@ -1,0 +1,225 @@
+use std::collections::HashMap;
+use std::error::Error;
+use std::future::Future;
+use std::net::{SocketAddr, ToSocketAddrs};
+use std::pin::Pin;
+use std::sync::Arc;
+use std::task::{self, Poll};
+use std::time::{Duration, Instant};
+use std::{fmt, io, vec};
+
+use hyper_util::client::legacy::connect::dns::Name;
+use hyper_util::client::legacy::connect::HttpConnector;
+use tokio::sync::Semaphore;
+use tokio::{sync::RwLock, task::JoinHandle};
+use tower_service::Service;
+
+#[derive(Clone)]
+struct CacheEntry {
+    timestamp: Instant,
+    gai_addrs: GaiAddrs,
+}
+
+struct CacheLock {
+    semaphore: Arc<Semaphore>,
+    entry: Option<CacheEntry>,
+}
+
+/// A resolver using blocking `getaddrinfo` calls in a threadpool.
+#[derive(Clone)]
+pub struct CachedDnsResolver {
+    cache: Arc<RwLock<HashMap<Name, Arc<CacheLock>>>>,
+    cache_duration: Duration,
+}
+
+/// An iterator of IP addresses returned from `getaddrinfo`.
+#[derive(Clone)]
+pub struct GaiAddrs {
+    inner: SocketAddrs,
+}
+
+/// A future to resolve a name returned by `GaiResolver`.
+pub struct GaiFuture {
+    inner: JoinHandle<Result<GaiAddrs, io::Error>>,
+}
+
+/// Error indicating a given string was not a valid domain name.
+#[derive(Debug)]
+pub struct InvalidNameError(());
+
+impl fmt::Display for InvalidNameError {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.write_str("Not a valid domain name")
+    }
+}
+
+impl Error for InvalidNameError {}
+
+impl Default for CachedDnsResolver {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+impl CachedDnsResolver {
+    pub fn new() -> Self {
+        Self {
+            cache: Arc::new(RwLock::new(HashMap::new())),
+            cache_duration: Duration::from_secs(300),
+        }
+    }
+
+    pub fn into_http_connector(self) -> HttpConnector<Self> {
+        HttpConnector::<Self>::new_with_resolver(self)
+    }
+}
+
+impl Service<Name> for CachedDnsResolver {
+    type Response = GaiAddrs;
+    type Error = io::Error;
+    type Future = GaiFuture;
+
+    fn poll_ready(&mut self, _cx: &mut task::Context<'_>) -> Poll<Result<(), io::Error>> {
+        Poll::Ready(Ok(()))
+    }
+
+    fn call(&mut self, name: Name) -> Self::Future {
+        let cache = self.cache.clone();
+        let duration = self.cache_duration;
+        let handle = tokio::task::spawn(async move {
+            let entry = {
+                let cache_read = cache.read().await;
+                if let Some(entry) = cache_read.get(&name) {
+                    entry.clone()
+                } else {
+                    drop(cache_read); // Release read lock before acquiring write lock
+                    let mut cache_write = cache.write().await;
+                    cache_write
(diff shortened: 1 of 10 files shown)


modules/llrt_http/src/cached_dns_resolver.rs
```diff
@@ -0,0 +1,227 @@
...
+
+struct CacheLock {
+    semaphore: Arc<Semaphore>,
+    entry: Option<CacheEntry>,
+}
+
+/// A resolver using blocking `getaddrinfo` calls in a threadpool.
+#[derive(Clone)]
+pub struct CachedDnsResolver {
+    cache: Arc<RwLock<HashMap<Name, Arc<CacheLock>>>>,
```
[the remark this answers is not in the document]
Sytten: I used `use quick_cache::sync::Cache;` which is really made for that use-case. The performances would be much better than you current lock.

It's not perfect since it misses the TTL, but that could be easily added.

modules/llrt_http/src/cached_dns_resolver.rs
```diff
@@ -0,0 +1,227 @@
...
+            }
+
+            let name2 = name.clone();
+
+            let address = tokio::task::spawn_blocking(move || {
+                (name2.as_str(), 0)
+                    .to_socket_addrs()
+                    .map(|i| SocketAddrs { iter: i })
+            })
+            .await;
```
Sytten, line 130: Tokio provides DNS methods that do that, use it will simplify your code.

modules/llrt_http/src/fetch.rs
```diff
@@ -65,13 +72,39 @@ where
...
+                                2
+                            } else if error.starts_with("Connect") {
+                                if error.contains("dns error") {
+                                    0
+                                } else {
+                                    2
+                                }
+                            } else {
+                                0
+                            };
```
Sytten, line 97: This is ugly AF, comment is unclear.
I think you should be able to use the `.source` to go up the tree of errors instead and do some casting instead.

Sytten: Do we even want to retry? I feel like this is a user choice? I am surprised this is not in the fetch spec.

modules/llrt_http/src/lib.rs
```diff
@@ -132,11 +145,15 @@ pub static HTTP_CLIENT: Lazy<io::Result<HyperClient>> = Lazy::new(|| {
...
+    let mut cache_dns_connector = CachedDnsResolver::new().into_http_connector();
+    cache_dns_connector.enforce_http(false);
+
     let https = match get_http_version() {
-        #[cfg(feature = "http1")]
-        HttpVersion::Http1_1 => builder.enable_http1().build(),
         #[cfg(feature = "http2")]
-        HttpVersion::Http2 => builder.enable_all_versions().build(),
+        HttpVersion::Http2 => builder
+            .enable_all_versions()
```
Sytten, line 154: This method is not available AFAIK when the flag `h2` is not enabled on hyper.

modules/llrt_http/src/lib.rs
```diff
@@ -103,8 +111,12 @@ pub fn set_http_version(version: HttpVersion) {
 }
 
 fn get_http_version() -> HttpVersion {
-    HTTP_VERSION.get().cloned().unwrap_or({
-        #[cfg(feature = "http2")]
+    *HTTP_VERSION.get_or_init(|| {
+        #[cfg(any(feature = "http2", feature = "http1"))]
+        {
+            HttpVersion::Http1_1
+        }
```
Sytten, line 118: Unsure if you really want to use http/1.1 by default?

modules/llrt_http/src/response.rs
```diff
@@ -101,7 +102,22 @@ static STATUS_TEXTS: Lazy<HashMap<u16, &'static str>> = Lazy::new(|| {
...
     Incoming(Option<hyper::Response<Incoming>>),
     Cloned(Option<hyper::Response<IncomingReceiver>>),
-    Provided(Value<'js>),
+    Provided(Option<Value<'js>>),
+    Empty,
+}
+
+#[rquickjs::class]
+pub struct Response<'js> {
+    body: RwLock<BodyVariant<'js>>,
```
Sytten, line 111: I am not sure I see the advantage of the lock, especially an async one which is less performant. It would warrant a comment.
It is safer for sure on the JS side I guess since the `&mut self` can cause panics kinda easily on an async method.

modules/llrt_http/src/cached_dns_resolver.rs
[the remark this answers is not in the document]
richarddavison: This is much cleaner. I modified the existing dns implementation from hyper, hence the verboseness. I also totally neglected the fact that it was unbounded, some sort of LRU cache should suffice. Will take a look at quick_cache!

modules/llrt_http/src/cached_dns_resolver.rs
```diff
@@ -0,0 +1,227 @@
...
+                        }),
+                    );
+                    Ok(gai_addrs)
+                },
+                Ok(Err(err)) => Err(err),
+                Err(join_err) => {
+                    if join_err.is_cancelled() {
+                        Err(io::Error::new(io::ErrorKind::Interrupted, join_err))
+                    } else {
+                        panic!("gai background task failed: {:?}", join_err)
```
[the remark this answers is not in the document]
richarddavison: This is derived from:
https://github.com/hyperium/hyper-util/blob/26e1e15faaeed2659e48c1e13a7b13e7f4d37bda/src/client/legacy/connect/dns.rs#L149

modules/llrt_http/src/cached_dns_resolver.rs
```diff
@@ -0,0 +1,227 @@
...
+    type Future = GaiFuture;
+
+    fn poll_ready(&mut self, _cx: &mut task::Context<'_>) -> Poll<Result<(), io::Error>> {
+        Poll::Ready(Ok(()))
+    }
+
+    fn call(&mut self, name: Name) -> Self::Future {
+        let cache = self.cache.clone();
+        let duration = self.cache_duration;
+        let handle = tokio::task::spawn(async move {
```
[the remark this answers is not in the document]
richarddavison: Maybe can work without spawn, will try with boxed future

modules/llrt_http/src/fetch.rs
richarddavison: Yeah super ugly, I'll remove retries. This was a side effect of contention issues.

modules/llrt_http/src/lib.rs
richarddavison: Sadly 1.1 has much better performance when running multiple requests to same host concurrently. There is probably contention in multiplexing causing this.

modules/llrt_http/src/response.rs
richarddavison: This is a `sync` RwLock. It needs to be accessible concurrently. Without it doing something like:
`await Promise.all([response.text(),response.text()] `
Would crash the RT

Sytten: I am wondering if rquickjs should stop allowing mut self it seems like at least in async context it is unsound.

PR #806, opened 2025-01-30
(perf): Fetch improvements v2

richarddavison:
### Description of changes

Limit parallel fetches to 500 to reduce contention on network stack. This significantly improves performance when a lot of requests are executed in parallel. 
This PR also contains a DNS cache implemementation that caches DNS lookups for 300 seconds. DNS parallel lookups are also limited to 5 per host to limit excessive thread spawning.
`Response` also now also only allows for a body to be read once.

```javascript
const res = await fetch("https:///www.amazon.com");
await res.text() //OK
await res.text() //Error
```

### Checklist

### DIFF
--- a/modules/llrt_http/src/cached_dns_resolver.rs
+++ b/modules/llrt_http/src/cached_dns_resolver.rs
@@ -9,7 +9,7 @@ use std::{io, vec};
 use hyper_util::client::legacy::connect::dns::Name;
 use hyper_util::client::legacy::connect::HttpConnector;
 use quick_cache::sync::Cache;
-use std::io::{ErrorKind, Result};
+use std::io::Result;
 use tokio::sync::Semaphore;
 use tower_service::Service;
 
@@ -34,13 +34,13 @@ struct CacheEntry {
 
 #[derive(Clone)]
 struct CacheConcurrencyGuard {
-    semaphore: Option<Arc<Semaphore>>,
+    semaphore: Arc<Semaphore>,
     entry: Option<CacheEntry>,
 }
 impl CacheConcurrencyGuard {
     fn new(permits: u8) -> Self {
         Self {
-            semaphore: Some(Arc::new(Semaphore::new(permits as usize))),
+            semaphore: Arc::new(Semaphore::new(permits as usize)),
             entry: None,
         }
     }
@@ -82,10 +82,9 @@ impl Service<Name> for CachedDnsResolver {
                 }
             };
 
-            let binding = guard
-                .semaphore
-                .ok_or_else(|| io::Error::new(ErrorKind::InvalidData, "Missing semaphore"))?;
-            let lock = binding.acquire().await.unwrap();
+            let semaphore = guard.semaphore;
+            let semaphore2 = semaphore.clone();
+            let lock = semaphore2.acquire().await.unwrap();
 
             if let Some(item) = cache.get(&name).and_then(|guard| guard.entry) {
                 return Ok(item.addrs);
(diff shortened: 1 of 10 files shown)


modules/llrt_http/src/cached_dns_resolver.rs
```diff
@@ -0,0 +1,139 @@
...
+            };
+            if let Some(entry) = guard.entry {
+                if entry.ttl > Instant::now() {
+                    return Ok(entry.addrs);
+                }
+            };
+
+            let binding = guard
+                .semaphore
+                .ok_or_else(|| io::Error::new(ErrorKind::InvalidData, "Missing semaphore"))?;
```
Sytten, line 87: This is unsound. Consider the following:
1. Cache is empty, you ask for domain `example.com`
2. Cache add a `CacheConcurrencyGuard` with a semaphore
3. Semaphore is acquired and domain is resolved
4. Replace the name with resolved entry and semaphore none
5. Time passes (more than TTL)
6. You ask `example.com` again, cache returns element
7. entry ttl is expired (line 80), we continue
8. semaphore is None, return an error

Sytten: Maybe `moka` would be a better cache since it has built-in time to live expiration (https://docs.rs/moka/0.12.10/moka/future/struct.CacheBuilder.html#method.time_to_live).
It is more heavy than quick-cache though and I have not tried it myself yet.

Sytten (changes requested):

modules/llrt_http/src/cached_dns_resolver.rs
gc-victor: There is a mini version of moka.
https://github.com/moka-rs/mini-moka

Sytten: It doesnt have the atomic insertion, but we might not need it.
https://github.com/moka-rs/moka?tab=readme-ov-file#choosing-the-right-cache-for-your-use-case

modules/llrt_http/src/cached_dns_resolver.rs
```diff
@@ -0,0 +1,138 @@
...
+#[derive(Clone)]
+struct CacheEntry {
+    ttl: Instant,
+    addrs: SocketAddrs,
+}
+
+#[derive(Clone)]
+struct CacheConcurrencyGuard {
+    semaphore: Arc<Semaphore>,
+    entry: Option<CacheEntry>,
```
Sytten, line 38: Considering that each get will clone is it worth Arc?

modules/llrt_http/src/cached_dns_resolver.rs
```diff
@@ -0,0 +1,138 @@
...
+    fn poll_ready(&mut self, _cx: &mut task::Context<'_>) -> Poll<Result<()>> {
+        Poll::Ready(Ok(()))
+    }
+
+    fn call(&mut self, name: Name) -> Self::Future {
+        let cache = self.cache.clone();
+        let permits = self.concurrency;
+        let ttl = self.ttl;
+
+        Box::pin(async move {
```
Sytten, line 70: IMO we still need a global semaphore even if large to avoid spawning hundred of threads if each request ask for a different domain.

richarddavison: This will be up to the caller. In LLRT this is limited by the fetch call, I would like to avoid double semaphores.

modules/llrt_http/src/cached_dns_resolver.rs
richarddavison: Yeah needs to be!

PR #807, opened 2025-01-30
Add some functions to read and write uid/gid/euid/egid

ahaoboy:
### Issue # (if available)

Add some functions to read and write uid/gid/euid/egid
 
### Description of changes

These functions are only added on Unix platforms, and the setter functions only support number

### Checklist


richarddavison (approved):
LGTM

types/process.d.ts
```diff
@@ -183,5 +183,181 @@ declare module "process" {
...
+     */
+    getgid?: () => number
+    /**
+     * The `process.setgid()` method sets the group identity of the process. (See [`setgid(2)`](http://man7.org/linux/man-pages/man2/setgid.2.html).) The `id` can be passed as either a
+     * numeric ID or a group name
+     * string. If a group name is specified, this method blocks while resolving the
+     * associated numeric ID.
+     *
+     * ```js
+     * import process from 'node:process';
```
Sytten, line 211: Lets remove node references

Sytten (changes requested):

Sytten:
We should start thinking about a way to lockdown functionalities. Ideally runtime, like the v8 isolate.

types/process.d.ts
ahaoboy: We should also think about how to maintain the type in a simple way. Copying the type declaration of node and then removing `node:` doesn't seem to be a good idea.

PR #813, opened 2025-02-01
fix(utils): remove expect from ObjectBytes

### LINKED ISSUE #700: ObjectByres should not expect on detached buffer
Whoever has time, we need to make the `as_bytes` function return a result and fix all the places where it it called.

nabetti1720:
### Issue # (if available)

Closed #700

### Description of changes

This PR removes the `expect` instruction used in ObjectBytes's `as_bytes()` and allows it to return `std::result::Result` or `rquickjs::Result` types.

### Checklist

### DIFF
--- a/modules/llrt_crypto/src/lib.rs
+++ b/modules/llrt_crypto/src/lib.rs
@@ -246,9 +246,12 @@ impl ModuleDef for CryptoModule {
                 let class_name: &str = sha_algorithm.class_name();
                 let algo = sha_algorithm;
 
-                let ctor = Constructor::new_class::<ShaHash, _, _>(ctx.clone(), move || {
-                    ShaHash::default(algo.clone())
-                })?;
+                let ctor =
+                    Constructor::new_class::<ShaHash, _, _>(ctx.clone(), move |ctx, secret| {
+                        struct Args<'js>(Ctx<'js>, Opt<ObjectBytes<'js>>);
+                        let Args(ctx, secret) = Args(ctx, secret);
+                        ShaHash::new(ctx, algo.clone(), secret)
+                    })?;
 
                 default.set(class_name, ctor)?;
             }
--- a/modules/llrt_crypto/src/lib.rs
+++ b/modules/llrt_crypto/src/lib.rs
@@ -246,8 +246,8 @@ impl ModuleDef for CryptoModule {
                 let class_name: &str = sha_algorithm.class_name();
                 let algo = sha_algorithm;
 
-                let ctor = Constructor::new_class::<ShaHash, _, _>(ctx.clone(), move |secret| {
-                    ShaHash::new(algo.clone(), secret)
+                let ctor = Constructor::new_class::<ShaHash, _, _>(ctx.clone(), move || {
+                    ShaHash::default(algo.clone())
                 })?;
 
                 default.set(class_name, ctor)?;
--- a/modules/llrt_crypto/src/sha_hash.rs
+++ b/modules/llrt_crypto/src/sha_hash.rs
@@ -28,7 +28,7 @@ impl Hmac {
         let algorithm = *algorithm.hmac_algorithm();
 
         Ok(Self {
-            context: HmacContext::with_key(&hmac::Key::new(algorithm, key_value.as_bytes())),
+            context: HmacContext::with_key(&hmac::Key::new(algorithm, key_value.as_bytes(&ctx)?)),
         })
     }
 
@@ -44,9 +44,10 @@ impl Hmac {
 
     fn update<'js>(
         this: This<Class<'js, Self>>,
+        ctx: Ctx<'js>,
         bytes: ObjectBytes<'js>,
     ) -> Result<Class<'js, Self>> {
-        let bytes = bytes.as_bytes();
+        let bytes = bytes.as_bytes(&ctx)?;
         this.0.borrow_mut().context.update(bytes);
 
         Ok(this.0)
@@ -94,9 +95,10 @@ impl Hash {
     #[qjs(rename = "update")]
     fn hash_update<'js>(
         this: This<Class<'js, Self>>,
+        ctx: Ctx<'js>,
         bytes: ObjectBytes<'js>,
     ) -> Result<Class<'js, Self>> {
-        let bytes = bytes.as_bytes();
+        let bytes = bytes.as_bytes(&ctx)?;
         this.0.borrow_mut().context.update(bytes);
         Ok(this.0)
     }
@@ -195,11 +197,27 @@ pub struct ShaHash {
 #[rquickjs::methods]
 impl ShaHash {
     #[qjs(skip)]
-    pub fn new(algorithm: ShaAlgorithm, secret: Opt<ObjectBytes<'_>>) -> Result<Self> {
-        let secret = secret.0.map(|bytes| bytes.into_bytes());
+    pub fn new<'js>(
+        ctx: &Ctx<'js>,
+        algorithm: ShaAlgorithm,
+        secret: Opt<ObjectBytes<'js>>,
+    ) -> Result<Self> {
+        let secret = secret
+            .0
+            .map(|bytes| bytes.into_bytes(ctx))
+            .and_then(|res| res.ok());
 
         Ok(ShaHash {
             secret,
+            bytes: Vec::new(),
+            algorithm,
+        })
+    }
+
+    #[qjs(skip)]
+    pub fn default(algorithm: ShaAlgorithm) -> Result<Self> {
+        Ok(ShaHash {
+            secret: None,
             bytes: Vec::new(),
             algorithm,
         })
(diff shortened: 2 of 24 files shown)


modules/llrt_crypto/src/lib.rs
```diff
@@ -246,8 +246,8 @@ impl ModuleDef for CryptoModule {
                 let class_name: &str = sha_algorithm.class_name();
                 let algo = sha_algorithm;
 
-                let ctor = Constructor::new_class::<ShaHash, _, _>(ctx.clone(), move |secret| {
-                    ShaHash::new(algo.clone(), secret)
+                let ctor = Constructor::new_class::<ShaHash, _, _>(ctx.clone(), move || {
```
Sytten, line 249: Need an explanation here this is weird and not related to the pr

modules/llrt_crypto/src/sha_hash.rs
```diff
@@ -195,8 +197,15 @@ pub struct ShaHash {
...
-        let secret = secret.0.map(|bytes| bytes.into_bytes());
+    pub fn new<'js>(
+        ctx: &Ctx<'js>,
+        algorithm: ShaAlgorithm,
+        secret: Opt<ObjectBytes<'js>>,
+    ) -> Result<Self> {
+        let secret = secret
+            .0
+            .map(|bytes| bytes.into_bytes(ctx))
+            .and_then(|res| res.ok());
```
Sytten, line 208: That is effectively silencing the error which is likely not what we want use transpose instead and return the error

modules/llrt_crypto/src/subtle/encryption_algorithm.rs
```diff
@@ -78,7 +78,9 @@ impl<'js> FromJs<'js> for EncryptionAlgorithm {
 
                 let additional_data = obj
                     .get_optional::<_, ObjectBytes>("additionalData")?
-                    .map(|v| v.into_bytes().into_boxed_slice());
+                    .map(|v| v.into_bytes(ctx))
+                    .and_then(|res| res.ok())
```
Sytten, line 82: Same here

modules/llrt_crypto/src/subtle/encryption_algorithm.rs
```diff
@@ -99,7 +101,10 @@ impl<'js> FromJs<'js> for EncryptionAlgorithm {
             "RSA-OAEP" => {
                 let label = obj
                     .get_optional::<_, ObjectBytes>("label")?
-                    .map(|bytes| bytes.into_bytes().into_boxed_slice());
+                    .map(|bytes| bytes.into_bytes(ctx))
+                    .and_then(|res| res.ok())
```
Sytten, line 105: And here

Sytten (changes requested):

modules/llrt_crypto/src/lib.rs
nabetti1720: `ShaHash::new()` calls `ObjectBytes::into_bytes()`, but because this function requires `ctx` as an argument, it was also necessary to add `ctx` to the arguments of `ShaHash::new()`.

Doing so caused a lifetime problem, which I was unable to resolve, so I prepared the simpler `ShaHash::default()`.

image

nabetti1720: image

Looks like we've made some progress. Is this error telling us that `ShaHash` needs to implement the `IntoJs` trait?

Sytten: It is just that it changes the constructor definition that was my concern. We should be able to get the context from somewhere.

nabetti1720: The lifetime issue occurred because the closure parameters did not include `ctx`.
However, I still get the following error:

```rust
                let ctor =
                    Constructor::new_class::<ShaHash, _, _>(ctx.clone(), move |ctx, secret| {
                        ShaHash::new(ctx, algo.clone(), secret)
                    })?;
```
```
error[E0277]: the trait bound `{closure@modules/llrt_crypto/src/lib.rs:254:74: 254:92}: IntoJsFunc<'_, _>` is not satisfied
   --> modules/llrt_crypto/src/lib.rs:254:55
    |
254 |                     Constructor::new_class::<ShaHash, _, _>(ctx.clone(), move |ctx, secret| {
    |                                                       ^ the trait `IntoJsFunc<'_, _>` is not implemented for closure `{closure@modules/llrt_crypto/src/lib.rs:254:74: 254:92}`
    |
    = help: the following other types implement trait `IntoJsFunc<'js, P>`:
              `MutFn<Fun>` implements `IntoJsFunc<'js, ()>`
              `MutFn<Fun>` implements `IntoJsFunc<'js, (A, B)>`
              `MutFn<Fun>` implements `IntoJsFunc<'js, (A, B, C)>`
              `MutFn<Fun>` implements `IntoJsFunc<'js, (A, B, C, D)>`
              `MutFn<Fun>` implements `IntoJsFunc<'js, (A, B, C, D, E)>`
              `MutFn<Fun>` implements `IntoJsFunc<'js, (A, B, C, D, E, F)>`
              `MutFn<Fun>` implements `IntoJsFunc<'js, (A, B, C, D, E, F, G)>`
              `MutFn<Fun>` implements `IntoJsFunc<'js, (A, B, C, D, E, F, G, H)>`
            and 41 others
note: required by a bound in `rquickjs::function::Constructor::<'js>::new_class`
   --> /Users/shinya/.cargo/git/checkouts/rquickjs-c5d9b5f474075f2d/6a60607/core/src/value/function.rs:199:12
    |
197 |     pub fn new_class<C, F, P>(ctx: Ctx<'js>, f: F) -> Result<Self>
    |            --------- required by a bound in this associated function
198 |     where
199 |         F: IntoJsFunc<'js, P> + 'js,
    |            ^^^^^^^^^^^^^^^^^^ required by this bound in `Constructor::<'js>::new_class`
```

At least at this point, I don't need to specify a secret variable for `ShaHash`, so I think a simple constructor like `ShaHash::default()` would be sufficient, right?

nabetti1720: Sorry, I understand why you didn't get the context.
But I'm still getting the error...

```rust
                let ctor =
                    Constructor::new_class::<ShaHash, _, _>(ctx.clone(), move |ctx, secret| {
                        ShaHash::new(ctx, algo.clone(), secret)
                    })?;
```
```
error[E0277]: the trait bound `{closure@modules/llrt_crypto/src/lib.rs:250:74: 250:92}: IntoJsFunc<'_, _>` is not satisfied
   --> modules/llrt_crypto/src/lib.rs:250:55
    |
250 |                     Constructor::new_class::<ShaHash, _, _>(ctx.clone(), move |ctx, secret| {
    |                                                       ^ the trait `IntoJsFunc<'_, _>` is not implemented for closure `{closure@modules/llrt_crypto/src/lib.rs:250:74: 250:92}`
    |
    = help: the following other types implement trait `IntoJsFunc<'js, P>`:
              `MutFn<Fun>` implements `IntoJsFunc<'js, ()>`
              `MutFn<Fun>` implements `IntoJsFunc<'js, (A, B)>`
              `MutFn<Fun>` implements `IntoJsFunc<'js, (A, B, C)>`
              `MutFn<Fun>` implements `IntoJsFunc<'js, (A, B, C, D)>`
              `MutFn<Fun>` implements `IntoJsFunc<'js, (A, B, C, D, E)>`
              `MutFn<Fun>` implements `IntoJsFunc<'js, (A, B, C, D, E, F)>`
              `MutFn<Fun>` implements `IntoJsFunc<'js, (A, B, C, D, E, F, G)>`
              `MutFn<Fun>` implements `IntoJsFunc<'js, (A, B, C, D, E, F, G, H)>`
            and 41 others
note: required by a bound in `rquickjs::function::Constructor::<'js>::new_class`
   --> /Users/shinya/.cargo/git/checkouts/rquickjs-c5d9b5f474075f2d/6a60607/core/src/value/function.rs:199:12
    |
197 |     pub fn new_class<C, F, P>(ctx: Ctx<'js>, f: F) -> Result<Self>
    |            --------- required by a bound in this associated function
198 |     where
199 |         F: IntoJsFunc<'js, P> + 'js,
    |            ^^^^^^^^^^^^^^^^^^ required by this bound in `Constructor::<'js>::new_class`
 ```

Comment (Sytten) [reply modules/llrt_crypto/src/lib.rs:249]: I think from what I remember you do have to use a tuple for parameters of a js function.
I will have to check the code I dont really have all the context here to understand if we need the secret parameter or not in the ctr.

Comment (richarddavison) [reply modules/llrt_crypto/src/lib.rs:249]: Yeah this changes the constructor. I suspect this is a lifetime issue.
Try with something like this:
```rust
 let ctor =
                    Constructor::new_class::<ShaHash, _, _>(ctx.clone(), move |ctx, secret| {
                        struct Args<'js>(Ctx<'js>,Opt<ObjectBytes<'js>>);
                        let Args(ctx,secret) = Args(ctx,secret);
                        ShaHash::new(ctx, algo.clone(), secret)
                    })?;
```

richarddavison (changes requested):
Thanks, just a minor comment regarding object bytes

nabetti1720:
Thanks @Sytten and @richarddavison. I think I've responded to all the reviews.

richarddavison (approved):

Issue #828, opened 2025-02-10
support fs.chmod and fs.chmodSync


ahaoboy:
https://nodejs.org/api/fs.html#fschmodpath-mode-callback

PR #835, opened 2025-02-10
fs: add chmodSync

ahaoboy:
### Issue # (if available)

close: https://github.com/awslabs/llrt/issues/828

### Checklist

### DIFF
--- a/modules/llrt_fs/src/chmod.rs
+++ b/modules/llrt_fs/src/chmod.rs
@@ -4,6 +4,7 @@ use rquickjs::{Ctx, Result};
 #[cfg(unix)]
 use std::os::unix::prelude::PermissionsExt;
 
+#[cfg(unix)]
 pub(crate) fn chmod_error(path: &str) -> String {
     ["Can't set permissions of \"", path, "\""].concat()
 }
--- a/modules/llrt_fs/src/chmod.rs
+++ b/modules/llrt_fs/src/chmod.rs
@@ -1,0 +1,36 @@
+#[cfg(unix)]
+use llrt_utils::result::ResultExt;
+use rquickjs::{Ctx, Result};
+#[cfg(unix)]
+use std::os::unix::prelude::PermissionsExt;
+
+pub async fn chmod(ctx: Ctx<'_>, path: String, mode: u32) -> Result<()> {
+    #[cfg(unix)]
+    {
+        tokio::fs::set_permissions(&path, PermissionsExt::from_mode(mode))
+            .await
+            .or_throw_msg(&ctx, &["Can't set permissions of \"", &path, "\""].concat())?;
+    }
+    #[cfg(not(unix))]
+    {
+        _ = ctx;
+        _ = path;
+        _ = mode;
+    }
+    Ok(())
+}
+
+pub fn chmod_sync(ctx: Ctx<'_>, path: String, mode: u32) -> Result<()> {
+    #[cfg(unix)]
+    {
+        std::fs::set_permissions(&path, PermissionsExt::from_mode(mode))
+            .or_throw_msg(&ctx, &["Can't set permissions of \"", &path, "\""].concat())?;
+    }
+    #[cfg(not(unix))]
+    {
+        _ = ctx;
+        _ = path;
+        _ = mode;
+    }
+    Ok(())
+}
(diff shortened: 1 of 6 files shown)


modules/llrt_fs/src/chmod.rs
```diff
@@ -0,0 +1,36 @@
...
+            .await
+            .or_throw_msg(&ctx, &["Can't set permissions of \"", &path, "\""].concat())?;
+    }
+    #[cfg(not(unix))]
+    {
+        _ = ctx;
+        _ = path;
+        _ = mode;
+    }
+    Ok(())
```
richarddavison, line 20: Is this the correct behavior on non-unix planforms?

modules/llrt_fs/src/chmod.rs
```diff
@@ -0,0 +1,36 @@
...
+}
+
+pub fn chmod_sync(ctx: Ctx<'_>, path: String, mode: u32) -> Result<()> {
+    #[cfg(unix)]
+    {
+        std::fs::set_permissions(&path, PermissionsExt::from_mode(mode))
+            .or_throw_msg(&ctx, &["Can't set permissions of \"", &path, "\""].concat())?;
```
richarddavison, line 27: Parsing of permissions can be moved to method to avoid code duplication

modules/llrt_fs/src/chmod.rs
ahaoboy: Both nodejs and deno treat chmodSync as an empty function on Windows

https://github.com/denoland/deno/blob/33bccf9090e8764790be84e56fe0988600ffe31a/ext/io/lib.rs#L819

richarddavison: Not really? They throw an error, if I'm not misstaken: 🤔 
https://github.com/denoland/deno/blob/33bccf9090e8764790be84e56fe0988600ffe31a/ext/io/lib.rs#L827C1-L828C31

modules/llrt_fs/src/chmod.rs
richarddavison: I think we can just move out this to a method and call :)
PermissionsExt::from_mode(mode)).or_throw_msg(&ctx, &["Can't set permissions of \"", &path, "\""].concat()

modules/llrt_fs/src/chmod.rs
ahaoboy: Maybe it's ignored at some level? The following code will not throw an error when run with node/bun/deno, nor will it have any effect

```
const fs = require('fs')
console.log(fs.statSync('./readme.md') ,fs.chmodSync("./readme.md", 0))
console.log(fs.statSync('./readme.md') ,fs.chmodSync("./readme.md", 0))
```

modules/llrt_fs/src/chmod.rs
ahaoboy: Something like this?

```
pub(crate) fn chmod_error(path: &str) -> String {
    ["Can't set permissions of \"", path, "\""].concat()
}

```

richarddd: Almost but more like:
```rust
#[cfg(unix)]
fn parse_permissions(ctx: &Ctx, mode:&str)-> Result<ReturnTypeHere>{
  PermissionsExt::from_mode(mode)).or_throw_msg(&ctx, &["Can't set permissions of \"", &path, "\""].concat())
}
```

ahaoboy: But or_throw_msg calls from std::fs::set_permissions and tokio::fs::set_permissions not from_mode

```rs
        std::fs::set_permissions(path, PermissionsExt::from_mode(mode))
            .or_throw_msg(&ctx, &chmod_error(path))?;

        tokio::fs::set_permissions(path, PermissionsExt::from_mode(mode))
            .await
            .or_throw_msg(&ctx, &chmod_error(path))?;
```

modules/llrt_fs/src/chmod.rs
richarddavison: Is this on Windows?

ahaoboy: Yes, the function exists on the Windows platform, but it seems to be an empty function and has no effect after execution.

modules/llrt_fs/src/chmod.rs
richarddavison: Oh I see. My misstake!

richarddavison (approved):

PR #869, opened 2025-03-05
feat(buffer): implement some static and prototype methods

nabetti1720:
### Description of changes

This PR implements some methods that I need for my project.
- [Static method: Buffer.allocUnsafe(size)](https://nodejs.org/api/buffer.html#static-method-bufferallocunsafesize)
- [Static method: Buffer.allocUnsafeSlow(size)](https://nodejs.org/api/buffer.html#static-method-bufferallocunsafeslowsize)
- [buf.copy(target[, targetStart[, sourceStart[, sourceEnd]]])](https://nodejs.org/api/buffer.html#bufcopytarget-targetstart-sourcestart-sourceend)
- [buf.write(string[, offset[, length]][, encoding])](https://nodejs.org/api/buffer.html#bufwritestring-offset-length-encoding)
- [buf.writeInt32BE(value[, offset])](https://nodejs.org/api/buffer.html#bufwriteint32bevalue-offset)

By generalizing `buf.writeInt32BE()`, we also support:
- [buf.writeBigInt64BE(value[, offset])](https://nodejs.org/api/buffer.html#bufwritebigint64bevalue-offset)
- [buf.writeBigInt64LE(value[, offset])](https://nodejs.org/api/buffer.html#bufwritebigint64levalue-offset)
- [buf.writeDoubleBE(value[, offset])](https://nodejs.org/api/buffer.html#bufwritedoublebevalue-offset)
- [buf.writeDoubleLE(value[, offset])](https://nodejs.org/api/buffer.html#bufwritedoublelevalue-offset)
- [buf.writeFloatBE(value[, offset])](https://nodejs.org/api/buffer.html#bufwritefloatbevalue-offset)
- [buf.writeFloatLE(value[, offset])](https://nodejs.org/api/buffer.html#bufwritefloatlevalue-offset)
- [buf.writeInt8(value[, offset])](https://nodejs.org/api/buffer.html#bufwriteint8value-offset)
- [buf.writeInt16BE(value[, offset])](https://nodejs.org/api/buffer.html#bufwriteint16bevalue-offset)
- [buf.writeInt16LE(value[, offset])](https://nodejs.org/api/buffer.html#bufwriteint16levalue-offset)
- [buf.writeInt32LE(value[, offset])](https://nodejs.org/api/buffer.html#bufwriteint32levalue-offset)
- [buf.writeUInt8(value[, offset])](https://nodejs.org/api/buffer.html#bufwriteuint8value-offset)
- [buf.writeUInt16BE(value[, offset])](https://nodejs.org/api/buffer.html#bufwriteuint16bevalue-offset)
- [buf.writeUInt16LE(value[, offset])](https://nodejs.org/api/buffer.html#bufwriteuint16levalue-offset)
- [buf.writeUInt32BE(value[, offset])](https://nodejs.org/api/buffer.html#bufwriteuint32bevalue-offset)
- [buf.writeUInt32LE(value[, offset])](https://nodejs.org/api/buffer.html#bufwriteuint32levalue-offset)

### Checklist

### DIFF
--- a/modules/llrt_buffer/src/buffer.rs
+++ b/modules/llrt_buffer/src/buffer.rs
@@ -479,7 +479,7 @@ pub enum Endian {
 }
 
 #[allow(clippy::too_many_arguments)]
-fn write_value<'js>(
+fn write_buf<'js>(
     this: &This<Object<'js>>,
     ctx: &Ctx<'js>,
     value: &Value<'js>,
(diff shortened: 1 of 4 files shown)


modules/llrt_buffer/src/buffer.rs
```diff
@@ -308,17 +288,256 @@ fn from<'js>(
...
+        let target_bytes = unsafe { slice::from_raw_parts_mut(raw.ptr.as_ptr(), raw.len) };
+
+        writable_length = offset + source_slice.len();
+
+        target_bytes[offset..writable_length].copy_from_slice(&source_slice);
+    }
+
+    Ok(writable_length)
+}
+
```
richarddavison, line 524: Can we make this a bit more generic to support all write methods?

modules/llrt_buffer/src/buffer.rs
```diff
@@ -220,6 +133,73 @@ fn alloc_byte_ref<'js>(ctx: &Ctx<'js>, byte_ref: &[u8], length: usize) -> Result
...
+fn maybeuninit_to_u8(vec: Vec<MaybeUninit<u8>>) -> Vec<u8> {
+    let len = vec.len();
+    let capacity = vec.capacity();
+    let ptr = vec.as_ptr() as *mut u8;
+
+    std::mem::forget(vec);
+
+    unsafe { Vec::from_raw_parts(ptr, len, capacity) }
+}
```
richarddavison, line 153: Can you elaborate what this does. Please add a safety comment

richarddavison (changes requested):
Thanks for the PR! Some comments!

modules/llrt_buffer/src/buffer.rs
```diff
@@ -308,17 +288,256 @@ fn from<'js>(
...
+        },
+
+        (Some(v1), None, None) => {
+            let offset = v1.as_int().unwrap_or(0) as usize;
+            Ok((offset, len - offset, encoding))
+        },
+
+        _ => Ok((0, len, encoding)),
+    }
+}
```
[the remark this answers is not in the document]
nabetti1720: I looked into the implementation of child_process but didn't find anything useful.

What makes parameter analysis difficult this time is that different types of encodings can appear for the first, second and third arguments.

From top to bottom, I expect the following patterns, which makes branching complicated:

- encoding
- offset, encoding
- offset, length, encoding
- offset, length
- offset
- none

It might be possible to implement this more trickily using if statements, but do you have any good ideas for a cleaner implementation?

modules/llrt_buffer/src/buffer.rs
nabetti1720: Would comments like this help you understand?

> This conversion is safe because MaybeUninit<u8> has the same memory layout as u8, meaning the underlying bytes are identical. Since Vec<MaybeUninit<u8>> and Vec<u8> share the same memory representation, a simple reinterpretation of the pointer is valid.
> Additionally, Vec::from_raw_parts correctly reconstructs the vector using the original length and capacity, ensuring that memory ownership remains consistent. The call to std::mem::forget(vec) prevents the original Vec<MaybeUninit<u8>> from being dropped, avoiding double frees or memory corruption.
> However, this conversion is only safe if all elements of MaybeUninit<u8> are properly initialized. If any uninitialized values exist, reading them as u8 would lead to undefined behavior.

modules/llrt_buffer/src/buffer.rs
nabetti1720: I think I can do it. However, I won't have much time for a few days from today, so please be patient.

modules/llrt_buffer/src/buffer.rs
```diff
@@ -478,50 +478,99 @@ fn safe_byte_slice(s: &str, end: usize) -> (&[u8], usize) {
...
+
+fn write_int32_le<'js>(
+    this: This<Object<'js>>,
+    ctx: Ctx<'js>,
+    value: i32,
+    offset: Opt<usize>,
+) -> Result<usize> {
+    write_int(&this, &ctx, value, &offset, Endian::Little, 4)
+}
+
```
richarddavison, line 579: Great work. However, I think we can simplify by not using generics and cast with range check instead. Something like this should be able to work with all wirteXBE/LE in buffer. This means thet we can also use closures instead of dedicated fn for the API
```rust
fn write_value<'js>(
    this: &This<Object<'js>>,
    ctx: &Ctx<'js>,
    value: Value<'js>,
    offset: &Opt<usize>,
    endian: Endian,
    bits: u8,
    signed: bool,
    is_float: bool,
    is_bigint: bool,
) -> Result<usize> {
    // Strict type validation
    match (is_bigint, is_float) {
        (true, _) if !value.is_bigint() => return Err(Exception::throw_type(ctx, "Expected BigInt")),
        (_, true) if !value.is_number() => return Err(Exception::throw_type(ctx, "Expected number")),
        _ => (),
    }

    // Extract and convert value
    let (byte_count, bytes) = if is_bigint {
        let bigint = value.as_big_int().unwrap();
        let (val, byte_count) = match (signed, bits) {
            (true, 64) => (bigint.to_i64().ok_or_else(|| 
                Exception::throw_range(ctx, "BigInt out of range"))? as u64, 8),
            (false, 64) => (bigint.to_u64().ok_or_else(|| 
                Exception::throw_type(ctx, "Positive BigInt required"))?, 8),
            _ => return Err(Exception::throw_range(ctx, "Invalid BigInt size")),
        };
        (byte_count, endian_bytes(val, endian))
    } else if is_float {
        let float_val = value.as_float().unwrap();
        match bits {
            32 => {
                let f32_val = float_val as f32;
                (4, f32_val.to_bits().to_be_bytes().to_vec())
            },
            64 => (8, float_val.to_bits().to_be_bytes().to_vec()),
            _ => return Err(Exception::throw_range(ctx, "Invalid float size")),
        }
    } else {
        let int_val = value.as_int().unwrap() as i64;
        let bit_mask = (1i64 << bits) - 1;
        let max_val = if signed { (1i64 << (bits - 1)) - 1 } else { bit_mask };
        let min_val = if signed { -max_val - 1 } else { 0 };

        if int_val < min_val || int_val > max_val {
            return Err(Exception::throw_range(ctx, "Value out of range"));
        }

        let masked = int_val & bit_mask;
        ((bits / 8) as usize, shifted_bytes(masked as u64, bits, endian))
    };

     let target = ObjectBytes::from(ctx, this.0.as_inner())?;
    let mut writable_length = 0;

    if let Some((array_buffer, _, _)) = target.get_array_buffer()? {
        let raw = array_buffer
            .as_raw()
            .ok_or(ERROR_MSG_ARRAY_BUFFER_DETACHED)
            .or_throw(ctx)?;

        let target_bytes = unsafe { slice::from_raw_parts_mut(raw.ptr.as_ptr(), raw.len) };

        writable_length = offset + bytes.len();
        target_bytes[offset..writable_length].copy_from_slice(& bytes);
    }

    Ok(writable_length)
}
    
   

// Pure mathematical byte generation
fn endian_bytes(mut val: u64, endian: Endian) -> Vec<u8> {
    let mut bytes = vec![0u8; 8];
    for i in 0..8 {
        bytes[i] = match endian {
            Endian::Big => (val >> (56 - i * 8)) as u8,
            Endian::Little => (val >> (i * 8)) as u8,
        };
        val &= !(0xFF << ((7 - i) * 8));  // Clear processed bits
    }
    bytes
}

fn shifted_bytes(mut val: u64, bits: u8, endian: Endian) -> Vec<u8> {
    let byte_count = (bits / 8) as usize;
    let mut bytes = vec![0u8; byte_count];
    
    for i in 0..byte_count {
        let shift = match endian {
            Endian::Big => (byte_count - 1 - i) * 8,
            Endian::Little => i * 8,
        };
        bytes[i] = (val >> shift) as u8;
        val &= !(0xFF << shift);  // Clear processed bits
    }
    bytes
}
```

richarddavison (changes requested):
Some suggestions

modules/llrt_buffer/src/buffer.rs
nabetti1720: Thank you. I would like to refactor this code, but I have some issues.

- RquickJS does not support BigInt very well. For example, `Value.isBigint()` and `Bigint.to_u64()` do not seem to be supported. I think you can write `Value.isBigint()` as `Value.as_big_ing().is_some()`. However, I don't think `Bigint.to_u64()` can be replaced.

- I would like to write it in closure format and eliminate individual fn, but I don't know how to receive the parameter equivalent to `this`. For example, if you write it like this, it won't know what to pass to `this`.

```rust
    prototype.set(
        "writeInt8",
        Func::from(|c: Ctx<'js>, v: Value<'js>, o: Opt<usize>| {
            write_value(&this, &c, &v, &o, Endian::Little, 8, true, false, false)
        }),
    )?;
 ```

modules/llrt_buffer/src/buffer.rs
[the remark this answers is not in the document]
nabetti1720: I simplified it by using `if` statements instead of `match`.

richarddavison (approved):
Thanks! LGTM!

PR #892, opened 2025-03-17
refactor(console): separate the default and Lambda consoles

### LINKED ISSUE #351: Splitting Lambda specific parts out of llrt_core
After splitting LLRT up in 2 parts `llrt` and `llrt_core` the next step would be to look for a way to split lambda logic out of LLRT to its own thing.

Options would be:
- `llrt`, `llrt_core` and `llrt_lambda` or something like it;
- `llrt` (with a feature flag lambda) and `llrt_core`;

The AWS libraries would also be stripped out of `llrt_core`. 
@richarddavison gave the solution for that [here](https://github.com/awslabs/llrt/pull/326#issuecomment-2079780245).
The `ConsoleModule` will require the most effort to split in two parts, in issue #341 it was stated that this module could be rewritten. @richarddavison would you happen to know when this is planned?

### LINKED ISSUE #885: The print function behaves inconsistently with console.log
```
// TypeError: Error converting from js 'int' into type 'string'
// print(1)


// TypeError: Error converting from js 'string' into type 'bool'
print('1', '2')
```

nabetti1720:
### Issue # (if available)

Closed #351
Closed #885 

### Description of changes

The previously closely intertwined console.rs has been separated as follows:
- libs/llrt_logging: Manages processes that can be used in common by the default console and the Lambda console
- modules/llrt_console: Default console implementation
- core/console: Lambda console implementation
- llrt:util: A collection of useful functions used only by llrt_core and its surroundings

I have also confirmed that I can build modules for lambda on my laptop.

Once this PR is complete, we will be able to extract more `util` modules into crates (`format()` was blocking it).

~There is still some work to do, such as addressing `Consolable` trait, but I would like to release it as a draft.~
It doesn't seem like traits are necessarily necessary, so I'll pass on them this time.

### Checklist


modules/llrt_console/Cargo.toml
```diff
@@ -0,0 +1,18 @@
+[package]
+name = "llrt_console"
+description = "LLRT Module console"
+version = "0.5.1-beta"
+edition = "2021"
+license = "Apache-2.0"
+repository = "https://github.com/awslabs/llrt"
+
```
Sytten, line 8: Lets add a feature for the output.
Stdout is a good default but log / tracing should be another.

llrt_core/Cargo.toml
```diff
@@ -13,23 +13,51 @@ macro = ["rquickjs/macro"]
...
+rustls = { version = "0.23", default-features = false, features = [
+  "tls12",
+  "ring",
+] }
+rustls-pemfile = "2"
+ryu = "1"
+simd-json = { version = "0.14", default-features = false, features = [
+  "big-int-as-float",
+] }
+terminal_size = "0.4"
```
Sytten, line 57: Check if we still need it in core

modules/llrt_console/src/lib.rs
```diff
@@ -0,0 +1,163 @@
...
+    console.set("log", Func::from(log))?;
+    console.set("clear", Func::from(clear))?;
+    console.set("debug", Func::from(log_debug))?;
+    console.set("info", Func::from(log))?;
+    console.set("trace", Func::from(log_trace))?;
+    console.set("error", Func::from(log_error))?;
+    console.set("warn", Func::from(log_warn))?;
+    console.set("assert", Func::from(log_assert))?;
+    console.prop("__dimensions", Accessor::from(get_dimensions))?;
+    console.set("__format", Func::from(|ctx, args| format(&ctx, true, args)))?;
```
Sytten, line 158: I would consider if we still need those private functions since they are not standard. That would also remove the terminal size dependency.

Sytten (changes requested):

modules/llrt_console/src/lib.rs
nabetti1720: `__format` is not used anywhere, so I think it can be deleted.
On the other hand, `__dimensions` is used in the test suite, so it cannot be deleted, but I think it can be separated into `llrt:test`.

modules/llrt_console/Cargo.toml
nabetti1720: Sorry, I haven't caught up with your thoughts yet. What exactly do you think should be done with console.log/trace? Is there any information in the Node.js documentation?

[console.log([data][, ...args])](https://nodejs.org/api/console.html#consolelogdata-args)
[console.trace([message][, ...args])](https://nodejs.org/api/console.html#consoletracemessage-args)

llrt_core/Cargo.toml
nabetti1720: The get_dimensions function that uses this crate has been moved to llrt:util, so it is still needed.

nabetti1720:
@richarddavison @Sytten Is there anything else you need to work on in this PR? If not, you're ready to go :)

richarddavison (approved):
This is great. LGTM!

PR #900, opened 2025-03-21
Make FormatOptions immutable

Sytten:
### Description of changes

This makes the FormatOptions immutable, allowing us to store it if we want and re-use it.
I also think it makes it clearer what is an override versus the user intention.
Let me know if you like the pattern, I am not sure.

### Checklist

### DIFF
--- a/libs/llrt_logging/src/lib.rs
+++ b/libs/llrt_logging/src/lib.rs
@@ -125,6 +125,7 @@ impl LogLevel {
     }
 }
 
+#[derive(Debug, Clone)]
 pub struct FormatOptions<'js> {
     color: bool,
     newline: bool,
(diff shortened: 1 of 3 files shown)


nabetti1720:
This is just my personal opinion without any consideration of performance or memory usage. I don't really like the approach of passing `options` and `options_override` between multiple functions. Would it be okay to clone `options` (and change their value) only if a change is needed?

richarddavison:
@Sytten @nabetti1720 i agree. It’s a bit clunky to pass both
> I think this approach is effective if you only need to make the parameters of functions exposed to the outside world immutable.

richarddavison:
> The goal was to avoid the clone for each format essentially, otherwise I would probably just pass it as an owned value instead. @richarddavison

Can you refactor to owned value instead? If not possible, Rc is also fine

Sytten:
Yeah I will do owned value!

Sytten:
Sorry I am dropping the ball on this, I will try to do it soon

Sytten:
Closing for now, I dont have time to prioritize that work

PR #908, opened 2025-03-24
refactor(core): Clarifying responsibilities of modules and libs

nabetti1720:
### Description of changes

We are making some refactorings to the core to improve future maintainability.

- Consolidated module_loader assets into the ~`llrt_core/src/modules/module`~`llrt_core/src/modules/require` directory in preparation for extracting it into the ~`llrt_module`~`llrt_require` crate in the future.
- We've made the libs directory accessible to crates via the identifiers exposed in `llrt_core/src/libs.rs` (this is equivalent to the current `llrt_modules` meta-library). ~In the future, we will split out `llrt_libs` into a meta-library and make each library available as an opt-in.~
- ~`llrt_build` and `llrt_test`, which were in the libs library, have been moved to the `helpers` directory. These are not strictly libraries but helper methods.~
- We created `llrt/src/core.rs` in the llrt (execution module) source. By consolidating the public information about `llrt_core` here, I made it clear that all you need to create llrt (the execution module) is `llrt_core`.
- Other minor adjustments have been made.

### Checklist


nabetti1720:
@richarddavison That's all for this PR.

Sytten:
I am not sure I get the distinctions between libraries and helpers, libraries are not meant to be used by end users directly so I don't think we should group them in `llrt_libs`. I am not even sure we need `llrt_modules` anymore TBH. I think `llrt_require` might be a better name than `llrt_module`.

nabetti1720:
@Sytten Thank you for your valuable comments. :)  I will sort out my thoughts one by one.

> I am not sure I get the distinctions between libraries and helpers, libraries are not meant to be used by end users directly so I don't think we should group them in `llrt_libs`.

The difference between libraries and helpers here is what will be included in the future release of llrt_libs, and what will not. Another difference is the positional relationship when considering the description in Cargo.toml.

For example, I was hoping that llrt_core/Cargo.toml could be written as follows:
```toml
[dependencies]
llrt_libs = { path = "../llrt_libs", features = ["context", "encoding", "json", "logging", "numbers", "utils"] }
llrt_modules = { path = "../llrt_modules" }
any_other_crates = {}
[build-dependencies]
llrt_build = { path = "../helpers/llrt_build" }
any_other_crates = {}
[dev-dependencies]
llrt_test = { path = "../helpers/llrt_test" }
any_other_crates = {}
```

> I am not even sure we need `llrt_modules` anymore TBH.

Regarding whether llrt_modules is necessary, it is true that LLRT bundles all modules, so it may not be necessary, but I was thinking that it might be necessary in projects other than LLRT.

> I think `llrt_require` might be a better name than `llrt_module`.

`llrt_core/src/modules/module` has a function that exposes `createRequire`, which in node.js is included in `node:module`, so I believe that LLRT should also expose it as a `llrt_module` crate.
https://nodejs.org/api/module.html#modulecreaterequirefilename

However, it seems that there is no problem with extracting the implementation that previously existed in the `module_loader` directory as `llrt_require`, is this what you are referring to?

nabetti1720:
Should we deprecate llrt_modules?: It would be too big to tackle in this PR, as we would need to prepare READMEs for each crate.
Should we release llrt_libs?: Regardless of the conclusion, I don't think it's an issue for this PR.

Also, I think this PR has prepared the groundwork for releasing llrt_require.

@richarddavison @Sytten Are there any other improvements that should be made? If not, then you can go ahead and proceed as we have achieved everything we wanted to achieve with this PR.

Sytten:
My take is we dont need to do an llrt_libs module so we should not split helpers from libs. The rest is good and I agree we should put require in a module.

nabetti1720:
> My take is we dont need to do an llrt_libs module so we should not split helpers from libs. The rest is good and I agree we should put require in a module.

@Sytten Thank you for your comment. I've reverted the excessive changes.

richarddavison (approved):
Thanks. This looks better!

Issue #911, opened 2025-03-25
terminated by signal SIGILL (Illegal instruction): Multi-byte utf8 string calls slice function


ahaoboy:
```
var s = "🌍🌎🌏"
console.log(s.slice(1))
```

```
 qjs a.js
��🌎🌏


 llrt ./a.js
fish: Job 1, 'llrt ./a.js' terminated by signal SIGILL (Illegal instruction)
```

PR #928, opened 2025-04-03
fix: Print utf8 replacement char for broken surrogate pairs

richarddavison:
### Issue # (if available)

Fixes https://github.com/awslabs/llrt/issues/911

### Description of changes

Fixes broken surrogate pairs. Thanks for reporting @ahaoboy 

### Checklist

### DIFF
--- a/libs/llrt_logging/src/lib.rs
+++ b/libs/llrt_logging/src/lib.rs
@@ -3,7 +3,9 @@
 use std::{
     collections::HashSet,
     io::{stdout, IsTerminal},
+    mem,
     ops::Deref,
+    slice,
 };
 
 use llrt_json::stringify::json_stringify;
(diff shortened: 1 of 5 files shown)


Sytten:
@richarddavison You can use the CString primitive maybe? Also see my comment about using the new UTF-16 API, at this point if you are doing UTF-8 decoding it would probably be better anyway.

richarddavison:
> @richarddavison You can use the CString primitive maybe? Also see my comment about using the new UTF-16 API, at this point if you are doing UTF-8 decoding it would probably be better anyway. Also I would consider this a bug in quickjs-ng to be honest since the high surrogate should be transformed by the spec. As for the buffer we should really really stop using the UTF-8 conversion as right now we have a bug compared to the node buffer when doing utf-16e decoding on a string with high surrogate, you should not get the ? characters but we do because of the loosy conversion.

Agree, there are better ways to do this. But for now I will merge to fix the panics 👌

PR #936, opened 2025-04-09
refactor(require/module): further separation of require and module

nabetti1720:
### Description of changes

We've refactored `require` and `module` to be more independent.

- Previously, the module list output by `ModuleBuilder` was passed directly to `Module::init()`, but now it is passed via global (although I'm not sure if this is a good way to do it).
- This allowed `Module::init()` to have the same specifications as other `init()`, so the attachment method could also be made consistent.
- The `module` contained the implementation of `require`, but we were able to separate the implementation part.

This makes it more likely that we can move `module` to the module directory and `require` to the libs directory. However, in order to make it an independent crate, there are some things to consider about the implementation of `require` (what to do with LLRT and Lambda-specific logic), so we will not implement it in this PR.

NOTE:
We already have an idea for this issue. It's to split `CustomResolver/Loader` into `NpmJsResolver/Loader` and `EmbeddedResolver/Loader`. `NpmJsResolver/Loader` will be exposed, but `EmbeddedResolver/Loader` will continue to be under the management of core.

However, some integration will be required in `require`. But it seems that we can deal with it by accepting one user-defined function in `require_resolver()` and making ByteCodeCache judgment.

### Checklist

### DIFF
--- a/llrt_core/src/modules/module.rs
+++ b/llrt_core/src/modules/module.rs
@@ -10,7 +10,10 @@ use rquickjs::{
 };
 
 use crate::libs::utils::module::{export_default, ModuleInfo};
-use crate::modules::require::{require, ModuleNames, RequireState, CJS_IMPORT_PREFIX};
+use crate::modules::{
+    require::{require, RequireState, CJS_IMPORT_PREFIX},
+    ModuleNames,
+};
 use crate::utils::ctx::CtxExt;
 
 pub struct ModuleModule;
@@ -21,9 +24,9 @@ fn create_require(ctx: Ctx<'_>) -> Result<Value<'_>> {
 
 fn is_builtin(ctx: Ctx<'_>, name: String) -> bool {
     let binding = ctx.userdata::<RefCell<ModuleNames>>().unwrap();
-    let module_names = binding.borrow();
+    let module_list = binding.borrow().get_list();
 
-    module_names.list.contains(&name)
+    module_list.contains(&name)
 }
 
 impl ModuleDef for ModuleModule {
@@ -39,9 +42,9 @@ impl ModuleDef for ModuleModule {
     fn evaluate<'js>(ctx: &Ctx<'js>, exports: &Exports<'js>) -> Result<()> {
         export_default(ctx, exports, |default| {
             let binding = ctx.userdata::<RefCell<ModuleNames>>().unwrap();
-            let module_names = binding.borrow();
+            let module_list = binding.borrow().get_list();
 
-            default.set("builtinModules", module_names.list.clone())?;
+            default.set("builtinModules", module_list)?;
             default.set("createRequire", Func::from(create_require))?;
             default.set("isBuiltin", Func::from(is_builtin))?;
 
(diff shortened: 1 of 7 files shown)


Sytten:
AFAIK this is all unnecessary the moment we move to https://github.com/rquickjs/rquickjs-extension and I think we should prioritize that instead.

nabetti1720:
@richarddavison @Sytten In the next PR, I will try to change it to an implementation using rquickjs-extension. :)

Sytten:
Make sense, the code changes make sense to me

nabetti1720:
@richarddavison  Since you've rebased onto the latest commit, this should be fine, and you're ready to merge if nothing else bothers you. :)

Sytten:
We should not store that statically since static will be shared between multiple VM running in the same process. Instead you should attached it to VM with user data if possible.

nabetti1720:
> We should not store that statically since static will be shared between multiple VM running in the same process. Instead you should attached it to VM with user data if possible.

Are you referring to the following static variables?

```rust
static MODULE_NAMES: OnceCell<HashSet<String>> = OnceCell::new();
```

Sytten:
Yes

llrt_core/src/modules/module.rs
```diff
@@ -1,62 +1,52 @@
...
-}
-
 pub struct ModuleModule;
 
 fn create_require(ctx: Ctx<'_>) -> Result<Value<'_>> {
     ctx.globals().get("require")
 }
 
+fn is_builtin(ctx: Ctx<'_>, name: String) -> bool {
+    let binding = ctx.userdata::<RefCell<ModuleNames>>().unwrap();
```
richarddavison, line 26: This seems a bit fragile. If implementors doesn't use our module_builder this would panic. An expect is better in this case

llrt_core/src/modules/require/mod.rs
```diff
@@ -24,3 +36,164 @@ pub static LLRT_PLATFORM: Lazy<String> = Lazy::new(|| {
...
+
+unsafe impl<'js> JsLifetime<'js> for RequireState<'js> {
+    type Changed<'to> = RequireState<'to>;
+}
+
+pub fn require(ctx: Ctx<'_>, specifier: String) -> Result<Value<'_>> {
+    struct Args<'js>(Ctx<'js>);
+    let Args(ctx) = Args(ctx);
+
+    let binding = ctx.userdata::<RefCell<ModuleNames>>().unwrap();
```
richarddavison, line 55: Why do we need a RefCell? We're not mutating the ModuleList

richarddavison (changes requested):
Some minor comments

llrt_core/src/modules/module.rs
nabetti1720: Is there a way to find out the name of a custom bound module without using the `module_builder`? To be honest, I couldn't think of a good solution, so I implemented a workaround to return an error saying "not supported" for existence checks like isBuiltin, and an empty list if you want to get a list of modules.

richarddavison (approved):
Thanks, this is better organized now!

llrt_core/src/modules/module.rs
richarddavison: This is fine for now. We'll likely be fixed by the new module builder once that's get in :)

PR #948, opened 2025-04-19
refactor(require): further separation of npm-require and bytecode-require

nabetti1720:
### Issue # (if available)

Related #936 

### Description of changes

> NOTE: We already have an idea for this issue. It's to split `CustomResolver/Loader` into `NpmJsResolver/Loader` and `EmbeddedResolver/Loader`. `NpmJsResolver/Loader` will be exposed, but `EmbeddedResolver/Loader` will continue to be under the management of core.
> 
> However, some integration will be required in `require`. But it seems that we can deal with it by accepting one user-defined function in `require_resolver()` and making ByteCodeCache judgment.

As mentioned in #936, we have extracted the LLRT-specific bytecode from the previous require. In the future, we plan to expose the require directory as llrt_require, and leave the embedded directory under the management of llrt_core.

However, there are still some tangled parts (src/bytecode.rs, src/modules/utils/*) and we cannot expose require as it is. I need some more time to sort this out, so I will hold off on this for now.

### Checklist

### DIFF
--- a/llrt_core/src/compiler.rs
+++ b/llrt_core/src/compiler.rs
@@ -9,7 +9,7 @@ use zstd::bulk::Compressor;
 use crate::bytecode::add_bytecode_header;
 use crate::compiler_common::{human_file_size, DummyLoader, DummyResolver};
 use crate::libs::{logging::print_error_and_exit, utils::result::ResultExt};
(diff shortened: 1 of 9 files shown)


richarddavison (approved):
Thank you! Looks good to me!

PR #949, opened 2025-04-20
feat: Add execFile Support

### LINKED ISSUE #818: child_process: support execFile
A function very similar to spawn

https://nodejs.org/api/child_process.html#child_processexecfilefile-args-options-callback

llrt's child_process module does not have a corresponding sync function, it would be nice if it could be supported

nithinkjoy-tech:
### Issue
Closes #818 

### Description of changes

Added execFile support to the child_process module.

### Checklist

- [X] Updated documentation if needed ([API.md](https://github.com/awslabs/llrt/pull/API.md)/[README.md](https://github.com/awslabs/llrt/pull/README.md)/Other)


modules/llrt_child_process/src/lib.rs
```diff
@@ -418,44 +625,204 @@ fn spawn<'js>(
...
+    {
+        command.process_group(0);
+    }
+
+    //tokio command does not have all std command features stabilized
+    let mut command = Command::from(command);
+    ChildProcess::spawn(ctx.clone(), cmd, command_args, command.spawn())
+}
+
+fn exec_file<'js>(
```
Sytten, line 682: The refactor of `spawn` to accommodate `exec_file` is good. Let's split the helpers in another file (I would split functions together, the class on its own and the helpers together).

modules/llrt_child_process/src/lib.rs
```diff
@@ -418,44 +625,204 @@ fn spawn<'js>(
...
+    }
+
+    //tokio command does not have all std command features stabilized
+    let mut command = Command::from(command);
+    ChildProcess::exec_file(ctx.clone(), cmd, command_args, command.spawn(), cb)
+}
+
+fn get_callback_fn<'js>(
+    ctx: &Ctx<'js>,
+    args: Vec<Option<&Value<'js>>>,
```
Sytten, line 756: we should be able to avoid a vec allocation here I am pretty sure `&[Option<&Value<'js>>]` should work

modules/llrt_child_process/src/lib.rs
```diff
@@ -418,44 +625,204 @@ fn spawn<'js>(
...
+
+    //tokio command does not have all std command features stabilized
+    let mut command = Command::from(command);
+    ChildProcess::spawn(ctx.clone(), cmd, command_args, command.spawn())
+}
+
+fn exec_file<'js>(
+    ctx: Ctx<'js>,
+    cmd: String,
+    args_and_opts: Rest<Value<'js>>,
```
Sytten, line 685: I don't like this style of arguments, this execFile and spawn don't have variadic arguments.
I know this is the style of @richarddavison :D
At minimum you can write:
```rust
fn exec_file<'js>(
    ctx: Ctx<'js>,
    cmd: String,
    args_0: Opt<Value<'js>>,
    args_1: Opt<Value<'js>>,
    args_2: Opt<Value<'js>>,
)
```

Even better would be to start using proper typing directly in the arguments (this is approximate code):
```rust
fn exec_file<'js>(
    ctx: Ctx<'js>,
    cmd: String,
    args_0: Opt<Either<Either<Array<'js>, Object<'js>, Function<'js>>>>, // I wouldnt againt be an Either3 implementation
    args_1: Opt<Either<Object<'js>, Function<'js>>,
    args_2: Opt<Function<'js>>,
)
```

Opt is different from Option since it doesnt require undefined to be passed to the function

modules/llrt_stream/src/readable.rs
```diff
@@ -280,17 +289,27 @@ where
...
+                                            if listener == "data" {
+                                                Self::emit_str(
+                                                    This(this2.clone()),
+                                                    &ctx3,
+                                                    "data",
+                                                    vec![Buffer(buffer.clone()).into_js(&ctx3)?],
+                                                    false
+                                                )?;
+                                            }
+                                        }
```
Sytten, line 309: This is a change in behaviour for which I not qualified to comment, @richarddavison

modules/llrt_stream/src/readable.rs
```diff
@@ -223,14 +225,16 @@ where
...
+        Self::do_process(this, ctx, readable, on_end, None, None)
     }
 
     fn do_process<T: AsyncRead + 'js + Unpin, C: FnOnce() + Sized + 'js>(
         this: Class<'js, Self>,
         ctx: &Ctx<'js>,
         readable: T,
         on_end: C,
+        combined_std_buffer: Option<Arc<Mutex<Vec<u8>>>>,
+        cb: Option<Function<'js>>,
```
Sytten, line 237: Is there a case where we would want to use a combined_std_buffer independently from cb? All usages in this PR assume that cb and combined_std_buffer are either some or none.
I am not really a want of adding it to the stream, I wonder if we could instead either pipe the stdout/err through an accumulation middleware before forwarding it to the DefaultReadableStream.

modules/llrt_child_process/src/lib.rs
```diff
@@ -351,6 +344,216 @@ impl<'js> ChildProcess<'js> {
...
+                if let Some(child_stdin) = child.stdin.take() {
+                    DefaultWritableStream::process(stdin_instance.clone(), &ctx, child_stdin)?;
+                };
+
+                let stdout_new: Vec<u8> = Vec::new();
+                let stderr_new: Vec<u8> = Vec::new();
+                let stdout_arc = Arc::new(Mutex::new(stdout_new));
+                let stderr_arc = Arc::new(Mutex::new(stderr_new));
+                let combined_stdout_buffer = Some(Arc::clone(&stdout_arc));
+                let combined_stderr_buffer = Some(Arc::clone(&stderr_arc));
```
Sytten, line 391: As far as I can see, we don't use None on this value anywhere.

Also depending on the usecase, it is usually fine to use a `RefCell` since the JS engine is single-threaded. As long as you don't hold the borrow across an await point or a callback to JS (anything that yields control back to the JS engine really).

Sytten (changes requested):
First pass on it

Sytten:
For reference, the node implementation uses spawn under the cover: https://github.com/nodejs/node/blob/1720b18260d7f4bbd9fe0d2cbbbb05cc33e7f945/lib/child_process.js#L323 and they do in fact set a listener on `'data'` to accumulate their callback value. We should go that route too if possible.

richarddavison (changes requested):
Thanks for the PR! @Sytten covered most points. Mainly we should build `exec` and `execFile` upon `spawn`

modules/llrt_child_process/src/lib.rs
richarddavison: Agree. We can do Either to clean it up. Similar approach can be added in `spawn`

nithinkjoy-tech:
Thank you @Sytten and @richarddavison for the review. I will refactor it to build `exec` and `execFile` upon `spawn` and will try to set a listener on 'data' to accumulate their callback value.

Sytten:
@nithinkjoy-tech Do you plan on finishing the PR?

nithinkjoy-tech:
> @nithinkjoy-tech Do you plan on finishing the PR?
Hi @Sytten,  Thanks for following up! I don’t think I’ll be able to continue with this PR since I got stuck earlier and haven’t had the time to get back to it. Totally fine if someone else wants to take it forward.

Sytten:
@richarddavison Do you want to take that on or not?

richarddavison:
Replaced by https://github.com/awslabs/llrt/pull/1708

PR #971, opened 2025-05-04
fix(timers): not completing all microtasks before completing a macrotask

nabetti1720:
### Description of changes

This code verifies that all microtasks have been processed before processing a macrotask. If it works correctly, the output will start in alphabetical order.

<details>
  <summary> realitySettimeout-nested.js</summary>

```javascript
// realitySettimeout-nested.js
console.log("[A] 🦖 MAINLINE: Start");

setTimeout(() => {
  console.log("[D] ⏰ TIMERS: setTimeout[0ms]");

  Promise.resolve("1st Promise")
    .then((value) => console.log("[E] 👦 MICRO: Resolved value:", value))
    .then(() => console.log("[F] 👦 MICRO: Next chain"));

  setTimeout(() => {
    console.log("[J] ⏰ TIMERS: setTimeout[0ms]");

    Promise.resolve("2nd Promise")
      .then((value) => console.log("[K] 👦 MICRO: Resolved value:", value))
      .then(() => console.log("[L] 👦 MICRO: Next chain"));
    });
  });

setTimeout(() => {
  console.log("[G] ⏰ TIMERS: setTimeout[0ms]");

  Promise.resolve("3rd Promise")
    .then((value) => console.log("[H] 👦 MICRO: Resolved value:", value))
    .then(() => console.log("[I] 👦 MICRO: Next chain"));
  });

Promise.resolve().then(() => console.log("[C] 👦 MICRO: then"));

console.log("[B] 🦖 MAINLINE: End");
```

</details>

Node.js (The results for Deno and Bun are similar):
```
% node realitySettimeout-nested.cjs
[A] 🦖 MAINLINE: Start
[B] 🦖 MAINLINE: End
[C] 👦 MICRO: then
[D] ⏰ TIMERS: setTimeout[0ms]
[E] 👦 MICRO: Resolved value: 1st Promise
[F] 👦 MICRO: Next chain
[G] ⏰ TIMERS: setTimeout[0ms]
[H] 👦 MICRO: Resolved value: 3rd Promise
[I] 👦 MICRO: Next chain
[J] ⏰ TIMERS: setTimeout[0ms]
[K] 👦 MICRO: Resolved value: 2nd Promise
[L] 👦 MICRO: Next chain
```

LLRT:
```
 % llrt realitySettimeout-nested.cjs
[A] 🦖 MAINLINE: Start
[B] 🦖 MAINLINE: End
[D] ⏰ TIMERS: setTimeout[0ms]
[G] ⏰ TIMERS: setTimeout[0ms]
[C] 👦 MICRO: then
[E] 👦 MICRO: Resolved value: 1st Promise
[H] 👦 MICRO: Resolved value: 3rd Promise
[F] 👦 MICRO: Next chain
[I] 👦 MICRO: Next chain
[J] ⏰ TIMERS: setTimeout[0ms]
[K] 👦 MICRO: Resolved value: 2nd Promise
[L] 👦 MICRO: Next chain
```

This PR checks that all microtasks have been processed before executing a macrotask (a timer task in LLRT).

LLRT(After correction):
```
% llrt realitySettimeout-nested.cjs
[A] 🦖 MAINLINE: Start
[B] 🦖 MAINLINE: End
[C] 👦 MICRO: then
[D] ⏰ TIMERS: setTimeout[0ms]
[E] 👦 MICRO: Resolved value: 1st Promise
[F] 👦 MICRO: Next chain
[G] ⏰ TIMERS: setTimeout[0ms]
[H] 👦 MICRO: Resolved value: 3rd Promise
[I] 👦 MICRO: Next chain
[J] ⏰ TIMERS: setTimeout[0ms]
[K] 👦 MICRO: Resolved value: 2nd Promise
[L] 👦 MICRO: Next chain
```

### Checklist

### DIFF
--- a/modules/llrt_timers/src/lib.rs
+++ b/modules/llrt_timers/src/lib.rs
@@ -312,15 +312,21 @@ pub fn poll_timers(
 
     drop(rt_timers);
 
+    let mut is_first_time = true;
     for item in call_vec.iter_mut() {
         if let Some(ExecutingTimer(ctx, timeout)) = item.take() {
             let ctx2 = unsafe { Ctx::from_raw(ctx) };
 
-            while ctx2.execute_pending_job() {}
+            if is_first_time {
+                while ctx2.execute_pending_job() {}
+                is_first_time = false;
+            }
 
             if let Ok(timeout) = timeout.restore(&ctx2) {
                 timeout.call::<_, ()>(())?;
             }
+
+            while ctx2.execute_pending_job() {}
         }
     }
     call_vec.clear();
--- a/modules/llrt_timers/src/lib.rs
+++ b/modules/llrt_timers/src/lib.rs
@@ -315,6 +315,9 @@ pub fn poll_timers(
     for item in call_vec.iter_mut() {
         if let Some(ExecutingTimer(ctx, timeout)) = item.take() {
             let ctx2 = unsafe { Ctx::from_raw(ctx) };
+
+            while ctx2.execute_pending_job() {}
+
             if let Ok(timeout) = timeout.restore(&ctx2) {
                 timeout.call::<_, ()>(())?;
             }
--- a/tests/unit/asynchronous-processing-discipline.test.ts
+++ b/tests/unit/asynchronous-processing-discipline.test.ts
@@ -1,3 +1,5 @@
+const CWD = process.cwd();
+
 async function waitAndExpectSequentialNumbers(tracked, delayMs) {
   await new Promise((resolve) => {
     setTimeout(() => {
@@ -135,5 +137,96 @@ describe("Asynchronous Microtask Function", () => {
     tracked.push(1);
 
     await waitAndExpectSequentialNumbers(tracked, 10);
+  });
+});
+
+describe("Asynchronous Require/Import", () => {
+  it("When a require is executed synchronously within the main thread, the following statements are executed immediately after the module is loaded.", async () => {
+    const tracked = [];
+
+    tracked.push(1);
+    require(`${CWD}/fixtures/empty.js`);
+    tracked.push(2);
+
+    await waitAndExpectSequentialNumbers(tracked, 10);
+  });
+
+  it("When require is placed inside a queueMicrotask, it executes after the current synchronous execution context finishes and before any scheduled macrotasks.", async () => {
+    const tracked = [];
+
+    queueMicrotask(() => {
+      tracked.push(2);
+      require(`${CWD}/fixtures/empty.js`);
+      tracked.push(3);
+    });
+    queueMicrotask(() => tracked.push(4));
+
+    tracked.push(1);
+
+    await waitAndExpectSequentialNumbers(tracked, 10);
+  });
+
+  it("When require is placed inside a setTimeout, it executes after all synchronous code and all microtasks have been completed, following the macrotask scheduling.", async () => {
+    const tracked = [];
+
+    setTimeout(() => {
+      tracked.push(3);
+      require(`${CWD}/fixtures/empty.js`);
+      tracked.push(4);
+    }, 0);
+    queueMicrotask(() => tracked.push(2));
+
+    tracked.push(1);
+
+    await waitAndExpectSequentialNumbers(tracked, 10);
+  });
+
+  it("When require is placed inside a Promise.then, it executes after the current synchronous execution context finishes but before any macrotasks, following the microtask scheduling rules.", async () => {
+    const tracked = [];
+
+    Promise.resolve().then(() => {
+      tracked.push(2);
+      require(`${CWD}/fixtures/empty.js`);
+      tracked.push(3);
+    });
+    queueMicrotask(() => tracked.push(4));
+
+    tracked.push(1);
+
+    await waitAndExpectSequentialNumbers(tracked, 10);
+  });
+
+  it("When multiple queueMicrotask calls contain require, they execute in the order they were scheduled, immediately after all synchronous code, according to FIFO (first-in, first-out) microtask queue behavior.", async () => {
+    const tracked = [];
+
+    queueMicrotask(() => {
+      tracked.push(2);
+      require(`${CWD}/fixtures/empty.js`);
+      tracked.push(3);
+    });
+    queueMicrotask(() => {
+      tracked.push(4);
+      require(`${CWD}/fixtures/empty.js`);
+      tracked.push(5);
+    });
+
+    tracked.push(1);
+
+    await waitAndExpectSequentialNumbers(tracked, 10);
+  });
+
+  it("When using await import(), the code following the await resumes only after all earlier queued microtasks have been executed, as await yields execution back to the event loop.", async () => {
+    const tracked = [];
+
+    (async () => {
+      tracked.push(1);
+      await import(`${CWD}/fixtures/empty.js`);
+      tracked.push(4);
+    })();
+
+    queueMicrotask(() => tracked.push(3));
+    tracked.push(2);
+
+    await waitAndExpectSequentialNumbers(tracked, 20);
   });
 });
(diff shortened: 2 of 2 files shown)


nabetti1720:
> I am a bit concerned that this doesnt yield the thread like ever. So running on a shared thread will be kinda bad.

Thank you for your comment. It looks like `poll_timers()`, which runs the timer tasks, is called asynchronously within `create_spawn_loop()`, which calls it.

Does that mean that only the `execute_pending_job()` part should be called further asynchronously?

Sytten:
Yeah my concern is about the while loop of executing pending jobs because it never yields the thread ever. So if you have a pending job waiting on say IO on the same thread it will never resolve I think.

richarddavison:
> Yeah my concern is about the while loop of executing pending jobs because it never yields the thread ever. So if you have a pending job waiting on say IO on the same thread it will never resolve I think.

This may be a valid concern. However, I don't think it's possible to solve without patching `rquickjs` which already uses a very complex scheduler. The tricky part here is that we're mixing async runtimes (quickjs promises + Tokio).

nabetti1720:
@Sytten I've tried several approaches to decouple `execute_pending_job()` from an asynchronous process, but none seem to work.

For example, the following code:

```rust
while is_job_pending(rt) {
    let ctx3 = ctx2.clone();
    ctx3.clone().spawn_exit_simple(async move {
        ctx3.execute_pending_job();
        Ok(())
    });
}

```

```rust
fn do_this_instead_of_a_while_loop(ctx: NonNull<JSContext>) -> Result<()> {
    let ctx2 = unsafe { Ctx::from_raw(ctx) };
    let mut is_pending_job = true;

    while is_pending_job {
        let ctx3 = unsafe { Ctx::from_raw(ctx) };

        let handle = tokio::task::spawn_blocking(move || ctx3.execute_pending_job());

        is_pending_job = tokio::runtime::Runtime::new()
            .unwrap()
            .block_on(handle)
            .or_throw(&ctx2)?;
    }
    Ok(())
}
```

@richarddavison Perhaps we are trying to address a problem that is difficult to solve at this point? Would you like to wait on this assignment until the problem you mentioned is completely resolved?

richarddavison:
@nabetti1720 @Sytten Actually it might be fine blocking async tasks while we process microtasks. It does't seem like node yields the async runtime (event loop) while running microtasks:
```
import fs from "fs/promises";

const task = fs.readFile("a-very-large-1.5gb-file.bin");

let breakAt = 0;

queueMicrotask(() => {
  let now = Date.now();
  let current = 0;
  let i = 0;
  while (true) {
    current = Date.now();
    if (current - now > 1000) {
      i++;
      console.log(`waiting ${i}s...`, task);
      now = current;
      if (i > 4) {
        breakAt = Date.now();
        break;
      }
    }
  }
});

task.then((data) => {
  console.log("done", data, Date.now() - breakAt);
});
```
Since I'm blocking the event loop here in my microtask, it still taks about 500ms to read the file after i break out of the loop. If the async task was progressing in the background it should have already been completed reading after 5 seconds.

nabetti1720:
So am I correct in understanding that there is no difference between nodejs and this PR in that no other processing is done while the microtask is running?

nabetti1720:
The correct idea for an event loop was to process all microtasks after processing one macrotask. However, it is also necessary that there are no microtasks before processing the first macrotask, so we reconsidered the implementation.

richarddavison:
> @richarddavison I trust you on that one since you know more, I just felt a bit uneasy with this while loop busy "wait". It feels wrong for that code to live in the timers.

I see what you mean and I agree but I tend to think of as yielding back to the JS runtime.

@nabetti1720 I think we need some more tests for this to cover edge cases such as as doing a require inside a micro task, or an async micro task function etc.

nabetti1720:
I haven't been able to cover all the edge cases yet, but I would like to create test cases that demonstrate discipline in this way. This is because preparing individual tests might have left the incorrect implementation undetectable due to cross-function priorities.

```javascript
// realitySettimeout-nested.cjs
const tracked = [];

// [2 vs 4] In synchronous processing, if both macrotasks and microtasks are registered, the microtasks will be processed first.
queueMicrotask(() => (tracked.push(2)));

setTimeout(() => {
  // [4 vs 7] Timers that fire simultaneously will be executed in the order they were registered.
  queueMicrotask(() => (tracked.push(4)));

  Promise.resolve("")
    .then((value) => queueMicrotask(() => tracked.push(5)))
    .then(() => queueMicrotask(() => tracked.push(6)));

  setTimeout(() => {
    // [10 vs 7] When you register a new timer inside a timer, it will be stacked at the end of the timers already registered.
    tracked.push(10);

    Promise.resolve("")
      .then((value) => queueMicrotask(() => tracked.push(11)))
      .then(() => queueMicrotask(() => tracked.push(12)));
  });
});

setTimeout(() => {
  queueMicrotask(() => (tracked.push(7)));

  Promise.resolve("")
    .then((value) => queueMicrotask(() => tracked.push(8)))
    .then(() => queueMicrotask(() => tracked.push(9)));

  setTimeout(() => {
    tracked.push(13);

    Promise.resolve("")
      .then((value) =>queueMicrotask(() => tracked.push(14)))
      .then(() => queueMicrotask(() => tracked.push(15)));
  });  
});

Promise.resolve().then(() => queueMicrotask(() => (tracked.push(3))));

// [1 vs other] The synchronization process is performed as quickly as possible.
tracked.push(1);


setTimeout(() => {
  console.log(tracked);
}, 10);
```

```
% node realitySettimeout-nested.cjs    
[
   1,  2,  3,  4,  5,  6,
   7,  8,  9, 10, 11, 12,
  13, 14, 15
]
shinya@MBA2022M2 llrt-test % deno run realitySettimeout-nested.cjs
[
   1,  2,  3,  4,  5,  6,
   7,  8,  9, 10, 11, 12,
  13, 14, 15
]
shinya@MBA2022M2 llrt-test % bun realitySettimeout-nested.cjs     
[
  1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15
]
shinya@MBA2022M2 llrt-test % llrt realitySettimeout-nested.cjs    
[ 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15 ]
```

nabetti1720:
I don't have a test case yet combining require, but so far it's working well. I'm working on making sure the same results are achieved in other runtimes.

nabetti1720:
There is an incompatibility when calling `require()` inside an asynchronous prosessing.

```javascript
// asynchronous-processing-discipline.cjs
const tracked = [];

async function waitAndResultOutout(tracked, delayMs) {
   await new Promise((resolve) => {
      setTimeout(() => {
         console.log(tracked);
         resolve();
      }, delayMs);
   });
}

queueMicrotask(() => {
   tracked.push(2);
   const a = require('/Users/shinya/Workspaces/llrt/fixtures/hello.js');
   tracked.push(3);
});
queueMicrotask(() => tracked.push(4));

tracked.push(1);

await waitAndResultOutout(tracked, 10);
```

```
% llrt asynchronous-processing-discipline.cjs
[ 1, 2, 4, 3 ]
% bun asynchronous-processing-discipline.cjs 
[ 1, 2, 3, 4 ]
% deno run asynchronous-processing-discipline.cjs
[ 1, 2, 3, 4 ]
% node asynchronous-processing-discipline.cjs
[ 1, 2, 3, 4 ]
```

Is this because the implementation of `require()` runs its own event loop or something?

nabetti1720:
@richarddavison Sorry for the late reply. I've added some test cases for you to check.

richarddavison:
@nabetti1720 nice job! This now looks great! Tests are also pretty extensive and should cover edge cases!

richarddavison (approved):

Issue #787, opened 2025-01-16
ENHANCEMENT: Standalone executable compile option for local builds


jasongodev:
LLRT, as many have already noticed, is the smallest JavaScript runtime implementing the largest number of Node.js APIs and WinterCG proposals. Major runtimes like Bun and Deno approach feature completeness and Node.js compatibility, but none match LLRT's current usable feature set within such a small footprint.

Standalone executables weren't part of LLRT's initial design. However, its local binaries (i.e. non-Lambda and non-Container) work effectively as a JavaScript CLI runtime and REPL. This is unsurprising, given that the underlying QuickJS engine provides these features and the option to compile standalone executables.

Just look at the current standalone executable sizes between Bun, Deno, and LLRT (LLRT + bytecode size)

| Windows x64 Runtime                    | Standalone Executable Size |
|----------------------------------------|----------------------------|
| Bun                                    | 107 MB                     |
| Deno                                   | 81 MB                      |
| LLRT + bytecode                        | 9 MB                       |
| LLRT, upx -9 --ultra-brute, + bytecode | 3 MB                       |

Here are some of the use cases for a tiny standalone executable runtime:
- rapid prototyping of CLI programs
- distribution of development prototypes
- teaching JavaScript in K-12, without the burden of the whole package management ecosystem
- alternative to Python or Bash scripting for day-to-day administration task
- game development

ahaoboy:
tjs is also a very small js runtime


https://github.com/saghul/txiki.js

https://github.com/ahaoboy/js-engine-benchmark

richarddavison:
@darcyclarke @jasongodev Not quite sure what you mean by "Standalone executable compile option".
Something like `llrt --compile --target arm64 bundle.js -o llrt-executable`?

Ray-D-Song:
我基于llrt实现了这个功能，如果有人需要的话，可以查看这个repo：

https://github.com/Ray-D-Song/lexe

PR #990, opened 2025-05-22
feat: add `--executable` flag to `llrt compile`

darcyclarke:
Fixes: #787

### Description of changes

This introduces support for generating self-contained executables by passing the `--executable` option to `llrt compile`, enabling embedding of bytecode along with the runtime for easier distribution (ex. CLI tools). 

### Checklist

### DIFF
--- a/llrt/src/main.rs
+++ b/llrt/src/main.rs
@@ -3,6 +3,7 @@
 use std::{
     env,
     error::Error,
+    fs,
     path::{Path, PathBuf},
     process::{exit, ExitCode},
     sync::atomic::Ordering,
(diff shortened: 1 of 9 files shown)


nabetti1720:
I think this is a great initiative, and I just have a few suggestions. :)
`core/module/require` was originally a mix of llrt-specific code and nodejs-derived resolver code, but was separated some time ago. The reason is that in the future `core/module/require` will be separated into a separate crate called `llrt_require`. Is it possible to make the code added this time part of `core/module/embedded`?

llrt/src/main.rs
```diff
@@ -51,6 +52,81 @@ async fn main() -> Result<ExitCode, Box<dyn Error + Send + Sync>> {
...
+                            [bytecode_start..bytecode_start + bytecode_size]
+                            .to_vec();
+                        trace!("Extracted bytecode of size {} bytes", bytecode.len());
+
+                        // Create a VM and run the bytecode
+                        let vm = Vm::new().await?;
+                        trace!("Initialized VM in {}ms", now.elapsed().as_millis());
+
+                        // Run the extracted bytecode directly (raw bytecode format from QuickJS)
+                        let result = vm.run_raw_bytecode(bytecode).await;
```
richarddavison, line 114: LLRT already has primitives for reading compressed and uncompressed bytecode as a module. Adding compression (default) and let users opt out is preferred since it creates way smaller binaries.

llrt/src/main.rs
```diff
@@ -51,6 +52,81 @@ async fn main() -> Result<ExitCode, Box<dyn Error + Send + Sync>> {
...
+
+    // Only check for embedded bytecode if we're not running the llrt binary itself
+    if !executable_filename.starts_with("llrt") {
+        trace!(
+            "Checking if {} is a self-contained executable",
+            executable_path.display()
+        );
+
+        // Try to read the executable file to check for embedded bytecode
+        if let Ok(executable_content) = fs::read(&executable_path) {
```
richarddavison, line 71: This implementation conflicts with lambda build variant as it's a self extracting executable. Disable this feature on Lambda variants.

llrt_core/src/modules/require/loader.rs
richarddavison, line 1: We already have some magic headers and makers in https://github.com/awslabs/llrt/blob/935d7553f54e0e871578293a1fb1fc7af32ee1d7/llrt_core/src/bytecode.rs
I think a lot of the logic in this PR could be simplified if we change BYTECODE_CACHE to allow for runtime mutations. That way, module loading is already implemented.

richarddavison (changes requested):
Thanks for this PR 🎉. It's an amazing feature but needs some rework to avoid redundancy from existing primitives.
1. Chage BYTECODE_CACHE to be a global mutable HashMap instead. The current implementation needs to change so that we populate the cache at from embedded values generated at compile time (for the Lambda and std build variants)
2. Now that BYTECODE_CACHE its dynamic. It can be appended from the data appended to the executable.
3. This solves all the existing require/esm compression complexity and removes a lot of duplicated code.
Let me know if you need additional support!

darcyclarke:
@richarddavison Hey! Sorry this PR has stalled out - feel free to make any changes you want to get it over the finish-line (as I have minimal capacity at the moment for back-and-fourth). FWIW, I think I addressed most of the feedback here & pushed that up about a month or so ago. Notably, I've been using my fork this past month & love rolling CLI tools with this feature.

richarddavison:
Apologies for the delayed response, 'm currently on PTO. Overall, I think this is a promising idea. That said, we may need to reconsider the approach slightly to better support targeting multiple architectures.

One potential direction is to take inspiration from Deno's model, where a fresh copy of the runtime is downloaded and the bytecode is appended to the end, creating a self-contained executable. Combining this with SDK bundling could make LLRT more flexible and general-purpose. For instance, the AWS SDK could be bundled like any other dependency, eliminating the need for a separate build step.

richarddavison:
Superseded by https://github.com/awslabs/llrt/pull/1095

PR #996, opened 2025-05-31
fix(fetch/headers): implement Header Guards according to the Fetch Standard

nabetti1720:
### Description of changes

We have started implementing [Headers Guard](https://fetch.spec.whatwg.org/#headers-guard) in Fetch Standard. However, implementations other than `request-no-cors` have not been thoroughly tested.

This fixed the following test to pass:
 - fetch/api/headers > should pass headers-no-cors.any.js tests

### Checklist

### DIFF
--- a/modules/llrt_fetch/src/request.rs
+++ b/modules/llrt_fetch/src/request.rs
@@ -14,6 +14,40 @@ use super::{
     Blob,
 };
 
+#[derive(Clone, Default, PartialEq)]
+pub enum RequestMode {
+    #[default]
+    Cors,
+    NoCors,
+    SameOrigin,
+    Navigate,
+}
+
+impl TryFrom<String> for RequestMode {
+    type Error = String;
+
+    fn try_from(s: String) -> std::result::Result<Self, Self::Error> {
+        Ok(match s.to_ascii_lowercase().as_str() {
+            "cors" => RequestMode::Cors,
+            "no-cors" => RequestMode::NoCors,
+            "same-origin" => RequestMode::SameOrigin,
+            "navigate" => RequestMode::Navigate,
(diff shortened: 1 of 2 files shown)


modules/llrt_fetch/src/request.rs
```diff
@@ -35,6 +38,7 @@ pub struct Request<'js> {
     headers: Option<Class<'js, Headers>>,
     body: Option<Value<'js>>,
     signal: Option<Class<'js, AbortSignal<'js>>>,
+    mode: String,
```
richarddavison: Should mode be string or an enum? At lest it should be validated to be only: cors, no-cors, same-origin, or navigate. Default to cors.

modules/llrt_fetch/src/request.rs
```diff
@@ -53,11 +57,12 @@ impl<'js> Request<'js> {
...
     pub fn new(ctx: Ctx<'js>, input: Value<'js>, options: Opt<Value<'js>>) -> Result<Self> {
         let mut request = Self {
-            url: String::from(""),
-            method: "GET".to_string(),
+            url: "".into(),
+            method: "GET".into(),
             headers: None,
             body: None,
             signal: None,
+            mode: "no-cors".into(),
```
richarddavison: Default should also be cors as per MDN docs:

> For example, when a Request object is created using the [Request()](https://developer.mozilla.org/en-US/docs/Web/API/Request/Request) constructor, the value of the mode property for that Request is set to cors.

richarddavison (changes requested):
Thanks for the PR. Some minor comments!

nabetti1720:
@richarddavison I have fixed it, so please leave a review. :)

richarddavison (approved):
LGTM! Thank you!

PR #1002, opened 2025-06-01
chore: crate organization

nabetti1720:
### Description of changes

As the size of our assets has grown, in this PR we will be organizing our crates.
- Removed crates that were never referenced in the code.
- We set `default-features=false` for all external crates, whether they have feature flags or not, to prevent unintentional introduction of unwanted features, and honestly. We didn't have time to evaluate them one by one. :)
- llrt_crypto: Fixed some missing feature flag settings.
- llrt_os: Introduced the systems feature flag. Turning off all of the included feature flags will prevent the inclusion of the sysinfo crate. 

Looking at the diff of Cargo.lock, we have made progress in removing dependencies on several crates, resulting in a 48KB reduction in size. Below is the execution result using full-sdk.

before:
```sh
% du -k ./target/aarch64-apple-darwin/release/llrt
10604   ./target/aarch64-apple-darwin/release/llrt
```

after:
```sh
% du -k ./target/aarch64-apple-darwin/release/llrt
10556   ./target/aarch64-apple-darwin/release/llrt
```

### Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2058,7 +2058,6 @@ name = "llrt_test"
 version = "0.5.1-beta"
 dependencies = [
  "nanoid",
- "rand 0.8.5",
  "rquickjs",
  "tokio",
 ]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1757,7 +1757,6 @@ dependencies = [
  "p256",
  "p384",
  "p521",
- "pkcs8",
  "rand 0.8.5",
  "ring",
  "rquickjs",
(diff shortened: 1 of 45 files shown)


Sytten:
Good change, we should also do feature testing of modules to ensure they dont break and deps are correctly set.

nabetti1720:
> Good change, we should also do feature testing of modules to ensure they dont break and deps are correctly set.

Is there a good and generic way to do this?

NOTE: At least some detection was possible by running the following command locally. I'd like to make it possible.
```sh
% cargo check -p [crate_name]
```

nabetti1720:
@richarddavison I've accomplished what I wanted to do with this PR.

richarddavison:
> @richarddavison I've accomplished what I wanted to do with this PR.

Currently traveling but I'll take a look ASAP! Thanks for this!

nabetti1720:
Did you enjoy the travel? I made some tweaks, but I think I accomplished what I wanted to do with this PR. :)

Issue #1007, opened 2025-06-03
Further improving the footprint


nabetti1720:
LLRT already has features that other runtimes do not have, but we believe that we need to further refine them, even if we have to forego some compatibility.

To further improve the small footprint, one of the features of this product, we are considering the following:

- llrt_fetch: Keep http2 support as a crate, but default LLRT to http1 only (this will prevent the `h2` crate from being included)
- llrt_fetch: By adopting `rustls-platform-verifier` instead of `webkpi-roots`, certificates are not stored internally. See also https://github.com/awslabs/llrt/issues/768#issuecomment-2930897137
- llrt_os: We'll put all the functions the `sysinfo` crate needs behind the feature flags and change the default to off.

Is there anything else you can think of?

ahaoboy:
I'm not sure if there is a definite target size, such as within 10MB?, but judging from the bloat results, there is little room for optimization.

![Image](https://github.com/user-attachments/assets/fa672691-9da2-4e2f-bb9d-de2a6e2aaeda)

nabetti1720:
Yes. I don't expect any dramatic improvements, such as cutting the footprint in half. There are also no target values. The idea is that features that are not high priority can be omitted.

I'm interested in the image you presented. For example, if we could remove the h2 crate, would that lead to roughly a 10% reduction overall?

NOTE: I would be grateful if you could tell me how to output this image. I would love to try it myself. :)

ahaoboy:
> I'm interested in the image you presented. For example, if we could remove the h2 crate, would that lead to roughly a 10% reduction overall?

I'm not sure, since size optimization often depends on dependencies. For example, phf is a dependency of rquickjs-core, so simply removing it from llrt won't necessarily reduce the final binary size.



> NOTE: I would be grateful if you could tell me how to output this image. I would love to try it myself. :)

This is a tool  https://github.com/ahaoboy/bloaty-metafile I previously used to analyze the size of llrt modules. Feel free to follow the instructions in the README to try it out. If you run into any issues, you're welcome to open an issue.

nabetti1720:
@ahaoboy Thank you! It worked fine on my device. I think it would be a good idea to improve the README.md file a bit. I'll do a PR on another occasion.

![Image](https://github.com/user-attachments/assets/7ed1e04f-c92f-452e-97fa-8c9b45bb1d0d)

Sytten:
[...]

PR #1020, opened 2025-06-12
feat(fetch/os/zlib): make features that use some crates optional

nabetti1720:
### Issue # (if available)

Closed: #1007

### Description of changes

When building llrt, some crates' large or less important features are optional. Optional features can be enabled by including the following in the features section of llrt_modules.

- fetch-brotli
- fetch-http2
- os-network
- os-statistics
- os-system
- zlib-brotli

By having these features disabled by default, binary sizes are further reduced.

before(full-sdk):
```
% du -k ./target/aarch64-apple-darwin/release/llrt
10492   ./target/aarch64-apple-darwin/release/llrt

% cargo bloat --profile=flame --crates

 File  .text     Size Crate
13.6%  20.8%   1.4MiB rquickjs_core
11.1%  16.9%   1.2MiB [Unknown]
 7.4%  11.2% 795.0KiB std
 3.3%   5.0% 353.6KiB h2
 3.1%   4.8% 339.6KiB llrt_stream_web
 2.4%   3.7% 260.3KiB rustls
 2.0%   3.1% 217.7KiB zstd_sys
 1.7%   2.7% 188.9KiB tokio
 1.4%   2.1% 147.9KiB ring
 1.3%   2.0% 141.0KiB llrt_crypto
 1.2%   1.8% 124.7KiB hyper
 1.1%   1.7% 117.0KiB llrt_core
 1.0%   1.6% 112.9KiB hyper_util
 0.9%   1.4% 100.7KiB rquickjs_sys
 0.9%   1.3%  92.0KiB brotlic_sys
 0.8%   1.2%  83.6KiB llrt
 0.6%   1.0%  67.3KiB num_bigint_dig
 0.6%   0.9%  66.8KiB primeorder
 0.6%   0.9%  63.2KiB llrt_utils
 0.6%   0.9%  61.0KiB http
10.2%  15.6%   1.1MiB And 120 more crates. Use -n N to show more.
65.4% 100.0%   6.9MiB .text section size, the file size is 10.5MiB
```

after(full-sdk):
```
% du -k ./target/aarch64-apple-darwin/release/llrt
9004    ./target/aarch64-apple-darwin/release/llrt

% cargo bloat --profile=flame --crates

 File  .text     Size Crate
16.1%  23.4%   1.4MiB rquickjs_core
10.7%  15.6% 963.7KiB [Unknown]
 8.2%  12.0% 740.6KiB std
 3.8%   5.5% 339.6KiB llrt_stream_web
 2.8%   4.1% 254.0KiB rustls
 2.4%   3.5% 217.7KiB zstd_sys
 2.1%   3.0% 187.1KiB tokio
 1.6%   2.4% 147.9KiB ring
 1.6%   2.3% 141.2KiB llrt_crypto
 1.2%   1.8% 111.3KiB llrt_core
 1.1%   1.6% 100.7KiB rquickjs_sys
 1.1%   1.6%  97.4KiB hyper
 1.0%   1.5%  91.3KiB hyper_util
 0.9%   1.3%  82.8KiB llrt
 0.7%   1.1%  67.2KiB num_bigint_dig
 0.7%   1.1%  66.8KiB primeorder
 0.7%   1.0%  63.0KiB llrt_utils
 0.6%   0.9%  54.6KiB llrt_fetch
 0.6%   0.8%  51.8KiB der
 0.6%   0.8%  50.0KiB url
10.4%  15.2% 938.3KiB And 118 more crates. Use -n N to show more.
68.7% 100.0%   6.0MiB .text section size, the file size is 8.8MiB
```

### Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1745,9 +1745,7 @@ dependencies = [
  "home",
  "http-body-util",
  "hyper",
- "itoa",
  "jwalk",
- "libc",
  "llrt_build",
  "llrt_context",
  "llrt_encoding",
@@ -1767,7 +1765,6 @@ dependencies = [
  "rquickjs",
  "rustls",
  "rustls-pemfile",
- "ryu",
  "simd-json",
  "terminal_size",
  "tokio",
@@ -3475,7 +3472,6 @@ version = "0.1.41"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "784e0ac535deb450455cbfa28a6f0df145ea1bb7ae51b821cf5e7927fdcfbdd0"
 dependencies = [
- "log",
  "pin-project-lite",
  "tracing-core",
 ]
--- a/README.md
+++ b/README.md
@@ -530,6 +530,9 @@ Set a memory threshold in MB for garbage collection. Default threshold is 20MB
 
 Extends the HTTP request version. By default, only HTTP/1.1 is enabled. Specifying '2' will enable HTTP/1.1 and HTTP/2.
 
+> [!NOTE]
+> Although it is supported by the llrt_fetch crate, the http/2 feature is disabled at build time to reduce binary size. Regardless of the setting of this environment variable, llrt only works with http/1.1.
+
 ### `LLRT_LOG=[target][=][level][,...]`
 
 Filter the log output by target module, level, or both (using `=`). Log levels are case-insensitive and will also enable any higher priority logs.
--- a/libs/llrt_compression/Cargo.toml
+++ b/libs/llrt_compression/Cargo.toml
@@ -11,10 +11,9 @@ name = "llrt_compression"
 path = "src/lib.rs"
 
 [features]
-default = ["all-c"]
-
-all-c = ["brotli-c", "flate2-c", "zstd-c"]
-all-rust = ["brotli-rust", "flate2-rust", "zstd-rust"]
+default = ["flate2-c", "zstd-c"]
+#default = ["std"]
+std = ["brotli-c", "flate2-c", "zstd-c"]
 
 brotli-c = ["brotlic"]
 brotli-rust = ["brotli"]
--- a/llrt/Cargo.toml
+++ b/llrt/Cargo.toml
@@ -26,9 +26,7 @@ tokio = { version = "1", features = [
   "macros",
   "rt-multi-thread",
 ], default-features = false }
-tracing = { version = "0.1", features = [
-  "log"
-], default-features = false }
+tracing = { version = "0.1", default-features = false }
 tracing-core = { version = "0.1", default-features = false }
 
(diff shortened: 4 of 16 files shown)


richarddavison (changes requested):
This is great, thank you! 🎉 The only change I'm a bit unsure about is that http2 is disabled by default. @Sytten what's your take?

Sytten:
Couple of things

**C/Rust implementations**
I am a bit annoyed that we made it harder to use the rust based crates. Like the fetch module just can't have rust based compression with this change. To avoid the explosion of flags I think we should consider a model like Rustls were the backend is enabled by the user.

Essentially in this model you define an interface but you don't enable any implementation by default. It does result in an error if a user doesn't enable one, but we can have documentation on that. You have to do that because we can't do negative features in Rust.

In our compression case what this would mean is:
- In fetch we have a `brotli` flag that enables the usage of the compression API
- The user MUST add `llrt_compression` to their Cargo and enable either `brotli-c` or `brotli-rs`

**Http**
Disabling http/2 by default at the module level is fine, but I would suggest enabling it for the llrt release otherwise performances in production will tank just to same a couple KB of runtime.

**STD**
I am not a fan of the `std` feature name. This has a specific meaning in rust (aka using the standard library) and this is not it. We should have our own like `basic`, `all`, `recommended`.

richarddavison:
> Couple of things
> 
> **C/Rust implementations** I am a bit annoyed that we made it harder to use the rust based crates. Like the fetch module just can't have rust based compression with this change. To avoid the explosion of flags I think we should consider a model like Rustls were the backend is enabled by the user.
> 
> Essentially in this model you define an interface but you don't enable any implementation by default. It does result in an error if a user doesn't enable one, but we can have documentation on that. You have to do that because we can't do negative features in Rust.
> 
> In our compression case what this would mean is:
> 
> * In fetch we have a `brotli` flag that enables the usage of the compression API
> * The user MUST add `llrt_compression` to their Cargo and enable either `brotli-c` or `brotli-rs`

I agree with all these points. @nabetti1720 is this something you can incorporate? 

> **Http** Disabling http/2 by default at the module level is fine, but I would suggest enabling it for the llrt release otherwise performances in production will tank just to same a couple KB of runtime.

Yeah I want to keep it as well, even if it bloats the runtime a bit

> **STD** I am not a fan of the `std` feature name. This has a specific meaning in rust (aka using the standard library) and this is not it. We should have our own like `basic`, `all`, `recommended`.

This is also something that doesn't sit quite right with me. We should think of a better naming convention. `basic`, `default`, `all` or something

nabetti1720:
@richarddavison @Sytten Thank you for your valuable feedback. Your feedback would make us rethink almost every aspect of this PR and I don't think it's a good idea to move forward with it. (Because there is little hope of reducing the binary size)

I apologize for not being able to make use of your valuable feedback.

PR #1031, opened 2025-06-29
fix(core): fix incompatibilities caused by the latest cargo fix

nabetti1720:
### Description of changes

After applying the latest rust toolchain, we found new incompatibilities and have fixed them.

```
% rustup update
info: syncing channel updates for 'stable-aarch64-apple-darwin'
info: syncing channel updates for 'nightly-aarch64-apple-darwin'
info: checking for self-update

   stable-aarch64-apple-darwin unchanged - rustc 1.88.0 (6b00bc388 2025-06-23)
  nightly-aarch64-apple-darwin unchanged - rustc 1.90.0-nightly (bdaba05a9 2025-06-27)
```

### Checklist

### DIFF
--- a/libs/llrt_logging/src/lib.rs
+++ b/libs/llrt_logging/src/lib.rs
@@ -856,21 +856,21 @@ pub fn print_error_and_exit<'js>(ctx: &Ctx<'js>, err: CaughtError<'js>) -> ! {
     use std::fmt::Write;
 
     let mut error_str = String::new();
-    write!(error_str, "Error: {:?}", err).unwrap();
+    write!(error_str, "Error: {err:?}").unwrap();
 
     if let Ok(error) = err.into_value(ctx) {
         if print_error(ctx, Rest(vec![error.clone()])).is_err() {
-            eprintln!("{}", error_str);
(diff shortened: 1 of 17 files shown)


richarddavison (changes requested):
Thank you very much 👏 However the format!() rules i think we can disable for readability? 👍

nabetti1720:
> Thank you very much 👏 However the format!() rules i think we can disable for readability? 👍

I suppose you could replace it with something like `["foo", "bar"].concat()`, but is this what you want? Or is there a smarter way to disable `format!()`?

richarddavison:
> > Thank you very much 👏 However the format!() rules i think we can disable for readability? 👍
> 
> I suppose you could replace it with something like `["foo", "bar"].concat()`, but is this what you want? Or is there a smarter way to disable `format!()`?

Sorry, that's not what I meant. There is new clippy rues for using format! arguments:
```rust
write!(error_str, "Error: {:?}", err).unwrap();
```
vs.
```rust
write!(error_str, "Error: {err:?}").unwrap();
```

These can be disabled by setting:
`#![allow(clippy::uninlined_format_args)]`

PR #1038, opened 2025-06-30
Add process.env types to process.d.ts

namse:
### Description of changes

Adds the env property to the process.d.ts type definition.

This allows TypeScript users to access process.env without encountering type errors.


richarddavison (approved):
Thanks!

PR #1069, opened 2025-07-21
fix(fetch): remove the BOM from body.text() and body.json()

nabetti1720:
### Description of changes

- The BOM must be ignored (do nothing) in `body.text()` and `body.json()`.
- The same is true for `Request` and `Response`, which is the body-mixin.

### Checklist

### DIFF
--- a/modules/llrt_fetch/src/lib.rs
+++ b/modules/llrt_fetch/src/lib.rs
@@ -138,7 +138,7 @@ fn get_http_version() -> HttpVersion {
     })
 }
 
-fn strip_bom<'a>(bytes: impl Into<Cow<'a, [u8]>>) -> Cow<'a, [u8]> {
+pub(crate) fn strip_bom<'a>(bytes: impl Into<Cow<'a, [u8]>>) -> Cow<'a, [u8]> {
     let cow = bytes.into();
     if bytes.starts_with(&[0xEF, 0xBB, 0xBF]) {
         Cow::Borrowed(&cow[3..])
--- a/modules/llrt_fetch/src/lib.rs
+++ b/modules/llrt_fetch/src/lib.rs
@@ -137,6 +137,14 @@ fn get_http_version() -> HttpVersion {
     })
 }
 
+pub(crate) fn strip_bom(bytes: &[u8]) -> &[u8] {
+    if bytes.starts_with(&[0xEF, 0xBB, 0xBF]) {
+        &bytes[3..]
+    } else {
+        bytes
+    }
+}
+
 pub type HyperClient =
     Client<HttpsConnector<HttpConnector<CachedDnsResolver>>, BoxBody<Bytes, Infallible>>;
 pub static HTTP_CLIENT: Lazy<io::Result<HyperClient>> = Lazy::new(|| {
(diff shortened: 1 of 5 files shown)


modules/llrt_fetch/src/lib.rs
```diff
@@ -137,6 +137,14 @@ fn get_http_version() -> HttpVersion {
...
 }
 
+pub(crate) fn strip_bom(bytes: &[u8]) -> &[u8] {
+    if bytes.starts_with(&[0xEF, 0xBB, 0xBF]) {
+        &bytes[3..]
+    } else {
+        bytes
+    }
+}
+
```
richarddavison: This should return a Cow instead. If bytes are unmodified we don't need the additional copy. Something like:
```rust
pub(crate) fn strip_bom(bytes: impl Into<Cow<'_, [u8]>>) -> Cow<[u8]> {
    let cow = bytes.into();
    if cow.starts_with(&[0xEF, 0xBB, 0xBF]) {
        Cow::Borrowed(&cow[3..])
    } else {
        cow
    }
}
```

richarddavison (changes requested):
Minor optimization

modules/llrt_fetch/src/lib.rs
```diff
@@ -137,6 +138,18 @@ fn get_http_version() -> HttpVersion {
     })
 }
 
+pub(crate) fn strip_bom<'a>(bytes: impl Into<Cow<'a, [u8]>>) -> Cow<'a, [u8]> {
+    let cow = bytes.into();
+    if cow.starts_with(&[0xEF, 0xBB, 0xBF]) {
+        match cow {
```
richarddavison, line 144: We don’t need to match as we need to return a borrowed cow always since we modify the bytes

richarddavison (changes requested):

modules/llrt_fetch/src/lib.rs
nabetti1720: The code you originally suggested gave me an error, so I added the Owned condition, even though I knew it wouldn't be used here.

The code below will compile, but does it do what we expected?

```rust
pub(crate) fn strip_bom<'a>(bytes: &'a [u8]) -> Cow<'a, [u8]> {
    if bytes.starts_with(&[0xEF, 0xBB, 0xBF]) {
        Cow::Borrowed(&bytes[3..])
    } else {
        Cow::Borrowed(&bytes)
    }
}
```

richarddavison: Not quite. If we take in something that can be converted into Cow<[u8]> and is not a bom and we pass vec, there is no additional copy;
```rust
fn strip_bom<'a>(bytes: impl Into<Cow<'a, [u8]>>) -> Cow<'a, [u8]> {
    let cow = bytes.into();
     if bytes.starts_with(&[0xEF, 0xBB, 0xBF]) {
        Cow::Borrowed(&cow[3..])
    } else {
        cow
    }
  }
```

nabetti1720:
@richarddavison Thank you for your assistance, this is exactly the problem I was not able to solve.

richarddavison (approved):

PR #1089, opened 2025-08-07
fix(fetch): improved compatibility of bodyUsed property

nabetti1720:
### Description of changes

- The support for the bodyUsed property of the Request class was insufficient, so it has been improved.
- Ideally, body handling should be standardized according to the Body-Mixin concept, but this will take time, so we will do it another time.

### Checklist

### DIFF
--- a/modules/llrt_fetch/src/request.rs
+++ b/modules/llrt_fetch/src/request.rs
@@ -181,14 +181,14 @@ impl<'js> Request<'js> {
 
     pub async fn text(&mut self, ctx: Ctx<'js>) -> Result<String> {
         if let Some(bytes) = self.take_bytes(&ctx).await? {
-            return Ok(String::from_utf8_lossy(&strip_bom(&bytes)).to_string());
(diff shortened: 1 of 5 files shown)


modules/llrt_fetch/src/request.rs
```diff
@@ -177,32 +181,28 @@ impl<'js> Request<'js> {
 
     pub async fn text(&mut self, ctx: Ctx<'js>) -> Result<String> {
         if let Some(bytes) = self.take_bytes(&ctx).await? {
-            let bytes = bytes.as_bytes(&ctx)?;
-            return Ok(String::from_utf8_lossy(&strip_bom(bytes)).to_string());
+            return Ok(String::from_utf8_lossy(&strip_bom(&bytes)).to_string());
```
richarddavison: Why an extra borrow here? Creates an additional clone

richarddavison (changes requested):
Minor comment. Sorry for the delay! (On PTO)

PR #1095, opened 2025-08-10
feat: add --executable flag to llrt compile

### LINKED ISSUE #787: ENHANCEMENT: Standalone executable compile option for local builds

kyubisation:
### Issue # (if available)

Closes #787

This is a leaner implementation than #990, but still adapts a lot of the existing work from @darcyclarke.
Instead of reading the whole binary into memory, this implementation seeks to the end of it and checks for a specific marker.

I am fairly new to Rust, so I probably made some mistakes or did not follow best practices.
If this PR were to be an acceptable starting point, I can adapt/optimize the implementation as requested.

### Description of changes

This introduces support for generating self-contained executables by passing the `--executable` option to `llrt compile`, enabling embedding of bytecode along with the runtime for easier distribution (e.g. CLI tools).

### Checklist

- [ ] ~~Added relevant type info in `types/` directory~~
- [ ] ~~Updated documentation if needed ([API.md](API.md)/[README.md](README.md)/Other)~~

### DIFF
--- a/llrt/src/main.rs
+++ b/llrt/src/main.rs
@@ -7,6 +7,7 @@ use std::{
     error::Error,
     path::{Path, PathBuf},
     process::{exit, ExitCode},
+    string::String,
     sync::atomic::Ordering,
     time::Instant,
 };
@@ -94,6 +95,8 @@ Options:
                       if [output.lrt] is omitted, <input>.lrt is used.
                       lrt file is expected to be executed by the llrt version
                       that created it
+                      --executable      Create a self-contained executable that includes
+                                        the LLRT runtime
   test              Run tests with provided arguments:
                       <test_args> -d <directory> <test-filter>
 "#
@@ -110,6 +113,36 @@ async fn start_runtime(vm: &Vm) {
 }
 
 async fn start_cli(vm: &Vm) {
+    #[cfg(not(feature = "lambda"))]
+    {
+        use std::io::{Read, Seek};
+
+        let executable_path = env::current_exe()
+            .unwrap_or_else(|_| PathBuf::from(env::args().next().unwrap_or_default()));
+        trace!(
+            "Checking if {} is a self-contained executable",
+            executable_path.display()
+        );
+
+        if let Ok(mut f) = std::fs::File::open(executable_path) {
+            let _ = f.seek(std::io::SeekFrom::End(-12));
+            let mut size = [0; 8];
+            f.read_exact(&mut size).unwrap();
+            let mut marker = [0; 4];
+            f.read_exact(&mut marker).unwrap();
+
+            let marker_string = String::from_utf8_lossy(&marker);
+            if marker_string == crate::base::bytecode::BYTECODE_SELF_CONTAINED_EXECUTABLE_MARKER {
+                let size_number = u64::from_le_bytes(size);
+                let start = f.metadata().unwrap().len() - (size_number + 12u64);
+                let _ = f.seek(std::io::SeekFrom::Start(start));
+                let mut module = vec![0; size_number.try_into().unwrap()];
+                f.read_exact(&mut module).unwrap();
+                return vm.run_bytecode(&module).await;
+            }
+        }
+    }
+
     let args: Vec<String> = env::args().collect();
 
     if args.len() > 1 {
@@ -142,17 +175,31 @@ async fn start_cli(vm: &Vm) {
                         #[cfg(not(feature = "lambda"))]
                         {
                             if let Some(filename) = args.get(i + 1) {
-                                let output_filename = if let Some(arg) = args.get(i + 2) {
-                                    arg.to_string()
-                                } else {
+                                // Parse args for output_filename and --executable
+                                let mut output_filename = String::new();
+                                let mut create_executable = false;
+
+                                // Parse remaining arguments
+                                for arg in args.iter().skip(i + 2) {
+                                    if arg == "--executable" {
+                                        create_executable = true;
+                                    } else if output_filename.is_empty() && !arg.starts_with("--") {
+                                        output_filename = arg.clone();
+                                    }
+                                }
+
+                                // If no output filename was explicitly provided, generate one
+                                if output_filename.is_empty() {
                                     let mut buf = PathBuf::from(filename);
                                     buf.set_extension("lrt");
-                                    buf.to_string_lossy().to_string()
-                                };
+                                    output_filename = buf.to_string_lossy().to_string();
+                                }
 
                                 let filename = Path::new(filename);
                                 let output_filename = Path::new(&output_filename);
-                                if let Err(error) = compile_file(filename, output_filename).await {
+                                if let Err(error) =
+                                    compile_file(filename, output_filename, create_executable).await
+                                {
                                     eprintln!("{error}");
                                     exit(1);
                                 }
--- a/llrt_core/src/vm.rs
+++ b/llrt_core/src/vm.rs
@@ -185,6 +185,29 @@ impl Vm {
         self.run(source, strict, global).await;
     }
 
+    pub async fn run_bytecode(&self, bytecode: &[u8]) {
+        self.run_with(|ctx| {
+            // First, try to load and execute it as a module
+            match crate::modules::embedded::loader::EmbeddedLoader::load_bytecode_module(
+                ctx.clone(),
+                bytecode,
+            ) {
+                Ok(module) => match module.eval() {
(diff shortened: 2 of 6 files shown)


kyubisation:
Seems like the nightly builds fail with the following error:

```bash
   Compiling llrt_core v0.6.2-beta (/home/lukas/projects/llrt/llrt_core)
error[E0080]: evaluation panicked: panic_immediate_abort requires -C panic=abort
  --> /home/lukas/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/panic.rs:63:9
   |
63 |         $crate::panicking::panic_fmt($crate::const_format_args!($($t)+));
   |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ evaluation of `panicking::_` failed here
   |
  ::: /home/lukas/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/panicking.rs:36:15
   |
36 | const _: () = assert!(cfg!(panic = "abort"), "panic_immediate_abort requires -C panic=abort");
   |               ------------------------------------------------------------------------------- in this macro invocation

For more information about this error, try `rustc --explain E0080`.
error: could not compile `core` (lib) due to 1 previous error
```

I'm afraid I do not fully understand what the problem is.

kyubisation:
Ok, this problem also appears on main, so I assume this is a new problem from the current nightly version.

kyubisation:
The error still appears with `make fix` and `make check`. I'm afraid I do not know why.
However, I will try to take another look as to why `make test-ci` fails.

kyubisation:
The above error was a red herring.
I have refactored the test (and added a few more) and that seems to work now.
Sorry for the confusion.

kyubisation:
@richarddavison Can I ask for a quick assessment? Is the current state good to go or would you like an improvement in any kind?

llrt/src/main.rs
```diff
@@ -110,6 +113,36 @@ async fn start_runtime(vm: &Vm) {
...
+            "Checking if {} is a self-contained executable",
+            executable_path.display()
+        );
+
+        if let Ok(mut f) = std::fs::File::open(executable_path) {
+            let _ = f.seek(std::io::SeekFrom::End(-12));
+            let mut size = [0; 8];
+            f.read_exact(&mut size).unwrap();
+            let mut marker = [0; 4];
+            f.read_exact(&mut marker).unwrap();
```
richarddavison: -12,8 and 4 are hard coded values. I think this can be improved by using BYTECODE_SELF_CONTAINED_EXECUTABLE_MARKER length and seeking backwards, reading into a slice once and working of that slice. Or some more robust way of handling this. Also do proper error handling to avoid unwraps.

llrt/src/main.rs
```diff
@@ -110,6 +113,36 @@ async fn start_runtime(vm: &Vm) {
...
+
+            let marker_string = String::from_utf8_lossy(&marker);
+            if marker_string == crate::base::bytecode::BYTECODE_SELF_CONTAINED_EXECUTABLE_MARKER {
```
richarddavison: We can avoid string and compare byte slice of BYTECODE_SELF_CONTAINED_EXECUTABLE_MARKER instead

llrt/src/main.rs
```diff
@@ -110,6 +113,36 @@ async fn start_runtime(vm: &Vm) {
...
+                let size_number = u64::from_le_bytes(size);
+                let start = f.metadata().unwrap().len() - (size_number + 12u64);
+                let _ = f.seek(std::io::SeekFrom::Start(start));
+                let mut module = vec![0; size_number.try_into().unwrap()];
+                f.read_exact(&mut module).unwrap();
+                return vm.run_bytecode(&module).await;
```
richarddavison: avoid unwraps

llrt_core/src/vm.rs
```diff
@@ -185,6 +185,29 @@ impl Vm {
...
+                    Ok(_) => Ok(()),
+                    Err(err) => {
+                        eprintln!("Failed to evaluate module: {err:?}");
+                        Err(err)
+                    },
+                },
+                Err(err) => {
+                    eprintln!("Error extracting bytecode: {err:?}");
+                    Err(err)
+                },
```
richarddavison: This can be cleaned up

richarddavison (changes requested):
Hi @kyubisation. Thanks for this PR and sorry for the delay (I've been on PTO and have been catching up after a few weeks of :) ). I generally like the idea of this PR and code looks pretty good, some minor comments:

We should also handle a "--target" option in the future. With this current approach we can only compile for the target that we're running on. For example, if I'm on an ARM Mac I can't compile for Linux x86. This is beyond the scope of this PR but we should in the future download the LLRT executable from GitHub releases and append the compiled and compressed bytecode to that executable.

kyubisation:
@richarddavison Thank you very much for the feedback. I'll try to address the raised issues. 
I apologize, I did not mean to hurry you. I hope you were able to enjoy your time off. 😄

llrt/src/main.rs
kyubisation: For clarification: Using `f.seek(std::io::SeekFrom::End(-12))` is acceptable? (Apart from replaceing -12 with -8 subtracted by marker byte length)
My though was that I wanted to avoid loading the entire binary into memory, as my assumption is that seek is faster/more efficient.
Do I understand you correctly, that your suggestion is to then use `read_exact` to read 12 bytes to a slice?

I apologize if my questions seem naive. This is my first experience with rust.

kyubisation:
@richarddavison As mentioned, this is my first experience with rust. I tried to address the mentioned issues. If it is not to your expectation, I would be grateful for a few pointers or an example for improvement.
Thank you very much for your time.

llrt/src/main.rs
richarddd: > For clarification: Using `f.seek(std::io::SeekFrom::End(-12))` is acceptable? (Apart from replaceing -12 with -8 subtracted by marker byte length) My though was that I wanted to avoid loading the entire binary into memory, as my assumption is that seek is faster/more efficient. Do I understand you correctly, that your suggestion is to then use `read_exact` to read 12 bytes to a slice?
> 
> I apologize if my questions seem naive. This is my first experience with rust.

I dont think the approach is wrong at all, seek is faster (it depends on the underlying FS tho) i was just referring to avoiding using numbers like -12 as we dont know what they are drived from. use something like (psudocode)`f.seek(std::io::SeekFrom::End(-(BYTECODE_SELF_CONTAINED_EXECUTABLE_MARKER.len() + size_of::<u64>())` instead

richarddavison:
> @richarddavison As mentioned, this is my first experience with rust. I tried to address the mentioned issues. If it is not to your expectation, I would be grateful for a few pointers or an example for improvement. Thank you very much for your time.

The approach is right I was just commenting on the hard coded values. Use size_of::<f64>() and the length of `BYTECODE_SELF_CONTAINED_EXECUTABLE_MARKER.len()` instead :)

kyubisation:
Thank you for the guidance. I have replaced the remaining hardcoded 8 values with `size_of::<u64>()`.

kyubisation:
@richarddavison Thank you for the merge. If a problem comes up with this implementation, feel free to ping me.

PR #1103, opened 2025-08-15
fix(fetch): check for valid uri in fetch

nabetti1720:
### Description of changes

- Only schemes supported by the LLRT fetch implementation are allowed.
- Additionally, ports that should be blocked by specification are now verified.

### Checklist


modules/llrt_fetch/src/fetch.rs
```diff
@@ -35,6 +36,19 @@ use super::{
     Blob,
 };
 
+// https://fetch.spec.whatwg.org/#port-blocking
+static BLOCKED_PORTS: LazyLock<HashSet<u16>> = LazyLock::new(|| {
```
richarddavison: I use const array. HashSet might be significantly slower for low count integer. Also avoids lazy look

modules/llrt_fetch/src/fetch.rs
```diff
@@ -193,6 +208,17 @@ fn build_request(
     prev_status: &u16,
     initial_uri: &Uri,
 ) -> Result<(Request<BoxBody<Bytes, Infallible>>, HeadersGuard)> {
+    if let Some(scheme) = uri.scheme_str() {
+        if !matches!(scheme, "http" | "https") {
+            return Err(Exception::throw_type(ctx, "Invalid scheme in URL"));
+        }
+    }
+    if let Some(port) = uri.authority().and_then(|a| a.port_u16()) {
+        if BLOCKED_PORTS.contains(&port) {
```
Sytten: As per the spec if you negotiate the apln via TLS you dont need to block ports. So likely the https scheme could skip that.

nabetti1720: Thank you for your comment. Are the specifications you are referring to the following?

[2.9. Port blocking](https://fetch.spec.whatwg.org/#port-blocking)
> New protocols can avoid the need for blocking ports by negotiating the protocol through TLS using ALPN. The protocol cannot be spoofed through HTTP requests in that case. [[RFC7301]](https://fetch.spec.whatwg.org/#biblio-rfc7301)

Since hyper_rustls seems to use the ALPN protocol internally, would it be okay to omit the port check when using https? In addition, in order to pass the WPT test, we would like to retain the specification that checks the port in the case of http.

richarddavison: @nabetti1720 please revise!

PR #1111, opened 2025-08-24
feat: add in support for node: prefix for remaining packages

willfarrell:
### Description of changes

Allow `node:` import prefix to be used for all currently integrated node modules (`node:console` is already supported). 

Notes:
- update node modules unit tests to test all ways of importing (as found in unit tests currently)
- merged `path` unit tests for `win`/`unix` together as they were already drifting apart (will wait for CI to ensure Windows tests pass and make adjustments if needed)
- move `node:url` unit tests into own file
- move `Blob` & `File` in to `node:buffer` unit test file to be aligned to nodejs docs
- Noticed that `os`, `stream` `net`, `tty` don't have unit tests
- Noticed the test output doesn't support nested decriptions

### Checklist


richarddavison:
Thanks for the PR and you're right, they should be added as alternatives. However the current test approach is a bit strange. It should be enough to assert that the two imports import the same object rather than testing everything twice.
Something like
```
import assert from "node:assert"
import legacyAssert from "assert

assert(assert === legacyAssert)
```

willfarrell:
When going through the tests there was a lot of inconsistency. I wasn’t sure how under the hood things are working, so didn’t want to remove certain testing in case they were valuable. I agree testing the imported modules against each other at a high level makes the most sense. I can update when I’m back from vacation in a week.

Would you like all references (other test files, etc) to node modules to be updated to use the node: prefix while I’m at it?

richarddavison:
>  I can update when I’m back from vacation in a week.

Take your time and enjoy PTO 🌴

> Would you like all references (other test files, etc) to node modules to be updated to use the node: prefix while I’m at it?

Yes please! And also if you spot inconsistencies, corrected as time permits :)

willfarrell:
Update pushed :)

modules/llrt_child_process/src/lib.rs
```diff
@@ -659,7 +659,7 @@ mod tests {
                     ctx.clone(),
                     "test",
                     r#"
-                   import {spawn} from "child_process";
+                   import {spawn} from "node:child_process";
```
willfarrell, line 662: Causing CI to fail

```
thread 'tests::test_spawn_shell' (38647) panicked at modules/llrt_child_process/src/lib.rs:721:18:
called `Result::unwrap()` on an `Err` value: Exception(Exception { message: Some("Error loading module 'node:child_process'"), stack: Some("") })
```

modules/llrt_child_process/src/lib.rs
```diff
@@ -700,7 +700,7 @@ mod tests {
                     ctx.clone(),
                     "test",
                     r#"
-                    import {spawn} from "child_process";
+                    import {spawn} from "node:child_process";
```
willfarrell, line 703: Causing CI to fail

```
thread 'tests::test_spawn' (38646) panicked at modules/llrt_child_process/src/lib.rs:679:18:
called `Result::unwrap()` on an `Err` value: Exception(Exception { message: Some("Error loading module 'node:child_process'"), stack: Some("") })
```

richarddavison (approved):

PR #1117, opened 2025-08-31
fix(lib/utils)primordial::get() causes panic when executed in parallel

nabetti1720:
### Description of changes

- When running WPT, panics could occur frequently in `unwrap()` inside `premordial::get()`.
- If the UserData is not stored successfully, we have added a fallback routine that assumes it has already been registered and retrieves the UserData again.

### Checklist


richarddavison:
Maybe a semaphore/lock would be better here. I my initial observation was the user data API was thread safe but there might be issues there. I don't see how the new PR can solve issues as this is a sync API. If we fail to set user data because its already exists the first check should have returned it unless there is a bug in `rquickjs`

nabetti1720:
> Maybe a semaphore/lock would be better here. I my initial observation was the user data API was thread safe but there might be issues there. I don't see how the new PR can solve issues as this is a sync API. If we fail to set user data because its already exists the first check should have returned it unless there is a bug in `rquickjs`

Thank you for your comment. It's clear that `unwrap()` is causing the panic, but as you said, this PR doesn't solve the problem (even before applying this PR, it sometimes happens and sometimes doesn't). I would like to close this PR for now and take up this issue again.

PR #1127, opened 2025-09-02
Fix LLVM

richarddavison:
### Description of changes

There is a bug in latest Rusts nightly which causes conflict llvm version https://github.com/rust-lang/rust/issues/145981.

This PR addresses that.


nabetti1720:
Is there any reason why we should use nightly builds at this point?

richarddavison:
> Is there any reason why we should use nightly builds at this point?

Yes, we need it because we're building the std library + simd. I'm looking nightly until the issue is rust issue is fixed

Issue #1129, opened 2025-09-03
Wildcards in package.json are not handled correctly


nabetti1720:
I wrote the following import statement based on a blog post, and although it loaded correctly in other runtimes, it failed to load in LLRT.

```javascript
// reproduction.js
import { getQueryParam } from 'hono/utils/url'
console.log(getQueryParam);
```

```
% node reproduction.js    
[Function: _getQueryParam]

% deno run reproduction.js
[Function: _getQueryParam]

% bun reproduction.js 
[Function: _getQueryParam]

% llrt reproduction.js 
SyntaxError: Could not find export 'getQueryParam' in module '/Users/shinya/Workspaces/llrt-test/node_modules/hono/dist/index'
```

I checked the trace and it looks like it's trying to load a different asset :
```
% LLRT_LOG=trace llrt reproduction.js
...
2025-09-03T12:12:22.341Z TRACE llrt_core::modules::require::resolver: Try resolve 'hono/utils/url' from '/Users/shinya/Workspaces/llrt-test/reproduction.js' 
2025-09-03T12:12:22.341Z TRACE llrt_core::modules::require::resolver: require_resolve(x, y):(hono/utils/url, /Users/shinya/Workspaces/llrt-test/reproduction.js) 
2025-09-03T12:12:22.341Z TRACE llrt_core::modules::require::resolver: |  load_package_self(x, dir): (hono/utils/url, /Users/shinya/Workspaces/llrt-test) 
2025-09-03T12:12:22.342Z TRACE llrt_core::modules::require::resolver: |  split name and scope(name, scope): (url, hono/utils) 
2025-09-03T12:12:22.342Z TRACE llrt_core::modules::require::resolver: |  split name and scope(name, scope): (utils/url, hono) 
2025-09-03T12:12:22.342Z TRACE llrt_core::modules::require::resolver: |  split name and scope(name, scope): (., hono/utils/url) 
2025-09-03T12:12:22.342Z TRACE llrt_core::modules::require::resolver: |  load_node_modules(x, start): (hono/utils/url, /Users/shinya/Workspaces/llrt-test) 
2025-09-03T12:12:22.342Z TRACE llrt_core::modules::require::resolver: |  load_package_exports(x, dir): (hono/utils/url, /Users/shinya/Workspaces/llrt-test/node_modules) 
2025-09-03T12:12:22.342Z TRACE llrt_core::modules::require::resolver: |  split name and scope(name, scope): (url, hono/utils) 
[...]

PR #1132, opened 2025-09-06
fix(require):  correctly handle wildcards in package.json

nabetti1720:
### Issue # (if available)

Fixed: #1129

### Description of changes

This PR will allow wildcards in package.json exports to be resolved correctly.

```javascript
// reproduction.js
import { getQueryParam } from 'hono/utils/url'
console.log(getQueryParam);
```

```
% llrt reproduction.js 
[function: _getQueryParam]

% bun reproduction.js 
[Function: _getQueryParam]
```

### Checklist

### DIFF
--- a/llrt_core/src/modules/require/resolver.rs
+++ b/llrt_core/src/modules/require/resolver.rs
@@ -785,9 +785,8 @@ fn package_exports_resolve<'a>(
             }
             // Check for wildcard pattern
             if let Some(scope) = wildcard.1 {
-                let scope = scope.as_str();
                 // Check for exports -> scope -> platform(browser or node) -> [import | require]
-                if let Some(BorrowedValue::Object(name)) = exports.get(scope) {
+                if let Some(BorrowedValue::Object(name)) = exports.get(scope.as_str()) {
                     if let Some(BorrowedValue::Object(platform)) = name.get(LLRT_PLATFORM.as_str())
                     {
                         if let Some(BorrowedValue::String(ident)) = platform.get(ident) {
--- a/llrt_core/src/modules/require/resolver.rs
+++ b/llrt_core/src/modules/require/resolver.rs
@@ -745,11 +745,9 @@ fn package_exports_resolve<'a>(
         modules_name
     };
 
-    let n = modules_name.bytes().filter(|&b| b == b'/').count();
-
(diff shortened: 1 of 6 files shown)


llrt_core/src/modules/require/resolver.rs
```diff
@@ -744,6 +745,15 @@ fn package_exports_resolve<'a>(
...
 
+    let n = modules_name.bytes().filter(|&b| b == b'/').count();
+
+    let wildcard = if n > 1 {
+        let (_name, _scope, _) = get_name_and_scope(modules_name, n - 1);
+        (Some(_name), Some([_scope, "/*"].concat()))
+    } else {
+        (None, None)
+    };
+
```
richarddavison: Maybe better like this?

```rust
let wildcard  = if let Some(pos) = modules_name.rmatch_indices('/').nth(1) {
    let (name, mut scope, _) = get_name_and_scope(modules_name, pos.0);
    
    // reuse the existing scope string allocation
    scope.push_str("/*"); 
    
    (Some(name), Some(scope))
} else {
    (None, None);
}
```

llrt_core/src/modules/require/resolver.rs
```diff
@@ -753,59 +763,109 @@ fn package_exports_resolve<'a>(
...
+    Ok(("./index.js", None, true))
+}
+
+fn replace_star(scope: &str, name: &str) -> String {
+    if scope.contains('*') {
+        scope.replace("*", name)
+    } else {
+        scope.into()
+    }
 }
```
richarddavison: We can just call replace since it won't do any thing if there is no "*" and we still need to search in the entire string

```rust
fn replace_star(scope: &str, name: &str) -> String {
    scope.replace('*', name)
}
```

llrt_core/src/modules/require/resolver.rs
nabetti1720: Here the returned `scope` is a `&str`, not a `String`, so we couldn't reuse the `space`. I took your suggestion and tweaked it a bit.

nabetti1720:
Changes in the `uuid` package over the last few days seem to have us in a bit of a mess. Hopefully this will calm the resolver down...
https://github.com/uuidjs/uuid/releases/tag/v13.0.0

PR #1168, opened 2025-09-27
feat(llrt:qjs):  expose pseudo v8 module

nabetti1720:
### Description of changes

Some npm packages are built with the `node:v8` module in mind. For example, `mitata` and `@opensearch-project/opensearch`...
QuickJS also has a function to obtain memory/code information, so we will make it publicly available in the `llrt:qjs` module.
Although it is not fully compatible, to avoid being blocked by the lack of `node:v8` module, we have provided a small opt-in feature that allows you to use `node:v8` instead of `llrt:qjs` if any of the following conditions are met:
 - LLRT_PLATFORM=node
 - LLRT_PSEUDO_V8_STATS=1

### Checklist

### DIFF
--- a/llrt_core/src/environment.rs
+++ b/llrt_core/src/environment.rs
@@ -15,6 +15,9 @@ pub const ENV_LLRT_LOG: &str = "LLRT_LOG";
 //module
 pub const ENV_LLRT_PLATFORM: &str = "LLRT_PLATFORM";
 
+//llrt
+pub const ENV_LLRT_PSEUDO_V8_STATS: &str = "LLRT_PSEUDO_V8_STATS";
+
 //vm
 pub const ENV_LLRT_GC_THRESHOLD_MB: &str = "LLRT_GC_THRESHOLD_MB";
 
--- a/llrt_core/src/modules/llrt/mod.rs
+++ b/llrt_core/src/modules/llrt/mod.rs
@@ -1,6 +1,7 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 pub mod hex;
+pub mod qjs;
 pub mod util;
 pub mod uuid;
 pub mod xml;
--- a/llrt_core/src/modules/llrt/qjs.rs
+++ b/llrt_core/src/modules/llrt/qjs.rs
@@ -1,0 +1,129 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+use std::env;
+
+use once_cell::sync::Lazy;
+use rquickjs::{
+    module::{Declarations, Exports, ModuleDef},
+    prelude::Func,
+    qjs::{self},
+    Ctx, IntoJs, Object, Result, Value,
+};
+
+use crate::libs::utils::module::{export_default, ModuleInfo};
+use crate::modules::require::LLRT_PLATFORM;
+
+static LLRT_PSEUDO_V8_STATS: Lazy<String> = Lazy::new(|| {
+    env::var(crate::environment::ENV_LLRT_PSEUDO_V8_STATS)
+        .ok()
+        .filter(|flag| flag == "1")
+        .unwrap_or_else(|| "0".to_string())
+});
+
+fn is_v8_stats() -> bool {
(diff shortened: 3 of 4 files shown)


nabetti1720:
Are you having problems with the nightly builds again?

richarddavison:
> Are you having problems with the nightly builds again?

Yes, its panic flags. Simple fix tho!

richarddavison:
I'm like you a bit skeptical of this module since it's not actually v8 running under the hood. IMO `mitata` and OpenSearch modules should be more platform agnostic. On the other hand Bun also exposes a v8 module even though they're running CoreJS but they promise full Node.js compatibility.

(Btw, I was under the impression that mitata was platform agnostic? https://github.com/evanwashere/mitata?tab=readme-ov-file#universal-compatibility)

nabetti1720:
> I'm like you a bit skeptical of this module since it's not actually v8 running under the hood. IMO `mitata` and OpenSearch modules should be more platform agnostic. 

Thank you for your comment. I mentioned `mitata` and `@opensearch-project/opensearch` as examples that reference `node:v8`, but I think the approach will be very different going forward.

`mitata` seems to be taking a strategy of absorbing differences in runtime and engine. Therefore, if `llrt:qjs` or some other form of API that exposes quickjs statistics becomes available, we should encourage people to use it.

On the other hand, this is not the case with `@opensearch-project/opensearch`. It sticks to Node.js API, so leaving aside the question of what is "open"?, I don't think there is any way to avoid it at the moment. I haven't seen any use cases for this in other packages, but I suspect there are other packages that can only be circumvented in this way. There is no disagreement that the Opensearch API should be platform independent.

> On the other hand Bun also exposes a v8 module even though they're running CoreJS but they promise full Node.js compatibility.

I haven't looked into how this is achieved, but it certainly seems that way.

```javascript
// reproduction.js
import v8 from "node:v8";
console.log(v8);
```

```
% bun reproduction.js
{
  cachedDataVersionTag: [Function: cachedDataVersionTag],
  getHeapSnapshot: [Function: getHeapSnapshot],
  getHeapStatistics: [Function: getHeapStatistics],
  getHeapSpaceStatistics: [Function: getHeapSpaceStatistics],
  getHeapCodeStatistics: [Function: getHeapCodeStatistics],
  setFlagsFromString: [Function: setFlagsFromString],
  deserialize: [Function: deserialize],
  takeCoverage: [Function: takeCoverage],
  stopCoverage: [Function: stopCoverage],
  serialize: [Function: serialize],
  writeHeapSnapshot: [Function: writeHeapSnapshot],
  setHeapSnapshotNearHeapLimit: [Function: setHeapSnapshotNearHeapLimit],
  promiseHooks: {
    createHook: [Function],
    onInit: [Function],
    onBefore: [Function],
    onAfter: [Function],
    onSettled: [Function],
  },
  startupSnapshot: {
    addDeserializeCallback: [Function],
    addSerializeCallback: [Function],
    setDeserializeMainFunction: [Function],
    isBuildingSnapshot: [Function],
  },
  Deserializer: [class Deserializer],
  Serializer: [class Serializer],
  DefaultDeserializer: [class DefaultDeserializer extends Deserializer],
  DefaultSerializer: [class DefaultSerializer extends Serializer],
}
```

To reiterate, when considering what LLRT should be, I don't think it's necessary to aim for complete compatibility with the Node.js API. However, I believe that supporting use cases where users want to run packages that already have a proven track record in a serverless manner as much as possible will help expand the potential of LLRT.

nabetti1720:
Upon investigation, we found that `mitata` uses a dedicated API instead of `node:v8` to obtain Bun heap information.

```javascript
// mitata/src/main.mjs
...
    const tune = {
      $counters,
      inner_gc: 'inner' === this._gc,
      gc: !this._gc ? false : undefined,

      heap: await (async () => {
        if (globalThis.Bun) {
          const { memoryUsage } = await import('bun:jsc');
          return () => { const m = memoryUsage(); return m.current; };
        }

        try {
          const { getHeapStatistics } = await import('node:v8'); getHeapStatistics();
          return () => { const m = getHeapStatistics(); return m.used_heap_size + m.malloced_memory; }
        } catch { }
      })(),
...
```

nabetti1720:
I've come up with a more general mechanism for loading modules that LLRT won't support, so I'm closing this PR for now.

Issue #1116, opened 2025-08-29
LLRT crashes when axios is called with require


nabetti1720:
I know that LLRT is currently unavailable because `axios` requires a core module that is missing. However, I have found an issue where LLRT crashes, possibly for other reasons.

import: When you import `axios`, you will get an error that the `http` module is missing, but this is expected.

```javascript
// reproduction.js
import * as a from "axios";
console.log(a);
```

```
% LLRT_LOG=trace llrt reproduction.js
...
ReferenceError: Error resolving module 'http' from '/Users/shinya/Workspaces/llrt-test/node_modules/axios/lib/adapters/http.js'
```

require: If I call the same module with require, it crashes.
```javascript
// reproduction.js
const a = require("axios");
console.log(a);
```

```
% LLRT_LOG=trace llrt reproduction.js
...
2025-08-29T23:06:22.878Z TRACE llrt_core::modules::require::loader: Loading npm module: __cjs:/Users/shinya/Workspaces/llrt-test/node_modules/axios/./dist/browser/axios.cjs
 
zsh: trace trap  LLRT_LOG=trace llrt reproduction.js
```

The last module loaded is single-source and has no dependencies, so I tracked down which code was causing the crash and found the following code:

https://github.com/axios/axios/blob/a9f47afbf3224d2ca987dbd8188789c7ea853c5d/dist/browser/axios.cjs#L2782-L2795

```javascript
const supportsRequestStream = isReadableStreamSupported && test(() => {
  let duplexAccessed = false;

  const hasContentType = new Request(platform.origin, {
    body: new ReadableStream(),
    method: 'POST',
    get duplex() {
      duplexAccessed = true;
      return 'half';
    },
  }).headers.has('Content-Type');

  return duplexAccessed && !hasContentType;
});
```

However, if you take this part out along with the helper function written just before it and run it by itself, at least it won't crash.

```javascript
// reproduction.js
const isFetchSupported = typeof fetch === 'function' && typeof Request === 'function' && typeof Response === 'function';
const isReadableStreamSupported = isFetchSupported && typeof ReadableStream === 'function';

const test = (fn, ...args) => {
  try {
    return !!fn(...args);
  } catch (e) {
    return false
  }
};

const supportsRequestStream = isReadableStreamSupported && test(() => {
  let duplexAccessed = false;

  const hasContentType = new Request(platform.origin, {
    body: new ReadableStream(),
    method: 'POST',
    get duplex() {
      duplexAccessed = true;
[...]

Issue #1128, opened 2025-09-02
primordial::get() causes panic when executed in parallel


nabetti1720:
When running WPT, panics could occur frequently in `unwrap()` inside `premordial::get()`. This happens sometimes and sometimes it doesn't, making it a very unstable situation.

```　
streams/writable-streams > should pass aborting.any.js tests
undefined: undefined

----- LAST STDERR: -----

thread 'main' panicked at /Users/shinya/Workspaces/llrt/libs/llrt_utils/src/primordials.rs:58:35:
called `Option::unwrap()` on a `None` value
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

thread 'main' panicked at /Users/shinya/Workspaces/llrt/libs/llrt_utils/src/primordials.rs:58:35:
called `Option::unwrap()` on a `None` value
```

I've been trying to resolve this issue but haven't been able to. Please refer to #1117 for past efforts.

nabetti1720:
Resolved by #1182

PR #1182, opened 2025-10-01
fix: Require primordials to be initialized from module once to avoid data race

richarddavison:
### Issue # (if available)

Fixes https://github.com/awslabs/llrt/issues/1116 & https://github.com/awslabs/llrt/issues/1128

### Description of changes

Require the primordials to be initialized in the modules before use. The current userdata store is not thread safe as we might borrow user data from multiple threads. This PR ensures that userdata primordials are initialized in the module init. If not, a runtime exception is thrown.

### Checklist

### DIFF
--- a/API.md
+++ b/API.md
@@ -1,7 +1,9 @@
 # API documentation
 
 > [!NOTE]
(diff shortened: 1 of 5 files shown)


richarddavison:
This needs a rework in rquickjs as Userdata is not thread safe.

PR #1192, opened 2025-10-11
feat(core): Implement pseudo Node.js API import mechanism

nabetti1720:
### Issue # (if available)

Related #1168

### Description of changes

- This is a mechanism that can be used to emulate Node.js APIs that are not planned to be supported by LLRT or are unlikely to be supported.
- It is possible to create embedded modules in a similar way, but this will result in a larger bundle size.
- This feature is suitable for use cases where you want infrequently used modules to behave like Node.js APIs, but don't want to sacrifice bundle size by embedding them. However, loading performance will be slower than Rust native modules or embedded modules.

### Sample
```javascript
// reproduction.js
import { channel } from 'node:diagnostics_channel';

const testChannel = channel('test');

testChannel.subscribe((message, name) => {
  console.log(`channel[${name}] message:`, message);
});

testChannel.publish({ hello: 'world' });
```
```
% llrt reproduction.js 
ReferenceError: Error resolving module 'node:diagnostics_channel' from '/Users/shinya/Workspaces/llrt-test/reproduction.js'

% LLRT_PSEUDO_MODULE_DIR=/Users/shinya/Workspaces/llrt/example/modules/ llrt reproduction.js
channel[test] message: {
  hello: 'world'
}

% bun reproduction.js 
channel[test] message: {
  hello: "world",
}
```

In the future, I think v8 modules can be defined as follows:

```javascript
// v8.js
import { ComputeMemoryUsage } from "llrt:qjs";

export function getHeapStatistics() {
    const usage = ComputeMemoryUsage();

    return {
        total_heap_size: usage.memory_used_size,
        used_heap_size: usage.memory_used_size,
        heap_size_limit: usage.malloc_limit,
        malloced_memory: usage.malloc_size
    };
}
```

EDIT: Anything that can be written in JavaScript can be extended as a module. For example, we have added a simple HTTP server function. Please excuse the quality of the samples. :)

```javascript
// simple-server.js
import { createServer } from 'node:http';

const server = createServer((req, res) => {
  res.writeHead(200, { 'Content-Type': 'text/plain' })
  res.end('Hello Compatible Server')
})

server.listen(3000, () => {
  console.log('Server running at http://localhost:3000/')
})
```

```
% llrt simple-server.js
Server running at http://localhost:3000/
```

```
% bombardier -d 10s --fasthttp http://localhost:3000/

Bombarding http://localhost:3000/ for 10s using 125 connection(s)
[==================================================================================================================] 10s
Done!
Statistics        Avg      Stdev        Max
  Reqs/sec     24923.01    4574.01   33537.63
  Latency        5.13ms    11.29ms   398.57ms
  HTTP codes:
    1xx - 0, 2xx - 243407, 3xx - 0, 4xx - 0, 5xx - 0
    others - 0
  Throughput:     3.02MB/s
```

### Checklist

### DIFF
--- a/llrt_core/src/modules/pseudo/mod.rs
+++ b/llrt_core/src/modules/pseudo/mod.rs
@@ -1,6 +1,3 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 pub mod resolver;
-
-// added when .cjs files are imported
-const CJS_IMPORT_PREFIX: &str = "__cjs:";
--- a/llrt_core/src/modules/pseudo/resolver.rs
+++ b/llrt_core/src/modules/pseudo/resolver.rs
@@ -6,10 +6,8 @@ use once_cell::sync::Lazy;
 use rquickjs::{loader::Resolver, Ctx, Error, Result};
 use tracing::trace;
 
-use crate::modules::path;
+use crate::modules::{path, require::CJS_IMPORT_PREFIX};
 use crate::utils::io::JS_EXTENSIONS;
-
-use super::CJS_IMPORT_PREFIX;
 
 static LLRT_PSEUDO_MODULE_DIR: Lazy<Option<String>> =
     Lazy::new(|| env::var(crate::environment::ENV_LLRT_PSEUDO_MODULE_DIR).ok());


nabetti1720:
I've finished everything I wanted to do with this PR. I hope someone continues to develop the `http.js` as a workaround until a native implementation is available. :)

richarddavison:
I think this can be a nice workaround to increase compat for some packages that are nice to have but fully supporting them natively would bloat the runtime to much. Sorry for the delay, I've been traveling for work and have a huge backlog. Will review ASAP!

nabetti1720:
> I think this can be a nice workaround to increase compat for some packages that are nice to have but fully supporting them natively would bloat the runtime to much. Sorry for the delay, I've been traveling for work and have a huge backlog. Will review ASAP!

@richarddavison Thank you for your comment. There's no need to rush. :)

This time, I have proposed three compatibility modules, `diagnostics_channel`, `v8`, and `http` (which are far from completely compatible).

In my personal opinion, given the LLRT concept (running in a serverless environment like Lambda), there doesn't seem to be much value in implementing any of them natively.

As for `http`, the server functions are not suitable for use with Lambda, and the client functions can be replaced with fetch. That said, as many packages depend on the `http` module, I believe that compatibility functions should be provided, even if they are not native. I think that a Javascript pseudo-module or embedded module would be sufficient.

llrt_core/src/modules/pseudo/mod.rs
```diff
@@ -0,0 +1,6 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+pub mod resolver;
+
+// added when .cjs files are imported
+const CJS_IMPORT_PREFIX: &str = "__cjs:";
```
richarddavison: This string also exists in resolver elsewhere, it should not be duplicated. Also prefer static over const

llrt_core/src/vm.rs
```diff
@@ -113,6 +114,7 @@ impl Vm {
         let resolver = (
             module_resolver,
             EmbeddedResolver,
+            PseudoResolver,
```
richarddavison, line 117: Can we motivate the ordering here? Should it come before NpmResolver or file_resolver?

richarddavison (changes requested):
Some minor comments, thanks for the PR!

llrt_core/src/vm.rs
nabetti1720: The reason we thought it should be done before NpmResolver is that, when considering performance, we thought it would be possible to resolve the issue faster than NpmResolver, which has a complex search specification. The reason it is placed before file_resolver is because we believe that pseudo-modules should take precedence when there is a user-created js with the same name.

However, I think there is also the opposite way of thinking. For example, if both an Npm package and a pseudo module exist, one way of thinking is that the public Npm package should take priority. To put it in extreme terms, one idea is to set it at the end as a workaround in case you really can't solve the problem.

We will follow your preferences, so please let us know what you want. :)

kyubisation:
Hello @nabetti1720 👋 
Nice work! This will definitely help mitigating omitted functionality.

The following is more of a feature request than a critique;
The current implementation would not help with omitted functionality in an existing module, e.g. `existsSync` in `node:fs`.
It would also not be trivial to do this for self-contained binaries (my focus).
With this I wonder whether it would make sense to implement an API similar to Node.js [module.registerHooks(options)](https://nodejs.org/api/module.html#moduleregisterhooksoptions), which would allow more flexibility.

<details>
<summary>Hook code sample</summary>
 
```js
import { registerHooks } from "node:module";

registerHooks({
  resolve(specifier, context, nextResolve) {
    if (specifier === "fs" || specifier === "node:fs") {
      return {
        url: "llrt-polyfill:fs",
        format: "module",
        shortCircuit: true,
      };
    } else if (specifier.startsWith("llrt-original:")) {
      specifier = specifier.replace("llrt-original:", "node:");
    }
    return nextResolve(specifier, context);
  },
  load(url, context, nextLoad) {
    if (url === "llrt-polyfill:fs") {
      const code = `
        export * from "llrt-original:fs";
        import fs from "llrt-original:fs";
        export function existsSync(path) {
          try {
            fs.accessSync(path);
            return true;
          } catch {
            return false;
          }
        }
      `;

      return { format: "module", shortCircuit: true, source: code };
    }
    return nextLoad(url, context);
  },
});
```
</details>

Thoughts on this?

nabetti1720:
> The current implementation would not help with omitted functionality in an existing module, e.g. `existsSync` in `node:fs`.

You're right. We excluded modules that already have parts implemented in native Rust from this plan because we felt they were ready for the rest of the implementation and would be best enhanced with native implementations.

While I don't plan on doing so right now, I think it's possible to wrap a native module in a pseudo-module by using the following approach:

1. Modify it so that it can be resolved by the `llrt-original:` prefix, just like the `node:` prefix (this is easy to do since the current implementation also trims `node:`). Personally, I prefer the name `internal:` to `llrt-original`. :)
2. Change the order of resolvers to put the pseudo-module resolver implemented in this PR at the top.
3. Create and deploy the following pseudo module:

```javascript
// fs.js
export * from "llrt-original:fs";
import fs from "llrt-original:fs";
export function existsSync(path) {
  try {
    fs.accessSync(path);
    return true;
  } catch {
    return false;
  }
}
```

> It would also not be trivial to do this for self-contained binaries (my focus).

Certainly, they are not self-contained binaries. I'm ashamed to say that I didn't know about `module.registerHooks()`. And I'm very curious about it. I would like to actually use it with Node.js and understand it better. Please give me a little time. I don't intend to make this happen with this PR, but I'd like to try. :)

kyubisation:
Thank you for the response and consideration.
I did not mean to derail this PR and just wanted to check, whether you think my proposal was worthwhile.
Also `llrt-original:` was just meant as an example and I like your idea with `internal:`.

`module.registerHooks()` is fairly new (in contrast to [module.register(specifier[, parentURL][, options])](https://nodejs.org/api/module.html#moduleregisterspecifier-parenturl-options), which has been around for a while), so it is not very well known yet.
The advantage of that API is that it is separated into two; One for re-wiring import resolution (resolve) and one for actually loading the code, which allows for quite a bit of flexibility.

I can create a small example project, if that would help you?

nabetti1720:
> I can create a small example project, if that would help you?

Thank you. If you can provide it, please do. :)
I would appreciate it if you would report this suggestion as a new issue.

kyubisation:
I created a simple example: https://github.com/kyubisation/nodejs-hooks-example
You can clone the repository and run `npm test`

Another example would be my implementation in lit for a CSS hook (which uses the more verbose register function): https://github.com/lit/lit/blob/main/packages/labs/ssr-dom-shim/src/lib/css-hook.ts (which is registered here https://github.com/lit/lit/blob/main/packages/labs/ssr-dom-shim/src/register-css-hook.ts)

nabetti1720:
@richarddavison @kyubisation It's still rough around the edges, but `module.registerHooks()` is now working on my laptop. This PR seems like it can be replaced with `module.registerHooks()`, so I'm dropping it to draft for now.

nabetti1720:
I want to focus on #1218, so I'm closing this PR.

Issue #1191, opened 2025-10-09
Missing Request.formData() implementation?


perpil:
I've finally gotten my full api working using llrt, but ran into an issue parsing form data using hono. I'm not 100% sure what the underlying issue is, but I was able to workaround it by replacing my `parseBody` calls:

`await c.req.parseBody()`

with:

`Object.fromEntries(new URLSearchParams(await c.req.text()))`

I've created a [repro here](https://github.com/perpil/llrt-hono-repro)

When I run `parseBody`, I see the following stack:

```
2025-10-09T22:56:46.433Z	a2b604ad-0495-4cb1-b470-b634b08bde21	ERROR	TypeError: not a function
  at #cachedBody (/var/task/index.mjs:387:34)
      at formData (/var/task/index.mjs:402:12)
      at parseFormData (/var/task/index.mjs:59:34)
      at parseBody (/var/task/index.mjs:54:26)
      at parseBody (/var/task/index.mjs:370:58)
      at <anonymous> (/var/task/index.mjs:1761:15)
      at #dispatch (/var/task/index.mjs:839:39)
      at fetch (/var/task/index.mjs:865:27)
      at <anonymous> (/var/task/index.mjs:1529:34)
```

Is it possible that [formData](https://developer.mozilla.org/en-US/docs/Web/API/Request/formData) isn't properly implemented?

nabetti1720:
formData support is planned for #207 but not yet implemented.

gc-victor:
> Is it possible that [formData](https://developer.mozilla.org/en-US/docs/Web/API/Request/formData) isn't properly implemented?

@perpil, you can use polyfills. Here you have an example: [form-data.js](https://github.com/gc-victor/query/blob/main/crates/runtime/src/js/polyfills/form-data.js)

perpil:
Thanks all, I was unable to get the polyfill to work (probably due to skill issues) but I do have a workaround. Closing in favor of #207.

PR #1199, opened 2025-10-18
feat(fetch): expose FormData

### LINKED ISSUE #207: FormData Support

nabetti1720:
### Issue # (if available)

Closed #207
Related #1191

### Description of changes

- We will start by supporting the FormData interface, which can be used standalone.
- The test cases under fetch were confusing, so I organized them together.
- ~Support for the `formData()` method of Body/Request/Response should be included in the next enhancement.~
  Support for `formData()` in Body/Request/Response has been added in this PR.
- Added support for encoding/decoding multipart bytes. Initially implemented using `multer`, but switched to our own implementation to avoid dependency on `encoding_rs`.
- The new implementation now passes most WPT rules for `FormData`.

### Sample
```javascript
// reproduction.js
let formdata = new FormData();
formdata.append('foo', new Blob([JSON.stringify({ foo: "hoge", })], { type: "application/json" }));
formdata.append('bar', new Blob([JSON.stringify({ bar: "fuga", })], { type: "application/json" }));

let blob = await new Response(formdata).blob();
let body = await blob.text();
console.log(body); // multipart bytes

blob = new Blob([body.toLowerCase()], { type: blob.type.toLowerCase() });
let formdataWithLowercaseBody = await new Response(blob).formData();

console.log(formdataWithLowercaseBody.has("foo")); // true
console.log(formdataWithLowercaseBody.get("foo") instanceof File); // true

let foo = formdataWithLowercaseBody.get("foo");
console.log(foo.type); // "application/json"
```

```
% llrt reproduction.js
------WebKitFormBoundaryK9T8a9t9JrKBiAxTSO5I193P
Content-Disposition: form-data; name="foo"; filename="blob"
Content-Type: application/json

{"foo":"hoge"}
------WebKitFormBoundaryK9T8a9t9JrKBiAxTSO5I193P
Content-Disposition: form-data; name="bar"; filename="blob"
Content-Type: application/json

{"bar":"fuga"}
------WebKitFormBoundaryK9T8a9t9JrKBiAxTSO5I193P--

true
true
application/json
```

### Checklist

### DIFF
--- a/modules/llrt_fetch/src/request.rs
+++ b/modules/llrt_fetch/src/request.rs
@@ -209,17 +209,7 @@ impl<'js> Request<'js> {
     }
 
     async fn blob(&mut self, ctx: Ctx<'js>) -> Result<Blob> {
-        let mime_type = self
-            .headers()
-            .map(|headers| {
-                Headers::from_value(&ctx, headers.as_value().clone(), HeadersGuard::None)
-            })
-            .transpose()?
-            .and_then(|headers| {
-                headers
-                    .iter()
-                    .find_map(|(k, v)| (k == HEADERS_KEY_CONTENT_TYPE).then(|| v.to_string()))
-            });
+        let mime_type = self.get_header_value(&ctx, HEADERS_KEY_CONTENT_TYPE)?;
 
         if let Some(bytes) = self.take_bytes(&ctx).await? {
             return Ok(Blob::from_bytes(bytes, mime_type));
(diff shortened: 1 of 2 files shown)


modules/llrt_fetch/src/request.rs
```diff
@@ -226,6 +227,27 @@ impl<'js> Request<'js> {
...
+            .map(|headers| {
+                Headers::from_value(&ctx, headers.as_value().clone(), HeadersGuard::None)
+            })
+            .transpose()?
+            .and_then(|headers| {
+                headers
+                    .iter()
+                    .find_map(|(k, v)| (k == HEADERS_KEY_CONTENT_TYPE).then(|| v.to_string()))
+            })
+            .unwrap_or(MIME_TYPE_OCTET_STREAM.into());
```
richarddavison, line 223: We can extract the logic here to to get `Headers` as an object from a reference. So add this to a function and use here and for the blob method and for response as well.

(pseudo code)
```rust
fn get_headers(&self) -> Result<Headers>{
self
            .headers()
            .map(|headers| Headers::from_js(&ctx, headers.into_value()))
            .transpose()
}
```
We could even have a helper function to get a single header:
```rust
fn get_header(&self, ctx: Ctx<'js>, key: &str) -> Result<Option<(ImmutableString, ImmutableString)>>{
   self.get_headers()?
            .and_then(|headers| {
                headers
                    .iter()
                    .find(|(k, _)| (k == key)
            })
}
```

richarddavison (approved):

PR #1201, opened 2025-10-19
fix(test): register and run hooks in correct order

kyubisation:
### Description of changes

Currently if you define hooks in a describe block and nested describe blocks, they are not correctly executed.
This PR changes the hook registration and execution resolution and makes sure they are executed in the correct order.

### Checklist

### DIFF
--- a/llrt_core/src/modules/js/@llrt/test/index.ts
+++ b/llrt_core/src/modules/js/@llrt/test/index.ts
@@ -8,6 +8,7 @@ import { SocketReqMsg } from "./shared";
 import { platform } from "node:os";
 const IS_WINDOWS = platform() === "win32";
 import CircularBuffer from "./CircularBuffer";
+// @ts-ignore
 import { dimensions } from "llrt:util";
 
 type TestOptions = {
@@ -39,6 +40,7 @@ type RootSuite = TestProps & {
 };
 
 type WorkerData = {
+  writeInProgress: boolean;
   stdOutBuffer: CircularBuffer;
   stdErrBuffer: CircularBuffer;
   completed: boolean;
@@ -47,8 +49,8 @@ type WorkerData = {
   success: boolean;
   connectionTimeout: Timeout | null;
   currentTest: TestResult | null;
-  currentResult: SuiteResult | null;
-  currentFile: string | null;
+  result: SuiteResult | null;
+  file: string;
   currentPath: string[];
   currentTimeout: number;
 };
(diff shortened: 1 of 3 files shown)


richarddavison:
Hi @kyubisation thanks for your PR. I think there is some infinite loop or race conditions here in this code since tests seems to be suck in CI. These things are always a bit messy to debug. Best would be to run only tests for a few modules using the fuzzy mechanism (for instance `llrt test fs` to only tests files that are containing the word `fs`)

kyubisation:
I haven't had enough time to look at this in depth. What confused me is that this seems to happen consistently with unix nightlies, where all of the builds are stuck after successfully running tests.
I'm afraid I am not sure what the problem could be.

richarddavison:
> I haven't had enough time to look at this in depth. What confused me is that this seems to happen consistently with unix nightlies, where all of the builds are stuck after successfully running tests. I'm afraid I am not sure what the problem could be.

I can take a stab on its. It looks like a data race!

richarddavison:
@kyubisation I have identified the bug and its a sneaky one. First we fire the hooks now in wrong order, so we start all beforeAll hooks first and then run all afterAll. They should be running per test and not loop.
This caused an issue in the fetch tests where we basically created two servers and then tried to shutdown both. But we did overwrite the variable since `server` was a global variable leading bound ports opened that are impossible to close. This is a sample file to illustrate the problem. Try with node, this will not close and "hang":

```javascript
import net from "net";

let server = net.createServer((socket) => {
  socket.on("error", () => {}); //ignore errors as abort signals might cancel the socket
  socket.on("data", () => {
    socket.write(
      "HTTP/1.1 200 OK\r\nContent-Type: text/html\r\n\r\n<html></html>"
    );
    socket.end();
  });
});

server.listen(() => {});

server = net.createServer((socket) => {
  socket.on("error", () => {}); //ignore errors as abort signals might cancel the socket
  socket.on("data", () => {
    socket.write(
      "HTTP/1.1 200 OK\r\nContent-Type: text/html\r\n\r\n<html></html>"
    );
    socket.end();
  });
});
server.listen(() => {
  server.close();
});
```

This lead me to think about the test suite a bit differently. Rather than spawning workers that will request new modules to import, I now spawn one worker per test file (with a cap of workers = cpu count). If test reports complete and does not exit within 1 second, it's leaking resources and I throw an error to detected this. You could arguably just kill the worker but that prevents us from detecting resource leakage in code.

I'll push changes on this branch soon so stay in touch!

kyubisation:
@richarddavison Thank you very much for the analysis! I will await the changes. 👍 
Which reminds me; This is not directly related to this, but I noticed that if you abort (CTRL-C) a test run in LLRT, the workers will become orphaned but stick around.
I noticed because I implemented an HTTP server and run it in tests. When I aborted the test and tried to start it again, the system complained about the port already being in use.

kyubisation:
@richarddavison Nice work! Thank you very much. 😄

PR #1207, opened 2025-10-20
Ignore init for global clients not in us-east-1

perpil:
### Issue # (if available)

The roundtrips for instantiating clients with global endpoints added significant latency to Init Duration when not running in `us-east-1`. For example instantiating an `IAMClient` in `us-west-2` adds about 200 ms of latency to the cold start time. I believe this should only be done if the lambda is executing in the same region as the endpoint.

### Description of changes

Only make requests to `/sping` endpoint to prime global clients if they are in `us-east-1`. This code is not partition aware, let me know if I should attempt to fix this for the global endpoint logic as well.

I did not create tests since I'm not familiar with the rust ecosystem, but I did manually test in both us-west-2 and us-east-1 with the LLRT_LOG=trace turned to verify that it works correctly.

### Checklist

### DIFF
--- a/llrt_core/src/modules/embedded/loader.rs
+++ b/llrt_core/src/modules/embedded/loader.rs
@@ -207,16 +207,6 @@ fn init_client_connection(ctx: &Ctx<'_>, specifier: &str) -> Result<()> {
         endpoint
     };
 
-    let rt = unsafe { qjs::JS_GetRuntime(ctx.as_raw().as_ptr()) };
-    let rt_ptr = rt as usize; //hack to move, is safe since runtime is still alive in spawn
-
-    if check_client_inited(rt, endpoint) {
-        return Ok(());
-    }
-
-    let client = HTTP_CLIENT.as_ref().or_throw(ctx)?;
-
-    trace!("Started client init {}", client_name);
     let mut region = env::var("AWS_REGION").unwrap();
     let mut region_separator = ".";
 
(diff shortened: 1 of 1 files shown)


nabetti1720:
> I believe this should only be done if the lambda is executing in the same region as the endpoint.

If this is true, then most regional endpoints do not require an initialization process. Does this concept also apply to regional endpoints?

perpil:
> > I believe this should only be done if the lambda is executing in the same region as the endpoint.
> 
> If this is true, then most regional endpoints do not require an initialization process. Does this concept also apply to regional endpoints?

This is a good question. Put another way: what is the scenario where the extra round-trip to the service improves performance? Does it improve performance for the most common scenario or is it only in scenarios where you instantiate 2+ clients in the same region as the lambda that you see benefits? Maybe @richarddavison can comment on what scenarios the init logic improves performance.

richarddavison:
It improves performance for most scenarios as HTTP connection is initialized during INIT boost and is kept alive for consecutive calls (that happens inside the handler). So even if a global endpoint is called, it will be called later when we don't have INIT boost. So this means that we're shifting duration from init to handler.

However this is also memory dependent. The idea here is to utilize INIT boost. Lambda gives functions a CPU boost during initialization. However, this only applies up to 1 full vCPU so around 1.8 GB. More than that and "init priming" makes little sense.

So in your case, IAMClient in us-west-2 making a call inside the handler (with prming disabled) would be significantly slower if allocated function memory is low.

perpil:
I did an experiment with this code in `us-west-2`:

```javascript
import {IAMClient, GetRoleCommand} from "@aws-sdk/client-iam";
const client = new IAMClient({region: 'us-east-1'});

export const handler = async (event, context) => {
  try {
    await client.send(new GetRoleCommand({ RoleName: 'my-role' }));
  } finally {
    return {
      statusCode: 200,
      body: JSON.stringify({ message: 'Test completed' }),
    };
  }
};
```

Even at 128 MB, the roundtrip adds more time than the SSL setup with a fractional CPU.

With priming turned on, billed duration is 425 ms:

```
REPORT RequestId: 49bd8c8d-be21-496e-8693-6f886b8132de	Duration: 90.65 ms	Billed Duration: 425 ms	Memory Size: 128 MB	Max Memory Used: 32 MB	Init Duration: 334.09 ms
```

With priming turned off (LLRT_SDK_CONNECTION_WARMUP: '0') billed duration is 373 ms.

```
REPORT RequestId: f6754249-2ed6-4cfd-9d80-e4700c3441d8 Duration: 319.33 ms Billed Duration: 373 ms Memory Size: 128 MB Max Memory Used: 32 MB Init Duration: 53.51 ms
```

llrt_core/src/modules/embedded/loader.rs
```diff
@@ -232,6 +232,15 @@ fn init_client_connection(ctx: &Ctx<'_>, specifier: &str) -> Result<()> {
...
     ) {
+        if region != "us-east-1" {
+            trace!(
+                "Ignoring init for global client: {} because region {} not geographically close to us-east-1",
+                client_name,
+                region
+            );
+            mark_client_inited(rt_ptr as _);
+            return Ok(());
+        }
```
richarddavison, line 233: If we move this check to before `check_client_inited` we can just return without calling `mark_client_inited` . That saves us from running a mutex and decrementing an atomic counter.

perpil: Change made, thanks for the suggestion.

richarddavison:
Yeah looks right, global endpoint calls probably does not benefit from priming if you're too far away from the region.

PR #1209, opened 2025-10-22
Add basic Agent support

Sytten:
### Description of changes

I am adding the basic Agent support for fetch. I had a choice to follow the node-fetch/node:https spec or the undici/native node fetch spec, I decided to use the node-fetch/node:https. Open to discuss that if we prefer to use the dispatcher.

I started splitting the `tls` and `https` modules.

Eventually the Agent should replace the legacy hyper client as we will want more control over the pool.
For now my goal was to support the self-signed certificates with the `ca` and `rejectUnauthorized` props.

### Checklist

### DIFF
--- a/modules/llrt_tls/src/config.rs
+++ b/modules/llrt_tls/src/config.rs
@@ -65,8 +65,7 @@ pub fn build_client_config(
     let builder = match get_tls_versions() {
         Some(versions) => builder.with_protocol_versions(&versions),
         None => builder.with_safe_default_protocol_versions(),
-    }
-    .expect("TLS configuration failed");
+    }?;
 
     // Certificate verification
     let builder = if !options.reject_unauthorized {


libs/llrt_test_tls/src/lib.rs
```diff
@@ -0,0 +1,61 @@
...
+
+        // Bind to a random port on localhost.
+        let incoming = TcpListener::bind(&SocketAddr::from((Ipv4Addr::LOCALHOST, 0))).await?;
+        let addr = incoming.local_addr()?;
+
+        tokio::spawn(async move {
+            if let Err(e) = server::run(incoming, certs, shutdown_rx).await {
+                eprintln!("failed to run mock server: {e:#}");
+            }
+        });
```
richarddavison, line 39: Should we just print here?

libs/llrt_test_tls/src/lib.rs
```diff
@@ -0,0 +1,61 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+// This library is only needed until TLS is natively supported in wiremock.
```
richarddavison: nit: prefix with `XXX` or `FIXME` or `TODO`

modules/llrt_http/src/agent.rs
```diff
@@ -0,0 +1,83 @@
...
+    }
+}
+
+pub struct AgentOptions {
+    reject_unauthorized: Option<bool>,
+    ca: Option<Vec<Vec<u8>>>,
+}
+
+impl<'js> FromJs<'js> for AgentOptions {
+    fn from_js(_: &rquickjs::Ctx<'js>, value: rquickjs::Value<'js>) -> rquickjs::Result<Self> {
```
richarddavison: Why specify full path here? We already have `Value` and `Ctx` in a `use`

modules/llrt_http/src/agent.rs
```diff
@@ -0,0 +1,83 @@
...
+        let ty_name = value.type_name();
+        let obj = value
+            .as_object()
+            .ok_or(Error::new_from_js(ty_name, "Object"))?;
+
+        let reject_unauthorized = obj.get_optional::<_, bool>("rejectUnauthorized")?;
+        let ca = obj
+            .get_optional::<_, Either<String, Vec<String>>>("ca")?
```
richarddavison: Is this always a string or Array of strings?

modules/llrt_tls/src/no_verification.rs
richarddavison, line 1: nit: Maybe some use imports here and there would make this a bit more readable

modules/llrt_http/src/agent.rs
Sytten: In theory could also be a buffer, I can add it

richarddavison: ObjectBytes should be used I think

Sytten:
@richarddavison Ready for another review
I added a new AnyN util, let me know if you like it :)

libs/llrt_test_tls/src/lib.rs
```diff
@@ -0,0 +1,61 @@
...
+        let certs = MockServerCerts::load_default().await?;
+        let ca = String::from_utf8(tokio::fs::read(FileType::RootCert.default_path()).await?)?;
+
+        // Bind to a random port on localhost.
+        let incoming = TcpListener::bind(&SocketAddr::from((Ipv4Addr::LOCALHOST, 0))).await?;
+        let addr = incoming.local_addr()?;
+
+        tokio::spawn(async move {
+            if let Err(e) = server::run(incoming, certs, shutdown_rx).await {
+                println!("failed to run mock server: {e:#}");
```
richarddavison, line 37: Why do we just log here and not return the err?

libs/llrt_utils/src/any_of.rs
richarddavison, line 1: I like this! However not to confuse with Typescript `Any` we might name to `AnyOf` maybe?
But I also wonder where we would use this? Isn't this sort of semi defeat the purpose of pattern matching? And for cases we don't we have `Either`. Another option would maybe (if feasible) to have  something like an enumOf! macro so we can use pattern matching for incoming types? Im not to sure

modules/llrt_tls/src/config.rs
```diff
@@ -0,0 +1,98 @@
...
+) -> Result<ClientConfig, Box<dyn std::error::Error + Send + Sync>> {
+    let provider = Arc::new(ring::default_provider());
+    let builder = ClientConfig::builder_with_provider(provider.clone());
+
+    // TLS versions
+    let builder = match get_tls_versions() {
+        Some(versions) => builder.with_protocol_versions(&versions),
+        None => builder.with_safe_default_protocol_versions(),
+    }
+    .expect("TLS configuration failed");
```
richarddavison: Panic or return Error?

libs/llrt_utils/src/any_of.rs
Sytten: There are lots of place where we have many possible types as input so instead of doing:
```rust
let value: Either<String, Either<Vec<String>, Either<ObjectBytes, Vec<ObjectBytes>>>> = Either::Left("");
match value {
   Either::Left(a) => (),
   Either::Right(rest) => match rest {
       Either::Left(b) => ()
       Either::Right(rest) => match rest {
          Either::Left(c) => (),
          Either::Right(d) => (),
       }
    }
}
```
We do:
```rust
let value: AnyOf4<String, Vec<String>, ObjectBytes, Vec<ObjectBytes>> = AnyOf4::A("");
match value {
   AnyOf4::A(a) => (),
   AnyOf4::B(b) => (),
   AnyOf4::C(c) => (),
   AnyOf4::D(d) => (),
}
```

We dont need a special macro since its a regular enum

modules/llrt_tls/src/config.rs
Sytten: Ha yeah I can return error now

libs/llrt_test_tls/src/lib.rs
Sytten: Its a tokio spawn we cant return an error ;)

richarddavison: Doh! I totally missed that!

PR #1218, opened 2025-10-29
feat(module): expose module.registerHooks()

### LINKED ISSUE #1210: Support for module.registerHooks()
LLRT should support module:registerHooks() to achieve the following use cases:

- When you want to quickly create a completely new module.
- When you want to add missing functionality to an existing native module.

We expect the following implementation to work:

```javascript
// register-hook.js
import { registerHooks } from "node:module";

registerHooks({
  resolve(specifier, context, nextResolve) {
    if (specifier === "fs" || specifier === "node:fs") {
      return {
        url: "llrt-polyfill:fs",
        format: "module",
        shortCircuit: true,
      };
    } else if (specifier.startsWith("internal:")) {
      specifier = specifier.replace("internal:", "");
    }
    return nextResolve(specifier, context);
  },
  load(url, context, nextLoad) {
    if (url === "llrt-polyfill:fs") {
      const code = `
        export * from "internal:fs";
        import fs from "internal:fs";
        export function existsSync(path) {
          try {
            fs.accessSync(path);
            return true;
          } catch {
            return false;
          }
        }
      `;

      return { format: "module", shortCircuit: true, source: code };
    }
    return nextLoad(url, context);
  },
});
```
```javascript
// reproduction.js
import { existsSync } from 'node:fs';
console.log(existsSync("./register-hook.js"));
```

```
% llrt --import register-hook.js reproduction.js
true
```

~NOTE: The basic specifications follow `module.registerHooks`([url](https://nodejs.org/api/module.html#moduleregisterhooksoptions)), but we think it's better to specify the source for the hook using environment variables rather than the command line. Because it also needs to be taken into account that it will be executed by Lambda.~

nabetti1720:
### Issue # (if available)

Closed #1210

### Description of changes

This PR adds support for `module.registerHooks()`. To achieve this, we have made some improvements around Module, Resolver, and Loader.
- We decided to expose the `module` module, which was previously in core, from `llrt_modules`. The `module` module has basic functions such as resolver/loader, so it could not be exposed as a separate module.
- Resolvers/loaders that use register-hooks need to make partially resolved content available for subsequent ones, so we've integrated the resolver/loader functions for native implementations.
- Only basic functionality such as early return with `shortCircuit` and pre-resolving with `nextResolve/nextLoad` functions is supported.

~This PR has some conflicting assets with the earlier Agent Support, so it is left as a draft.~

### Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -595,9 +595,9 @@ dependencies = [
 
 [[package]]
 name = "crypto-common"
-version = "0.2.0-rc.4"
+version = "0.2.0-rc.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "6a8235645834fbc6832939736ce2f2d08192652269e11010a6240f61b908a1c6"
+checksum = "919bd05924682a5480aec713596b9e2aabed3a0a6022fab6847f85a99e5f190a"
 dependencies = [
  "hybrid-array",
 ]
@@ -691,9 +691,9 @@ dependencies = [
 
 [[package]]
 name = "digest"
-version = "0.11.0-rc.3"
+version = "0.11.0-rc.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "dac89f8a64533a9b0eaa73a68e424db0fb1fd6271c74cc0125336a05f090568d"
+checksum = "ea390c940e465846d64775e55e3115d5dc934acb953de6f6e6360bc232fe2bf7"
 dependencies = [
  "block-buffer",
  "const-oid",
@@ -1387,7 +1387,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "4b0f83760fb341a774ed326568e19f5a863af4a952def8c39f9ab92fd95b88e5"
 dependencies = [
  "equivalent",
- "hashbrown 0.15.5",
+ "hashbrown 0.16.0",
 ]
 
 [[package]]
@@ -2141,9 +2141,9 @@ checksum = "34080505efa8e45a4b816c349525ebe327ceaa8559756f0356cba97ef3bf7432"
 
 [[package]]
 name = "md-5"
-version = "0.11.0-rc.2"
+version = "0.11.0-rc.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f9ec86664728010f574d67ef01aec964e6f1299241a3402857c1a8a390a62478"
+checksum = "64dd2c9099caf8e29b629305199dddb1c6d981562b62c089afea54b0b4b5c333"
 dependencies = [
  "cfg-if",
  "digest",
@@ -2684,7 +2684,7 @@ dependencies = [
 [[package]]
 name = "rquickjs"
 version = "0.10.0"
-source = "git+https://github.com/DelSkayn/rquickjs.git#93a5ce59e4ed98ebe3638dbc9b82b614b1e7da43"
+source = "git+https://github.com/DelSkayn/rquickjs.git#7be0e73bbe96399d10702f73a782f7cfa6a21438"
 dependencies = [
  "either",
  "indexmap",
@@ -2695,7 +2695,7 @@ dependencies = [
 [[package]]
 name = "rquickjs-core"
 version = "0.10.0"
-source = "git+https://github.com/DelSkayn/rquickjs.git#93a5ce59e4ed98ebe3638dbc9b82b614b1e7da43"
+source = "git+https://github.com/DelSkayn/rquickjs.git#7be0e73bbe96399d10702f73a782f7cfa6a21438"
 dependencies = [
  "async-lock",
  "chrono",
@@ -2711,7 +2711,7 @@ dependencies = [
 [[package]]
 name = "rquickjs-macro"
 version = "0.10.0"
-source = "git+https://github.com/DelSkayn/rquickjs.git#93a5ce59e4ed98ebe3638dbc9b82b614b1e7da43"
+source = "git+https://github.com/DelSkayn/rquickjs.git#7be0e73bbe96399d10702f73a782f7cfa6a21438"
 dependencies = [
  "convert_case",
  "fnv",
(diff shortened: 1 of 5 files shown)


llrt_modules/src/require.rs
```diff
@@ -8,34 +8,22 @@ use std::{
...
-// added to force CJS imports in loader
-pub const CJS_LOADER_PREFIX: &str = "__cjsm:";
+use crate::module::ModuleNames;
+use crate::modules::{path::resolve_path, timers::poll_timers};
+use crate::package::resolver::require_resolve;
+use crate::CJS_IMPORT_PREFIX;
 
 pub static LLRT_PLATFORM: Lazy<String> = Lazy::new(|| {
-    env::var(environment::ENV_LLRT_PLATFORM)
+    env::var("LLRT_PLATFORM")
```
richarddavison: Was this a typo?

llrt/src/main.rs
```diff
@@ -47,6 +49,9 @@ use crate::base::{async_with, CatchResultExt};
 #[global_allocator]
 static ALLOC: snmalloc_rs::SnMalloc = snmalloc_rs::SnMalloc;
 
+static LLRT_REGISTER_HOOKS: Lazy<Option<String>> =
+    Lazy::new(|| env::var(ENV_LLRT_REGISTER_HOOKS).ok());
+
```
richarddavison: Do we need to cache it like this? What's the rationale?

richarddavison: Maybe this should be a CLI argument instead of an env? Can you elaborate a bit why env over CLI argument was implemented?

llrt_modules/src/module/loader.rs
```diff
@@ -0,0 +1,138 @@
...
+            trace!("|  Determined as `NormalCircuit`: {}", name);
+        }
+
+        trace!("+- Loading module: {}\n", name);
+        (load)(ctx.clone(), Vec::from(name))
+    }
+}
+
+pub fn module_hook_load<'js>(ctx: &Ctx<'js>, name: &str) -> Result<(bool, bool, Source<'js>)> {
+    let bind_state = ctx.userdata::<RefCell<ModuleHookState>>().or_throw(ctx)?;
```
richarddavison: I'm a little concerned about thread safety here. Can this be an Arc<RwLock<ModuleHookState>>> instead of a RefCell

llrt_modules/src/module/loader.rs
```diff
@@ -0,0 +1,138 @@
...
+
+    let source = result
+        .get_optional::<_, Source>("source")?
+        .unwrap_or(AnyOf2::A("".into()));
+
+    Ok((short_circuit, next_load, source))
+}
+
+#[allow(dependency_on_unit_never_type_fallback)]
+fn call_load_hooks<'js>(
```
richarddavison, line 96: Im a bit concerned what's going on in here. We create a new function for every iteration that calls itself recursively. Can we use a single next function that maintains an internal state which processes instead? That way we can probably let that borrow hooks or at least only clone them once.

llrt_modules/src/module/resolver.rs
```diff
@@ -0,0 +1,144 @@
...
+        .unwrap_or(false);
+
+    let next_resolve = result
+        .get_optional::<_, bool>("__nextResolve")?
+        .unwrap_or(false);
+
+    let url = result.get::<_, String>("url")?;
+
+    Ok((short_circuit, next_resolve, url))
+}
```
richarddavison, line 84: A bit same story here as the above comment. If you can solve that one elegantly then maybe you can create a similar concept with a generic internal state. This is basically how iteratos work.

richarddavison (changes requested):
Thanks for the PR and sorry for the delay. Added some comments, hope they make sense :) Let me know if you need more assistance!

llrt_modules/src/require.rs
nabetti1720: `environment::ENV_LLRT_PLATFORM` exists in llrt_core, but cannot be accessed from llrt_modules. In addition, since this environment variable is only used here, it has been hard-coded.

llrt/src/main.rs
nabetti1720: > Do we need to cache it like this? What's the rationale?

Indeed, considering that it will only be used once, there is no need to cache it.

> Maybe this should be a CLI argument instead of an env? Can you elaborate a bit why env over CLI argument was implemented?

As you say, in most use cases it makes more sense to specify it as a CLI argument. However, you should also assume that LLRT will run on Lambda. When running on Lambda, you cannot specify it as a CLI argument, so I used environment variables here.

I didn't have time to try it, but with a little more ingenuity, I might be able to make it work with both environment variables and CLI arguments.

llrt_modules/src/module/loader.rs
nabetti1720: I couldn't think of a good way to do this, so I had to take a somewhat forceful approach, organizing what needed to be done while interacting with the AI...

The tricky part here is that when `nextResolve()` or `nextLoad()` is called on the JS side, the registerHooks definitions need to be looked ahead and resolved.

Furthermore, if you discard the result of `nextResolve()` or `nextLoad()` and the hook ends with `shortCircuit: false`, you need to process the next hook.

If there is a better way, I'll be happy to consider it, so I'd be very grateful if you could provide me with some concrete code examples. :)

richarddavison:
@nabetti1720 i have made some minor optimizations on your branch. Will push later today!

nabetti1720:
> @nabetti1720 i have made some minor optimizations on your branch. Will push later today!

@richarddavison Thank you! I think it's much easier to read now.

Just let me confirm one thing. You've made some modifications to .cargo/config.toml, but it didn't work on my laptop (MacBook Air M2, 2022). So I've reverted it to the original setting.

If this is important, it would be greatly appreciated if you could do this in a separate PR. :)

nabetti1720:
> @nabetti1720 i have made some minor optimizations on your branch. Will push later today!

@richarddavison Thank you! I think it's much easier to read now.

Just let me confirm one thing. You've made some modifications to `.cargo/config.toml`, but it didn't work on my laptop (MacBook Air M2, 2022). So I've reverted it to the original setting.

If this is important, it would be greatly appreciated if you could do this in a separate PR. :)

richarddavison:
> > @nabetti1720 i have made some minor optimizations on your branch. Will push later today!
> 
> @richarddavison Thank you! I think it's much easier to read now.
> 
> Just let me confirm one thing. You've made some modifications to `.cargo/config.toml`, but it didn't work on my laptop (MacBook Air M2, 2022). So I've reverted it to the original setting.
> 
> If this is important, it would be greatly appreciated if you could do this in a separate PR. :)

That got included in the PR by misstake. It's new Apple Clang toolchains that mess with snmalloc build on latest version. I will move out to new PR if nessicierry!

PR #1228, opened 2025-11-03
Install Rust toolchain before adding target

jasongodev:
### Description of changes

In systems that uses rustup from the distro's repository (e.g. Debian 13, Arch Linux), the default toolchain may not be set during fresh install.

Screenshot_20251103_180238

Screenshot_20251103_170138

This will cause build errors with the current Makefile where adding a Rust target is done first before adding any toolchain that rustup can use.

The solution is to install a toolchain before adding a target. Doing this does not affect anything in the current build routine and is compatible with llrt's recommended installation of rustup via online shell script.

Note that rustup will suggest to set a default toolchain like `rustup default stable`. This is not needed nor mattered when building llrt. We can go straight install the nightly toolchain and the rest of the build commands in Makefile will be able to use this.

### Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -105,26 +105,26 @@ build: js
 
 ifeq ($(DETECTED_OS),windows)
 stdlib:
-	rustup target add $(TARGET_windows_x86_64)
 	rustup toolchain install $(RUST_VERSION) --target $(TARGET_windows_x86_64)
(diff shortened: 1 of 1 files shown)


richarddavison (approved):
Thanks! Nice catch!

PR #1240, opened 2025-11-18
feat: add dgram (UDP socket) module support

### LINKED ISSUE #1171: Support for `dgram` module
I know that LLRT does not aim to support all Node.js modules, but I can't help but hope for `dgram` support.

`dgram` is used in the following major packages (and their dependencies):
- `aws-embedded-metrics` via `@middy/cloudwatch-metrics`
- `aws-xray-sdk-core` via `@aws-lambda-powertools/tracer`

As a module that controls communications, I believe it is as important as the following modules that are already on the [roadmap](https://github.com/awslabs/llrt/issues/635):
- HTTP & HTTPs module
- TLS sockets/connections
- WebSockets

chessbyte:
### Issue # (if available)

Closes #1171

### Description of changes

This PR implements Node.js `dgram` module for UDP datagram socket support in LLRT.

**Motivation:** Enables support for AWS packages that depend on UDP functionality:
- `aws-embedded-metrics` (via `@middy/cloudwatch-metrics`)
- `aws-xray-sdk-core` (via `@aws-lambda-powertools/tracer`)

**Implementation highlights:**
- Full UDP socket support for IPv4 and IPv6 using Tokio's `UdpSocket`
- Event-driven API compatible with Node.js dgram
- Thread-safe state management with `Arc<AtomicBool>`
- Auto-bind functionality when sending without explicit bind
- Flexible argument parsing for `bind()` method supporting multiple calling conventions
- Background receive loop using async/await pattern

**APIs implemented:**
- `dgram.createSocket(type|options)` - Create UDP4/UDP6 sockets
- `socket.bind([port][, address][, callback])` - Bind to port/address (supports flexible arguments)
- `socket.send(msg, port, address[, callback])` - Send datagrams
- `socket.close([callback])` - Close socket
- `socket.address()` - Get socket address info
- `socket.unref()` / `socket.ref()` - Lifecycle management (API compatibility)

**Events supported:**
- `'message'` - Emitted when receiving datagrams (with Buffer and rinfo)
- `'listening'` - Emitted when socket starts listening
- `'close'` - Emitted when socket closes
- `'error'` - Emitted on errors

**Files changed:**
- New module: `modules/llrt_dgram/` (~400 lines)
- Tests: `tests/unit/dgram.test.ts` (138 lines, 11 test cases)
- TypeScript definitions: `types/dgram.d.ts` (381 lines)
- Build config: `build.mjs` (added dgram to external modules)
- Integration: minimal changes to 5 existing files

**Testing:**
- All 11 dgram tests pass locally
- Full test suite: 683 passed, 0 failed
- `make check` passes with no clippy warnings

### Checklist

  - Comprehensive test suite in `tests/unit/dgram.test.ts` (11 tests, all passing)
  - Local: `make check` passes with no warnings
  - Local: All clippy checks pass
  - All 683 unit tests pass
  - Complete TypeScript definitions in `types/dgram.d.ts`
  - Added reference to `types/index.d.ts`
  - Not needed for this refactor (no API changes, only internal implementation changes)

### DIFF
--- a/modules/llrt_dgram/src/socket.rs
+++ b/modules/llrt_dgram/src/socket.rs
@@ -260,7 +260,7 @@ impl<'js> Socket<'js> {
                                         result_tx.send(result.map_err(|e| e.to_string())).map_err(|_|Exception::throw_message(&send_ctx, "Failed to call callback in send, channel closed!"))?;
                                         continue;
                                     }
-                                    result?;
+                                    result.or_throw(&send_ctx)?;
                                 }
                             }
                         };


modules/llrt_dgram/src/socket.rs
```diff
@@ -0,0 +1,392 @@
...
+                This(this.clone()),
+                &ctx,
+                "listening",
+                cb,
+                true,
+                true,
+            )?;
+        }
+
+        let bind_addr = format!("{}:{}", address, port);
```
richarddavison: nit: [&address,&port].concat()

modules/llrt_dgram/src/socket.rs
```diff
@@ -0,0 +1,392 @@
...
+                                        Self::emit_str(
+                                            This(recv_class.clone()),
+                                            &ctx,
+                                            "error",
+                                            vec![error_val],
+                                            false,
+                                        )?;
+                                    }
+                                }
+                                break;
```
richarddavison: This is almost right, but not quite. Since we run in spawn_exit, we should return the error here (which eventually exits the program) unless we have an error listener. In case of error listener present, we sent the error val like you do here and return Ok

modules/llrt_dgram/src/socket.rs
```diff
@@ -0,0 +1,392 @@
...
+            }
+
+            // Emit 'close' event
+            Self::emit_str(This(socket_class), &ctx, "close", vec![], false)?;
+
+            Result::<()>::Ok(())
+        })?;
+
+        Ok(())
+    }
```
richarddavison: There is an issue here since we loop over receive in the receiver background task and await it: 
https://github.com/awslabs/llrt/pull/1240/files#diff-f2293a08ec812eb819df822013a6a939da94eaa8a9b0054d8557c4e6edb3fb5aR202-R207
In there we only check of we should close in next iteration of the loop. We should use a notify or broadcast with select! on the socket receive so we don't indenfinatly wait in there when we close.

modules/llrt_dgram/src/socket.rs
```diff
@@ -0,0 +1,392 @@
...
+        if let Some(family) = &borrow.local_family {
+            obj.set("family", family.clone())?;
+        }
+
+        Ok(obj)
+    }
+
+    pub fn unref(this: This<Class<'js, Self>>) -> Result<Class<'js, Self>> {
+        // In Node.js, unref() allows the process to exit if this is the only active handle
+        // In LLRT's context, this is a no-op but we keep it for API compatibility
```
richarddavison, line 529: a trace warn would be nice here, same below

tests/unit/dgram.test.ts
richarddavison, line 1: This is great! Can we add some rust tests as well? :)

richarddavison (changes requested):
Thanks a lot! This looks very good, just some minor changes

tests/unit/dgram.test.ts
chessbyte: @richarddavison addressed all your feedback (I hope) in second commit.

modules/llrt_dgram/src/socket.rs
```diff
@@ -0,0 +1,408 @@
...
+                                            }
+                                            let error_msg = format!("UDP receive error: {}", e);
+
+                                            if Err::<(), _>(Exception::throw_message(&ctx, &error_msg))
+                                                .emit_error("message", &ctx, recv_class.clone())?
+                                            {
+                                                // Error was handled by error listener, continue receiving
+                                            } else {
+                                                // No error listener, break loop
+                                                break;
```
richarddavison: We are swallowing the error here. Should propagate back to spawn_exit

chessbyte: @richarddavison oops.  Let's see if I got it right on 3rd commit.

chessbyte:
@richarddavison I believe this PR is ready for review again. (Commits changed because I rebased to latest main)

modules/llrt_dgram/src/socket.rs
```diff
@@ -0,0 +1,408 @@
...
+                ctx.clone().spawn_exit(async move {
+                    let mut buf = vec![0u8; 65536];
+
+                    let result: Result<()> = async {
+                        loop {
+                            tokio::select! {
+                                _ = close_rx.recv() => {
+                                    // Close signal received, exit loop
+                                    break;
+                                }
```
richarddavison: We should emit close event

modules/llrt_dgram/src/socket.rs
```diff
@@ -0,0 +1,408 @@
...
+                borrow.local_address = Some(local_addr.ip().to_string());
+                borrow.local_port = Some(local_addr.port());
+            }
+
+            is_bound.store(true, Ordering::SeqCst);
+
+            trace!("UDP socket bound to {}", local_addr);
+
+            // Emit 'listening' event
+            Self::emit_str(This(socket_class.clone()), &ctx, "listening", vec![], false)?;
```
richarddavison: Not really yet, we should emit inside ctx.clone().spawn_exit before the loop

modules/llrt_dgram/src/socket.rs
```diff
@@ -0,0 +1,408 @@
...
+                let recv_socket = socket_arc.clone();
+                let recv_class = socket_class.clone();
+                let recv_closed = is_closed.clone();
+                let recv_running = receiver_running.clone();
+                let mut close_rx = {
+                    let borrow = socket_class.borrow();
+                    borrow.close_tx.subscribe()
+                };
+
+                ctx.clone().spawn_exit(async move {
```
richarddavison: Since we are already in a spawned task, any reason we why spawn another task here?

modules/llrt_dgram/src/socket.rs
```diff
@@ -0,0 +1,408 @@
...
+            borrow.is_bound.clone()
+        };
+        let is_closed = {
+            let borrow = this.borrow();
+            borrow.is_closed.clone()
+        };
+        let receiver_running = {
+            let borrow = this.borrow();
+            borrow.receiver_running.clone()
+        };
```
richarddavison: Should we return here conditionally if we already have everything? Can we reopen if closed etc?

modules/llrt_dgram/src/socket.rs
```diff
@@ -0,0 +1,408 @@
...
+                {
+                    let mut borrow = socket_class.borrow_mut();
+                    borrow.socket = Some(socket_arc.clone());
+                    let local_addr = socket_arc.local_addr().or_throw(&ctx)?;
+                    borrow.local_address = Some(local_addr.ip().to_string());
+                    borrow.local_port = Some(local_addr.port());
+                    borrow.is_bound.store(true, Ordering::SeqCst);
+                }
+
+                socket_arc
```
richarddavison: Can this be shared with our bind function somehow? I also wonder if this is the most optimal way of doing this. In high performance apps we spawn a new task for every send which will be trashing the Tokio scheduler.

I suggestion a few things:
1. Create a mpsc (multi-producer, single-consumer) channel when the Socket is created. The send method simply pushes the payload into the channel (non-blocking).
2. A single long-running background task  reads from that channel and performs the actual socket.send_to. This can be skipped if already bound before (if used called bind). App should stay open until we call close

So consolidate with the bind function and we should be good to go :)

chessbyte:
@richarddavison - I want to make sure I understand all your feedback correctly before implementing:

Comments 1-3 (close/listening events, nested spawn_exit):

Is this the structure you're looking for?

```
spawn_exit {
    bind socket
    emit 'listening'
    loop {
        select! {
            close signal => emit 'close', break
            receive => handle message
        }
    }
}
```

Meaning:
- Remove the nested spawn_exit (line 206) so receive loop runs directly in the bind spawn_exit
- Emit 'listening' right before the loop starts (not immediately after bind)
- Emit 'close' when close signal received in select! block
- Remove 'close' event emission from close() method (line 369)

Comment 4 (bind state checks):
- Should bind() return early (success, no-op) if already bound?
- Should bind() return an error if the socket is closed?
- Can a closed socket be re-bound, or is it permanently closed?

Comment 5 (MPSC for sends):
- For "Can this be shared with our bind function somehow?" - should the auto-bind logic in send() (lines 301-315) be refactored to:
  - Extract common binding logic into a private helper function that both bind() and send() can call?
  - Have send() internally call bind() with default params if not already bound?
- Should I implement the MPSC architecture in this PR, or handle separately in a future PR?
- If implementing MPSC in this PR: should the channel be created in constructor or on-demand? Should sends queue if not bound, or error immediately?

richarddavison:
@chessbyte see below
> 
> Comments 1-3 (close/listening events, nested spawn_exit):
> 
> Is this the structure you're looking for?
> 
> ```
> spawn_exit {
>     bind socket
>     emit 'listening'
>     loop {
>         select! {
>             close signal => emit 'close', break
>             receive => handle message
>         }
>     }
> }
> ```
> 
> Meaning:
> 
> * Remove the nested spawn_exit (line 206) so receive loop runs directly in the bind spawn_exit

Yes

> * Emit 'listening' right before the loop starts (not immediately after bind)

Yes, because we're not actually receiving. It's a slight delay before the spawned task is actually executing.

> * Emit 'close' when close signal received in select! block

Yes

> * Remove 'close' event emission from close() method (line 369)

Yes, it should only emit close when the actual socket closes, not when we call close.

> Comment 4 (bind state checks):
> 
> * Should bind() return early (success, no-op) if already bound?

No, it should error: We should also return socket instance for bind and close. And we can't re-bind once closed, check this node.js example:

```shell
> const dgram = require("dgram")
undefined
> const socket = dgram.createSocket('udp6');
undefined
> socket.bind(1234)
<ref *1> Socket {
  _events: [Object: null prototype] {},
  _eventsCount: 0,
  _maxListeners: undefined,
  type: 'udp6',
  Symbol(shapeMode): false,
  Symbol(kCapture): false,
  Symbol(async_id_symbol): 171,
  Symbol(state symbol): {
    handle: UDP {
      lookup: [Function: bound lookup6],
      bind: [Function: bind6],
      connect: [Function: connect6],
      send: [Function: send6],
      Symbol(owner_symbol): [Circular *1]
    },
    receiving: false,
    bindState: 1,
    connectState: 0,
    queue: undefined,
    reuseAddr: undefined,
    reusePort: undefined,
    ipv6Only: undefined,
    recvBufferSize: undefined,
    sendBufferSize: undefined,
    receiveBlockList: undefined,
    sendBlockList: undefined
  }
}
> socket.bind(1234)
Uncaught Error [ERR_SOCKET_ALREADY_BOUND]: Socket is already bound
    at Socket.bind (node:dgram:274:11) {
  code: 'ERR_SOCKET_ALREADY_BOUND'
}
> socket.bind(1235)
Uncaught Error [ERR_SOCKET_ALREADY_BOUND]: Socket is already bound
    at Socket.bind (node:dgram:274:11) {
  code: 'ERR_SOCKET_ALREADY_BOUND'
}
> socket.close()
Socket {
  _events: [Object: null prototype] {},
  _eventsCount: 0,
  _maxListeners: undefined,
  type: 'udp6',
  Symbol(shapeMode): false,
  Symbol(kCapture): false,
  Symbol(async_id_symbol): 171,
  Symbol(state symbol): {
    handle: null,
    receiving: false,
    bindState: 2,
    connectState: 0,
    queue: undefined,
    reuseAddr: undefined,
    reusePort: undefined,
    ipv6Only: undefined,
    recvBufferSize: undefined,
    sendBufferSize: undefined,
    receiveBlockList: undefined,
    sendBlockList: undefined
  }
}
> socket.bind(1235)
Uncaught Error [ERR_SOCKET_DGRAM_NOT_RUNNING]: Not running
    at healthCheck (node:dgram:965:11)
    at Socket.bind (node:dgram:270:3) {
  code: 'ERR_SOCKET_DGRAM_NOT_RUNNING'
}
> socket.bind(1235)
Uncaught Error [ERR_SOCKET_DGRAM_NOT_RUNNING]: Not running
    at healthCheck (node:dgram:965:11)
    at Socket.bind (node:dgram:270:3) {
  code: 'ERR_SOCKET_DGRAM_NOT_RUNNING'
}
```

> * Should bind() return an error if the socket is closed?

Yes

> * Can a closed socket be re-bound, or is it permanently closed?

I think its permanently closed, check https://nodejs.org/api/dgram.html

> Comment 5 (MPSC for sends):
> 
> * For "Can this be shared with our bind function somehow?" - should the auto-bind logic in send() (lines 301-315) be refactored to:
>   
>   * Extract common binding logic into a private helper function that both bind() and send() can call?

Yes, something like "fn bind_udp(...){...}"

>   * Have send() internally call bind() with default params if not already bound?

Yes
> * Should I implement the MPSC architecture in this PR, or handle separately in a future PR?

Please do in this PR.

> * If implementing MPSC in this PR: should the channel be created in constructor or on-demand? Should sends queue if not bound, or error immediately?

It should be created in bind (on-demand), also use unbuffered channel.

This runs until close is sent. We can also use `emit_error` since socket implements `Emitter`. This automatically handles sending error if we have error listener. See usage here:
https://github.com/awslabs/llrt/blob/011cadca998d6ecc87f6f4cdeb030fd136fca521/modules/llrt_child_process/src/lib.rs#L286-L288

Hope It makes sense!

chessbyte:
@richarddavison 🤞 that we are getting closer to the finish line 🏁

modules/llrt_dgram/src/socket.rs
```diff
@@ -0,0 +1,468 @@
...
+            borrow.is_bound.clone()
+        };
+        let is_closed = {
+            let borrow = this.borrow();
+            borrow.is_closed.clone()
+        };
+        let receiver_running = {
+            let borrow = this.borrow();
+            borrow.receiver_running.clone()
+        };
```
richarddavison: We borrow and check above already, no need to do it again here

modules/llrt_dgram/src/socket.rs
```diff
@@ -0,0 +1,468 @@
...
+            }),
+            receiver_running: Arc::new(AtomicBool::new(false)),
+            close_tx,
+            send_tx: None,
+        };
+
+        Class::instance(ctx, instance)
+    }
+
+    pub fn bind(
```
richarddavison, line 304: This is much better! Some minor improvements:
- We don't have to repeat `let borrow = this.borrow();` in many blocks. If we only need immutable borrow we can do it once and return error etc.
- This function is very large, we can break it aport into smaller functions.
- In the outmost `spawn_exit`, we should borrow_mut once, set everything up and then drop the borrow. We don't have to borrow twice.
- From `send` we call this method using `Self::bind(This(this.clone()), ctx.clone(), Rest(vec![]))?;` but we already have access to the borrowed value and can use an inner method in bind using less checks. Something like:
```
bind(){
  checks and setup
  start_listening_task()
}
send(){
...
if needs to be auto bound {
  start_listening_task()
}
}
```

modules/llrt_dgram/src/socket.rs
```diff
@@ -0,0 +1,468 @@
...
+                                        let error_msg = format!("UDP send error: {}", e);
+                                        let _ = Err::<(), _>(Exception::throw_message(&send_ctx, &error_msg))
+                                            .emit_error("send", &send_ctx, send_class.clone());
+                                    }
+                                },
+                                None => {
+                                    // Channel closed, stop task
+                                    break;
+                                },
+                            }
```
richarddavison: This can be simplified

modules/llrt_dgram/src/socket.rs
```diff
@@ -0,0 +1,468 @@
...
+                                    },
+                                }
+                            }
+                        }
+                    }
+                    Ok(())
+                }.await;
+
+                recv_running.store(false, Ordering::SeqCst);
+                result?;
```
richarddavison: Use `emit_error` here, we don't want to crash if we have error handler.

modules/llrt_dgram/src/socket.rs
```diff
@@ -0,0 +1,468 @@
...
+            } else {
+                borrow.is_closed.store(true, Ordering::SeqCst);
+                // Send close signal to interrupt receive loop immediately
+                let _ = borrow.close_tx.send(());
+                false
+            }
+        };
+
+        if already_closed {
+            return Ok(this.0);
```
richarddavison: Is this same as node? Or does node error?

richarddavison (changes requested):
Almost there :) A few changes!

chessbyte:
@richarddavison addressed your feedback in latest commit.

modules/llrt_dgram/src/socket.rs
```diff
@@ -0,0 +1,451 @@
...
+
+        // Get or create send_tx channel
+        let send_tx = {
+            let borrow = this.borrow();
+            if let Some(tx) = &borrow.send_tx {
+                tx.clone()
+            } else {
+                // Auto-bind if not bound
+                drop(borrow);
+                Self::bind(This(this.clone()), ctx.clone(), Rest(vec![]))?;
```
richarddavison: I don't like calling bind here like this. I think the actual binding should be a separate method that both bind and send calls. We also might want to wait until we actually have bounded before we start sending messages to the channel. Does `send_tx.send` below really work now? My concern is that since we start listening in a background task which has not started yet, we sent to a channel which no one is listening to. We probably need to do something like:
Self::listen(this.clone(), ||{
           send_tx.send((bytes, dest_addr, callback.0))
            .map_err(|_| Exception::throw_message(&ctx, "Failed to send message"))?;
            Ok(())
})

richarddavison (changes requested):
Almost there, just a minor change :)

chessbyte:
@richarddavison made the following changes based on your feedback:

1. Extracted start_listening() method:
    - Private helper method containing all the binding logic
    - Sets up UDP socket, starts receive loop, creates send channel, and spawns background tasks
    - Both bind() and send() now call this method
2. Created ensure_listening() method:
    - Ensures the socket is bound (auto-binds if needed)
    - Waits for send_tx channel to be initialized before returning
    - Uses spin-wait to guarantee the background task has progressed
3. Refactored bind() method:
    - Simplified to just parse arguments and call start_listening()
4. Refactored send() method:
    - No longer calls public bind() method
    - Calls ensure_listening() before attempting to send

richarddavison:
> * spin-wait

@chessbyte  spin-wait is not the solution here, we should use latch (llrt_core/src/utils/latch.rs). Other than that, I'm happy with this PR and it will be good to merge! Well done!

chessbyte:
@richarddavison Quick question on the Latch approach:

I can do any of the options below.  Which do you prefer?

- Import from llrt_core - Use llrt_core::utils::latch::Latch directly, but this creates a cyclic dependency (llrt_core → llrt_modules → llrt_dgram → llrt_core)
- Move Latch to llrt_utils - Both llrt_core and llrt_dgram would import from llrt_utils::latch::Latch (DRY, no cycles)
- Duplicate Latch in llrt_dgram - Keep a local copy to avoid touching llrt_core in this PR

richarddavison:
> @richarddavison Quick question on the Latch approach:
> 
> I can do any of the options below. Which do you prefer?
> 
> * Import from llrt_core - Use llrt_core::utils::latch::Latch directly, but this creates a cyclic dependency (llrt_core → llrt_modules → llrt_dgram → llrt_core)
> * Move Latch to llrt_utils - Both llrt_core and llrt_dgram would import from llrt_utils::latch::Latch (DRY, no cycles)
> * Duplicate Latch in llrt_dgram - Keep a local copy to avoid touching llrt_core in this PR

Move Latch to llrt_utils would be best!

chessbyte:
> Move Latch to llrt_utils would be best!

@richarddavison added commit to implement the above.

chessbyte:
@richarddavison thanks for merging my first PR into this repo!  For selfish reasons, I really want to see this project take off!  Is there something I can work on next?  TLS or HTTP or something else?

PR #1241, opened 2025-11-18
Improve support for dns lookup

Sytten:
### Description of changes

- Adds support for `all` and `order`
- Refactored the code to not depend on unrelated cached dns crate

### Checklist


richarddavison (approved):
LGTM! Thanks a lot!

Sytten:
Kinda weird that the DNS lookup fails sometimes for google.com. @richarddavison Do you have a preferred domain I could use for testing that belongs to aws

nabetti1720:
> Kinda weird that the DNS lookup fails sometimes for google.com. @richarddavison Do you have a preferred domain I could use for testing that belongs to aws

I don't like using external domains for functional testing, how about looking up `localhost` instead?

Sytten:
@nabetti1720 We can't easily test ipv6 and ipv4 mixed with localhost, ideally this is an external domain or something we add to the /etc/hosts.

richarddavison:
> google.com

What about `amazon.com` 😏?

Sytten:
@richarddavison There is something fishy with the getaddrinfo used by rust, it doesnt seem to return ipv6 on my mac reliably. On Node they use ipv4 mapped addresses which is not great either. Whatever, this is an improvement.

PR #1243, opened 2025-11-20
feat(fs): implement `symlink` and `symlinkSync`

kyubisation:
### Description of changes

Implement `symlink` and `symlinkSync` in the `fs` module

### Checklist

### DIFF
--- a/modules/llrt_fs/src/lib.rs
+++ b/modules/llrt_fs/src/lib.rs
@@ -29,7 +29,7 @@ use self::read_dir::{read_dir, read_dir_sync, Dirent};
 use self::read_file::{read_file, read_file_sync};
 use self::rename::{rename, rename_sync};
 use self::rm::{rmdir, rmdir_sync, rmfile, rmfile_sync};
-use self::stats::{stat_fn, stat_fn_sync, Stats};
+use self::stats::{lstat_fn, lstat_fn_sync, stat_fn, stat_fn_sync, Stats};
 use self::symlink::{symlink, symlink_sync};
 use self::write_file::{write_file, write_file_sync};
 
@@ -53,6 +53,7 @@ impl ModuleDef for FsPromisesModule {
         declare.declare("rm")?;
         declare.declare("rmdir")?;
         declare.declare("stat")?;
+        declare.declare("lstat")?;
         declare.declare("constants")?;
         declare.declare("chmod")?;
         declare.declare("symlink")?;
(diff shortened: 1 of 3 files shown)


modules/llrt_fs/src/symlink.rs
richarddavison, line 1: I know this is draft but comment here anyway. I think we can move out a lot of the shared code between sync and async api.

Sytten: Agreed the code of tokio is basically a dumb wrapper around std + tokio::spawn_blocking. So IMO we do a blocking function and the async version spawns it on tokio blocking and the sync version uses it directly.
Also not unstable lib in the modules please, some of us are on stable rust and it needs to compile

kyubisation: This is currently a draft, because I had trouble running tests locally and I need to investigate why.

As previously mentioned, my Rust expertise is not the best, so I'm not sure what the best way to extract the sync/async functionality from shared functionality would be.
As for `junction_point`; I did not notice that this is nightly only. Should that code path just throw in that case?

richarddavison:
@kyubisation I did some minor refactoring. PTAL :)

kyubisation:
@richarddavison Thank you very much for your help again. I apologize for causing you more work. 🙇‍♂️

modules/llrt_fs/Cargo.toml
```diff
@@ -31,5 +31,8 @@ tokio = { version = "1", features = [
   "rt",
 ], default-features = false }
 
+[target.'cfg(windows)'.dependencies]
+junction = "1"
```
Sytten, line 35: Do we really need this lib or can we vendor the part we need?

richarddavison: No, it's apparently quite complex with a lot of edge cases. I suppose we could gate it so that "symlinks/junctions" are not suppored by default on windows but that's out of scope for this PR

richarddavison (approved):

kyubisation:
@richarddavison Thanks again for the refactoring. The code looks quite clean now. 👍

kyubisation:
@richarddavison Hello again!
I hope you are doing well and are not overwhelmed by this project.
Could I ask you for another release? Having access to `symlink`/`symlinkSync` would allow me to replace my current usage of the `ln` binary.

richarddavison:
Yes release will come this week, thanks for pointing out!

kyubisation:
No worries. Thank you very much! 🙇

kyubisation:
@richarddavison Is there a way I can help you with preparing the release?

richarddavison:
> @richarddavison Is there a way I can help you with preparing the release?

Sorry for the delay, some minor fixes I needed to get out first

PR #1246, opened 2025-11-22
fix(module): Correctly handle 'require' with parent directory specification

### LINKED ISSUE #1245: A request that specifies only the parent directory will fail
The following code is in the module `aws-embedded-metrics`:

```javascript
// aws-embedded-metrics/lib/logger/MetricsLoggerFactory.js
Object.defineProperty(exports, "__esModule", { value: true });
exports.createMetricsLogger = void 0;
const __1 = require("..");
const EnvironmentDetector_1 = require("../environment/EnvironmentDetector");
const MetricsContext_1 = require("./MetricsContext");
const createMetricsLogger = () => {
    const context = MetricsContext_1.MetricsContext.empty();
    return new __1.MetricsLogger(EnvironmentDetector_1.resolveEnvironment, context);
};
exports.createMetricsLogger = createMetricsLogger;
```

The current package resolver cannot resolve `require("..")`. Normally, when specified like this, it would need to look for `index.js` or similar in the parent directory.

nabetti1720:
### Issue # (if available)

Fixed #1245 

### Description of changes

When a command is specified as `require("..")`, it now searches for the specified file in the parent directory.

### Checklist

### DIFF
--- a/types/yarn.lock
+++ b/types/yarn.lock
@@ -1591,9 +1591,9 @@ isstream@~0.1.2:
   integrity sha512-Yljz7ffyPbrLpLngrMtZ7NduUgVvi6wG9RJ9IUcyCd59YQ911PBJphODUcbOVbqYfxe1wuYf/LJ8PauMRwsM/g==
 
 js-yaml@^4.1.0:
-  version "4.1.0"
-  resolved "https://registry.yarnpkg.com/js-yaml/-/js-yaml-4.1.0.tgz#c1fb65f8f5017901cdd2c951864ba18458a10602"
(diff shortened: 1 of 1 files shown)


llrt_modules/src/package/resolver.rs
```diff
@@ -144,7 +144,7 @@ pub fn require_resolve<'a>(
 
     let x_is_absolute = path::is_absolute(x);
     let x_starts_with_current_dir = x.starts_with("./");
-    let x_starts_with_parent_dir = x.starts_with("../");
+    let x_starts_with_parent_dir = x.starts_with("..");
```
richarddavison, line 147: Ahhh sneaky one! Thanks for spotting!

richarddavison (approved):
Fantastic!!!!

richarddavison (approved):

PR #1271, opened 2025-12-03
feat: add util.inspect with full options support

### LINKED ISSUE #1257: Support for `util.inspect` and `util.inspect.custom`
From what I've observed, some packages on `@opentelemetry` are built on the premise of `util.inspect`.

- @opentelemetry/resources
- @opentelemetry/sdk-metrics
- @opentelemetry/sdk-trace-base
- @opentelemetry/resource-detector-aws
- @opentelemetry/resource-detector-container

To avoid any misunderstanding, I believe that the components necessary for implementation already exist in `llrt_logging`, and simply implementing the following in the `util` module will work.

However, it completely ignores the options, so the output is not what you expect.

```rust
fn inspect<'js>(ctx: Ctx<'js>, args: Rest<Value<'js>>) -> Result<String> {
    let mut result = String::new();
    let mut options = llrt_logging::FormatOptions::new(&ctx, false, true)?;
    llrt_logging::build_formatted_string(&mut result, &ctx, args, &mut options)?;
    result.push(NEWLINE);
    Ok(result)
}

impl ModuleDef for UtilModule {
    // ...
    fn evaluate<'js>(ctx: &Ctx<'js>, exports: &Exports<'js>) -> Result<()> {
        export_default(ctx, exports, |default| {
            let globals = ctx.globals();
            // ...
            let inspect_fn = Function::new(ctx.clone(), inspect)?.with_name("inspect")?;
            inspect_fn.prop("custom", primordials.symbol_custom_inspect.clone())?;
            default.set("inspect", inspect_fn)?;
            Ok(())
        })
    }
}
```

I find it difficult to support the `util.inspect` options without breaking the existing `llrt_logging` implementation, so I hope that someone knowledgeable will take on the challenge of implementing it. :)

chessbyte:
### Issue # (if available)

Fixes #1257

### Description of changes

Implements util.inspect() function with the following features:
- depth: control recursion depth for nested objects (supports null for infinite)
- colors: ANSI color output support
- showHidden: show non-enumerable properties
- customInspect: call Symbol.for('nodejs.util.inspect.custom') methods
- maxArrayLength: limit array/set/map elements displayed
- maxStringLength: truncate long strings
- sorted: sort object keys (boolean or custom comparator function)
- breakLength: line length at which to break to multiple lines
- compact: control inline vs multiline formatting (number or false)

Also exports:
- util.inspect.custom symbol (nodejs.util.inspect.custom)
- util.inspect.defaultOptions object

### Checklist

### DIFF
--- a/API.md
+++ b/API.md
@@ -382,6 +382,8 @@ _Also available globally_
 
 [inherits](https://nodejs.org/api/util.html#utilinheritsconstructor-superconstructor)
 
+[inspect](https://nodejs.org/api/util.html#utilinspectobject-options)
+
 [TextDecoder](https://nodejs.org/api/util.html#class-utiltextdecoder)
 
 [TextEncoder](https://nodejs.org/api/util.html#class-utiltextdecoder)
--- a/libs/llrt_logging/src/builder.rs
+++ b/libs/llrt_logging/src/builder.rs
@@ -1,0 +1,586 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+//! IR Builder - converts JavaScript values to PrintIR.
+//!
+//! This module handles the first phase of the two-phase approach:
+//! traversing JavaScript values and building an intermediate representation
+//! that captures all the information needed for rendering.
+
+use std::collections::HashSet;
+
+use llrt_utils::{class::get_class_name, error::ErrorExtensions, hash};
+use rquickjs::{
+    atom::PredefinedAtom,
+    function::This,
+    object::Filter,
+    promise::PromiseState,
+    qjs, Ctx,
+    Error::{self},
+    Function, Object, Result, Symbol, Type, Value,
+};
+
+use crate::ir::{NumberIR, ObjectKey, PrintIR, PromiseStateIR, TruncatedKind};
+
+use std::{mem, slice};
+
+/// Options controlling how the IR is built
+pub struct BuildOptions<'js> {
+    /// Maximum recursion depth for objects
+    pub max_depth: usize,
+    /// Maximum number of array/object elements to include
+    pub max_array_length: usize,
+    /// Maximum string length before truncation
+    pub max_string_length: usize,
+    /// Whether to include non-enumerable properties
+    pub show_hidden: bool,
+    /// Whether to call custom inspect functions
+    pub custom_inspect: bool,
+    /// Filter for object properties
+    pub object_filter: Filter,
+    /// Object.prototype for comparison
+    pub object_prototype: Object<'js>,
+    /// Symbol for custom inspect
+    pub custom_inspect_symbol: Symbol<'js>,
+    /// Object.getOwnPropertyDescriptor function
+    pub get_own_property_desc_fn: Function<'js>,
+    /// Sort mode for object keys
+    pub sorted: SortMode,
+    /// Custom sort comparator function
+    pub sort_comparator: Option<Function<'js>>,
+}
+
+/// Sort mode for object keys
+#[derive(Clone, Default)]
+pub enum SortMode {
+    #[default]
+    None,
+    Alphabetical,
+    Custom,
+}
+
+impl<'js> BuildOptions<'js> {
+    #[allow(clippy::too_many_arguments)]
+    pub fn new(
+        ctx: &Ctx<'js>,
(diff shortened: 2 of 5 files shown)


Sytten:
I have a vague memory that console.log is a limited version of inspect but that they share codebase (I might be wrong on that). Could there be some re-use?

richarddavison:
> I have a vague memory that console.log is a limited version of inspect but that they share codebase (I might be wrong on that). Could there be some re-use?

Yes they do and we should do so here I as well.

@chessbyte when I created console initially I build it a bit simplified and "just" appended the string recursively.

In order to conform more to Node.js and other runtime we should do a two phase approach instead. So instead of building the string from the get-go in a single pass we can store the console line in an internal representation enum of the actual values. Then we can render them to stdout/stderr using a renderer which actually prints the message. This way we can adjust when to line break depending on how many items that are in the object/array/string etc and format depending on terminal width etc. This is exactly how it's done in Node where printing e deeply nested object on a small window renders different output and "pretty-print" vs doing so on a larger screen. The performance impact of this should be quite small since we're not going to deep anyway and is still appending to a string buffer first before printing.

So in summery:
Build print PrintIR from JsObjects
Then use a renderer to output it that deals with newlines, indentation, expansion, collapsing, truncating etc

@Sytten thoughts?

chessbyte:
@richarddavison @Sytten hoping I addressed feedback in 2nd commit

Sytten:
Agreed this is the way to go

chessbyte:
@Sytten @richarddavison what is the status of this PR? It has been sitting since my changes on Saturday without any activity.

Sytten:
No offense but we cant just merge 2k lines of  vibe code just like that. We work on llrt when we can and we will need to maintain that code for the years to come, this will take a least a couple weeks to get merged.

chessbyte:
When I ran the open-source project ManageIQ for a decade and received PRs, I treated them as serendipitous gifts.  When I felt they were too complex, I would request that they be chopped into multiple smaller PRs.  When I felt that the code may be unstable, I would ask for more tests.  When I did not like the code organization, I would suggest something more suitable to the project.  That project is still going strong at manageiq.org :-)

Your feedback is a bit insulting to someone who is genuinely trying to help move along a project that achieved NONE of its targeted goals for 2025.

Perfect is the enemy of good enough!  Best of luck with your approach!

richarddavison:
> When I ran the open-source project ManageIQ for a decade and received PRs, I treated them as serendipitous gifts. When I felt they were too complex, I would request that they be chopped into multiple smaller PRs. When I felt that the code may be unstable, I would ask for more tests. When I did not like the code organization, I would suggest something more suitable to the project. That project is still going strong at manageiq.org :-)
> 
> Your feedback is a bit insulting to someone who is genuinely trying to help move along a project that achieved NONE of its targeted goals for 2025.

Some things takes a while to get right. We prioritize to get them right then to ship to soon. Most of these features have a lot of edge cases that needs to be handled.

> 
> Perfect is the enemy of good enough! Best of luck with your approach!

We do highly appreciate all PR and effort to make this project better and add more features. Thanks for the work you have done. However, significant changes like this will take a while to get through. I also see a couple of optimization paths we can do with this. Thanks for contributing and I encourage you to continue to do so!

chessbyte:
> Some things takes a while to get right. We prioritize to get them right then to ship to soon. Most of these features have a lot of edge cases that needs to be handled.

@richarddavison I would like nothing more than to add as many tests as there are edge conditions, so that we all feel comfortable about the implemented code.

> We do highly appreciate all PR and effort to make this project better and add more features. Thanks for the work you have done. However, significant changes like this will take a while to get through. I also see a couple of optimization paths we can do with this. Thanks for contributing and I encourage you to continue to do so!

@richarddavison I have no issue with PRs taking time to merge because you want to get big changes right.  Having said that, before this project reaches 1.0 status, I encourage you to iterate quicker and trust your tests to avoid regressions.

This particular PR introduces new functionality to address a reported issue of missing functionality.  It should not break any existing users of llrt as nothing should be using `util.inspect`, and, therefore, requires a bit less scrutiny, IMHO.  Regardless, I defer to you as the project maintainer.

richarddavison:
Sorry for the delay but we've been going back and forth on these gen AI PRs. In conclusion, there are to many changes with little to no manual review first so I will close in favor of smaller PRs or more thorough checks.

PR #1275, opened 2025-12-06
feat(fetch): implement fetch body streams

### LINKED ISSUE #946: Fetch does not support webstreams
Currently, fetch only support reading entire responses and requests from and into memory.
The streaming API's of Request and Response is not implemented.
Since the introduction of native streaming API's by @jackkleeman this functionality is now possible.
Spec is here:
https://fetch.spec.whatwg.org/#body
`body` should return a ReadableStream.

Currently we return `null`:
https://github.com/awslabs/llrt/blob/38af49052380ce2e1dd4b5bc1b93ce3771eba572/modules/llrt_http/src/response.rs#L316

chessbyte:
### Issue # (if available)

Closes #946

### Description of changes

Implements Response.body and Request.body as ReadableStream per WHATWG Fetch spec:

- Response.body returns ReadableStream<Uint8Array> for streaming response data
- Request.body returns ReadableStream<Uint8Array> for request bodies
- Byte stream controller (type: "bytes") with BYOB reader support
- Response.clone() uses tee() to allow cloning after body access
- fetch() accepts ReadableStream as request body for streaming uploads
- Compressed responses (gzip/deflate/brotli) stream decompressed data
- pipeTo() support for piping body streams to WritableStream

Added 42 new tests covering fetch body stream functionality.

### Checklist

### DIFF
--- a/modules/llrt_fetch/src/body_stream.rs
+++ b/modules/llrt_fetch/src/body_stream.rs
@@ -1,0 +1,706 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+//! Body stream adapters for converting Hyper bodies to ReadableStream
+//!
+//! This module provides utilities to create JavaScript ReadableStream objects
+//! from hyper's Incoming body type, enabling streaming of fetch response bodies.
+
+use std::cell::RefCell;
+use std::io::Read;
+use std::rc::Rc;
+
+use bytes::{Bytes, BytesMut};
+use http_body_util::BodyExt;
+use hyper::body::Body;
+use llrt_stream_web::ReadableStreamClass;
+use llrt_utils::bytes::ObjectBytes;
+use rquickjs::{
+    function::Constructor, prelude::This, Class, Ctx, Function, Object, Promise, Result,
+    TypedArray, Value,
+};
+
+use crate::Blob;
+use llrt_utils::mc_oneshot;
+use tokio::sync::mpsc;
+
+/// Supported content encodings for streaming decompression
+#[derive(Clone, Copy, Debug)]
+pub enum ContentEncoding {
+    Gzip,
+    Deflate,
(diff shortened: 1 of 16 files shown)


Sytten:
Yess!! You are shipping fast, we will try to review all pending PR.
Is the fetch spec using web streams?

nabetti1720:
> Is the fetch spec using web streams?

At least, it uses `llrt_stream_web` instead of `llrt_stream`, so that doesn't seem to be a problem.
Anyway, the work is fast! I'm looking forward to running the WPT once this PR is merged. :)

chessbyte:
@Sytten @nabetti1720 thanks for the compliment.  I was hoping that this project would reach 1.0 status by re:Invent, but that did not happen.  Instead of idly waiting, I  decided to help push the project forward.

modules/llrt_fetch/src/body_stream.rs
```diff
@@ -0,0 +1,706 @@
...
+    }
+
+    /// Add more compressed data and return any newly decompressed output available.
+    fn decompress_chunk(&mut self, chunk: Bytes) -> std::io::Result<Option<Vec<u8>>> {
+        // Add the new compressed data
+        self.state.input_buffer.extend_from_slice(&chunk);
+
+        // Try to decompress with current data
+        self.try_decompress()
+    }
```
richarddavison: This seems ineffective. Can we use a decoder that continue to feed bytes into? Since we know ContentEncoding in constructor it would be great if we can have the decoder constantly decode on the fly while working on chunks?

richarddavison:
Thanks for the PR but after an initial review there are to many errors, bugs and duplicate code in this PR. For instance, half of the errors are just sightly ignored, we handle both incoming and "streaming" data from hyper, we copy data left and right, decoder is not reused etc

Sytten:
@richarddavison Mind creating an issue for it though, I would like to have that. My plugins are somethings downloading big files and I would like to stream them to disk.

PR #1276, opened 2025-12-07
feat(timezone): add timezone support with minimal Intl.DateTimeFormat

### LINKED ISSUE #814: Performant way to do Timezone date conversions with LLRT missing Intl
First off, amazing work with LLRT, it's performance is amazing!
Currently the biggest use-case preventing me from using LLRT with more of my lambda's is the missing Intl library. I've played around with various polyfill libraries(@formatjs, Intl) and their performance is really bad. Adding in a simple timezone conversion from UTC to America/Denver adds over 1.2 seconds to the execution time.  I'm curious if adding Intl support is on the road map, I wasn't able to find it anywhere on some of the published roadmaps. If Intl is not on the road map I'm curious if anyone has found clever solutions to date timezone conversion. I actually don't need any localization, I just need to be able to use a library like dayjs and convert a date into the appropriate timezone and perform common operations like startOf('day')
Here is an example.
```
import dayjs from 'dayjs';
import timezone from 'dayjs/plugin/timezone';

const date3 = dayjs('2022-03-02T15:45:34Z').tz('America/Denver').startOf('day');

console.log(date3.toISOString()); //2022-03-02T07:00:00.000Z
```
Pretty much every date library I've looked at just uses Intl under the hood to perform this type of operation.

Let me know if you have any guidance for this use-case.

chessbyte:
### Issue # (if available)

Fixes #814 

### Description of changes

Add lightweight timezone support for LLRT that enables dayjs and similar libraries to perform timezone conversions without polyfills.

Features:
- Timezone.getOffset() - Get UTC offset for any IANA timezone with DST support
- Timezone.list() - List all available IANA timezone names
- Intl.DateTimeFormat - Minimal implementation with format(), formatToParts(), and resolvedOptions() for timezone-aware date formatting
- Date.prototype.toLocaleString - Enhanced to support timeZone option

This addresses issue #814 by enabling native timezone conversions that previously required polyfills adding 1.2+ seconds of execution time.

Binary size impact: +1.2 MB (IANA timezone database via chrono-tz)

### Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1420,9 +1420,9 @@ dependencies = [
 
 [[package]]
 name = "itoa"
-version = "1.0.15"
+version = "1.0.16"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "4a5f13b858c8d314ee3e8f639011f7ccefe71f97f96e50151fb991f267928e2c"
+checksum = "7ee5b5339afb4c41626dde77b7a611bd4f2c202b897852b4bcf5d03eddc61010"
 
 [[package]]
 name = "jobserver"
@@ -2752,7 +2752,7 @@ dependencies = [
 [[package]]
 name = "rquickjs"
 version = "0.10.0"
-source = "git+https://github.com/DelSkayn/rquickjs.git#ff3126177dad850bd7025a2681e06abe31d1e30d"
+source = "git+https://github.com/DelSkayn/rquickjs.git#111951b1a31075bdff46684e0ff29f37ff2a04b2"
 dependencies = [
  "either",
  "indexmap",
@@ -2763,7 +2763,7 @@ dependencies = [
 [[package]]
 name = "rquickjs-core"
 version = "0.10.0"
-source = "git+https://github.com/DelSkayn/rquickjs.git#ff3126177dad850bd7025a2681e06abe31d1e30d"
+source = "git+https://github.com/DelSkayn/rquickjs.git#111951b1a31075bdff46684e0ff29f37ff2a04b2"
 dependencies = [
  "async-lock",
  "chrono",
@@ -2779,7 +2779,7 @@ dependencies = [
 [[package]]
 name = "rquickjs-macro"
 version = "0.10.0"
-source = "git+https://github.com/DelSkayn/rquickjs.git#ff3126177dad850bd7025a2681e06abe31d1e30d"
+source = "git+https://github.com/DelSkayn/rquickjs.git#111951b1a31075bdff46684e0ff29f37ff2a04b2"
 dependencies = [
  "convert_case",
  "fnv",
@@ -2797,7 +2797,7 @@ dependencies = [
 [[package]]
 name = "rquickjs-sys"
 version = "0.10.0"
-source = "git+https://github.com/DelSkayn/rquickjs.git#ff3126177dad850bd7025a2681e06abe31d1e30d"
+source = "git+https://github.com/DelSkayn/rquickjs.git#111951b1a31075bdff46684e0ff29f37ff2a04b2"
 dependencies = [
  "bindgen",
  "cc",
@@ -2936,9 +2936,9 @@ checksum = "b39cdef0fa800fc44525c84ccb54a029961a8215f9619753635a9c0d2538d46d"
 
 [[package]]
 name = "ryu"
-version = "1.0.20"
+version = "1.0.21"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "28d3b2b1366ec20994f1fd18c3c594f05c5dd4bc44d8bb0c1c632c8d6829481f"
+checksum = "62049b2877bf12821e8f9ad256ee38fdc31db7387ec2d3b3f403024de2034aea"
 
 [[package]]
 name = "same-file"
@@ -3363,9 +3363,9 @@ checksum = "8df9b6e13f2d32c91b9bd719c00d1958837bc7dec474d94952798cc8e69eeec3"
 
 [[package]]
 name = "tracing"
-version = "0.1.43"
+version = "0.1.44"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "2d15d90a0b5c19378952d479dc858407149d7bb45a14de0142f6c534b16fc647"
+checksum = "63e71662fa4b2a2c3a26f570f037eb95bb1f85397f3cd8076caed2f026a6d100"
 dependencies = [
  "log",
  "pin-project-lite",
@@ -3374,9 +3374,9 @@ dependencies = [
 
 [[package]]
 name = "tracing-core"
-version = "0.1.35"
+version = "0.1.36"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "7a04e24fab5c89c6a36eb8558c9656f30d81de51dfa4d3b45f26b21d61fa0a6c"
+checksum = "db97caf9d906fbde555dd62fa95ddba9eecfd14cb388e4f491a66d74cd5fb79a"
 dependencies = [
  "once_cell",
 ]
--- a/llrt_core/src/runtime_client.rs
+++ b/llrt_core/src/runtime_client.rs
@@ -22,7 +22,6 @@ use rquickjs::{
     CaughtError, Ctx, Exception, Function, IntoJs, Module, Object, Result, Value,
 };
 use tracing::info;
-use zstd::zstd_safe::WriteBuf;
 
 use crate::libs::{
     json::{
@@ -302,7 +301,7 @@ async fn next_invocation<'js, 'a>(
 
     if res.status() != StatusCode::OK {
         let res_bytes = res.collect().await.or_throw(ctx)?.to_bytes();
-        let res_str = String::from_utf8_lossy(res_bytes.as_slice());
+        let res_str = String::from_utf8_lossy(&res_bytes[..]);
         return Err(Exception::throw_message(
             ctx,
             &["Unexpected /invocation/next response: ", &res_str].concat(),
@@ -380,7 +379,7 @@ async fn invoke_response<'js>(
         StatusCode::ACCEPTED => Ok(()),
         _ => {
             let res_bytes = res.collect().await.or_throw(ctx)?.to_bytes();
-            let res_str = String::from_utf8_lossy(res_bytes.as_slice());
+            let res_str = String::from_utf8_lossy(&res_bytes[..]);
             Err(Exception::throw_message(
                 ctx,
                 &["Unexpected /invocation/response response: ", &res_str].concat(),
@@ -545,7 +544,7 @@ async fn post_error<'js>(
     let res = client.request(req).await.or_throw(ctx)?;
     if res.status() != StatusCode::ACCEPTED {
         let res_bytes = res.collect().await.or_throw(ctx)?.to_bytes();
-        let res_str = String::from_utf8_lossy(res_bytes.as_slice());
+        let res_str = String::from_utf8_lossy(&res_bytes[..]);
         return Err(Exception::throw_message(
             ctx,
             &["Unexpected ", path, " response: ", &res_str].concat(),
--- a/modules/llrt_buffer/src/buffer.rs
+++ b/modules/llrt_buffer/src/buffer.rs
@@ -694,15 +694,40 @@ pub(crate) fn set_prototype<'js>(ctx: &Ctx<'js>, constructor: Object<'js>) -> Re
 }
 
 pub fn atob(ctx: Ctx<'_>, encoded_value: Coerced<String>) -> Result<rquickjs::String<'_>> {
-    //fine to pass a slice here since we won't copy if not base64
     let vec = bytes_from_b64(encoded_value.as_bytes()).or_throw(&ctx)?;
-    // SAFETY: QuickJS will replace invalid characters with U+FFFD
-    let str = unsafe { String::from_utf8_unchecked(vec) };
+    // Convert bytes to Latin-1 string where each byte becomes a character with that code point.
+    // This matches the WHATWG spec: atob returns a "binary string" where each character's
+    // code point is 0-255, directly representing one byte of data.
+    let str: String = vec.iter().map(|&b| b as char).collect();
     rquickjs::String::from_str(ctx, &str)
 }
 
-pub fn btoa(value: Coerced<String>) -> String {
-    bytes_to_b64_string(value.as_bytes())
+pub fn btoa(ctx: Ctx<'_>, value: Coerced<String>) -> Result<String> {
+    // Per WHATWG spec, btoa() treats input as a "binary string" where each character
+    // must have a code point 0-255. Characters > 255 cause InvalidCharacterError.
+    let s: &str = value.as_str();
+
+    // Fast path: ASCII is a 1:1 mapping to bytes 0-127 (SIMD optimized)
+    if s.is_ascii() {
+        return Ok(bytes_to_b64_string(s.as_bytes()));
+    }
+
+    // Slow path: Check for Latin-1 (0-255)
+    let bytes: Vec<u8> = s
+        .chars()
+        .map(|c| {
+            let code_point = c as u32;
+            if code_point > 255 {
+                Err(Exception::throw_message(
+                    &ctx,
+                    "Invalid character: btoa() argument contains character with code point > 255",
+                ))
+            } else {
+                Ok(code_point as u8)
+            }
+        })
+        .collect::<Result<Vec<u8>>>()?;
+    Ok(bytes_to_b64_string(&bytes))
 }
 
 #[cfg(test)]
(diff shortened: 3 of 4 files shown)


chessbyte:
Current Implementation vs Full Intl API

  What's Implemented ✅

  | Component           | Coverage | Notes                                                                       |
  |---------------------|----------|-----------------------------------------------------------------------------|
  | Intl.DateTimeFormat | ~40%     | format(), formatToParts(), resolvedOptions() - timezone-focused, en-US only |
  | Date.toLocaleString | Partial  | Only timeZone option supported                                              |

  What's Missing ❌

  | Component               | Complexity | Use Cases                                      |
  |-------------------------|------------|------------------------------------------------|
  | Intl.Collator           | High       | Language-sensitive string sorting/comparison   |
  | Intl.NumberFormat       | Medium     | Currency, percentages, units, compact notation |
  | Intl.PluralRules        | Low        | "1 item" vs "2 items" pluralization            |
  | Intl.ListFormat         | Low        | "A, B, and C" formatting                       |
  | Intl.RelativeTimeFormat | Medium     | "2 days ago", "in 3 hours"                     |
  | Intl.DisplayNames       | Medium     | "United States", "English" translations        |
  | Intl.Segmenter          | High       | Word/sentence/grapheme segmentation            |
  | Intl.Locale             | Low        | Locale parsing and manipulation                |
  | Intl.DurationFormat     | Medium     | "1 hr, 23 min" formatting                      |

  Gaps in Current DateTimeFormat

  | Feature                  | Status    | Required For                    |
  |--------------------------|-----------|---------------------------------|
  | Multiple locales         | ❌ Missing | Non-English formatting          |
  | Month names (long/short) | ❌ Missing | "March" vs "Mar"                |
  | Weekday names            | ❌ Missing | "Wednesday" vs "Wed"            |
  | Era formatting           | ❌ Missing | "AD", "BC"                      |
  | formatRange()            | ❌ Missing | "Jan 1 – 5, 2024"               |
  | formatRangeToParts()     | ❌ Missing | Range formatting parts          |
  | Calendar systems         | ❌ Missing | Buddhist, Islamic, Hebrew, etc. |
  | Numbering systems        | ❌ Missing | Arabic-Indic, Thai, etc.        |

  Implementation Effort Estimate

  | Level   | Components                                 | Effort     | Binary Size Impact  |
  |---------|--------------------------------------------|------------|---------------------|
  | Current | DateTimeFormat (minimal)                   | Done       | +1.2 MB             |
  | Level 1 | + NumberFormat, PluralRules, ListFormat    |Small  | +50-100 KB          |
  | Level 2 | + RelativeTimeFormat, DisplayNames, Locale | Small  | +200-400 KB         |
  | Level 3 | + Full DateTimeFormat (all locales)        | Medium  | +2-5 MB (CLDR data) |
  | Level 4 | + Collator, Segmenter                      | Large | +5-10 MB (ICU data) |

  Key Dependencies for Full Implementation

  1. CLDR Data - Unicode Common Locale Data Repository for locale-specific formatting rules, translations, and patterns
  2. ICU Library - For Collator and Segmenter (complex Unicode algorithms)
  3. Locale negotiation - BCP 47 language tag parsing and matching

nabetti1720:
Just a little tweet.

Only Node.js compatible modules were included in `llrt_modules`, and modules in the `llrt` namespace were deployed under `llrt_core/src/modules/llrt`.

Would it make sense to organize it like this, taking future expansion into account?

- `libs/llrt_date`: Prototype definitions for Date primitives. Other common libraries, etc. Something like `llrt_numbers` or `llrt_json`.
- `core/modules/llrt/timezone`: LLRT's own `timezone` module
- `modules/llrt_intl`: A subset of `Intl` functionality. Adding prototypes related to `Data` and future Int'l should be done in the `init()` of the `Intl` module.

chessbyte:
@nabetti1720 hopefully, I addressed your feedback around code organization in second commit that I pushed

llrt_core/src/modules/llrt/timezone.rs
```diff
@@ -0,0 +1,92 @@
...
+    timezone.set("getOffset", Func::from(get_offset))?;
+    timezone.set("list", Func::from(list_timezones))?;
+    timezone.set(PredefinedAtom::SymbolToStringTag, "Timezone")?;
+
+    Ok(timezone)
+}
+
+pub fn init(ctx: &Ctx<'_>) -> Result<()> {
+    let globals = ctx.globals();
+    globals.set("Timezone", timezone_object(ctx)?)?;
```
nabetti1720: We may not need to be accessible via Global.

types/index.d.ts
```diff
@@ -25,6 +25,7 @@
 /// <reference types="./stream.d.ts" />
 /// <reference types="./stream/web.d.ts" />
 /// <reference types="./timers.d.ts" />
+/// <reference types="./timezone.d.ts" />
```
nabetti1720, line 28: The module types in the llrt namespace have been described elsewhere so far.
llrt_core/src/modules/js/llrt.d.ts

I'm not sure what use it would be to define them here, though...

chessbyte:
> Thank you! I added some comments to the points that bothered me a little. None of them are a big deal.

@nabetti1720 Hoping I addressed these concerns in 3rd commit.

modules/llrt_intl/src/date_time_format.rs
```diff
@@ -0,0 +1,459 @@
...
+        for i in 0..parts.len() {
+            if let Ok(part) = parts.get::<Object>(i) {
+                if let Ok(value) = part.get::<_, String>("value") {
+                    result.push_str(&value);
+                }
+            }
+        }
+
+        Ok(result)
+    }
```
richarddavison, line 363: This is suboptimal. We're moving to and from JS. format_to_parts could call a function build a Rust Vec<(String,String)> which then gets converted to a JsArray. Similarly format could use the same function but just use the values and join them.

modules/llrt_intl/src/date_time_format.rs
```diff
@@ -0,0 +1,459 @@
...
+
+    /// Format a date to parts
+    #[qjs(rename = "formatToParts")]
+    pub fn format_to_parts<'js>(&self, ctx: Ctx<'js>, date: Opt<Value<'js>>) -> Result<Array<'js>> {
```
richarddavison, line 367: avoid extensive use of format! here. Use` [].concat()` and `dtoa` and `itoa` crates for fast number conversions. They are already used extensively in LLRT

modules/llrt_intl/src/date_time_format.rs
```diff
@@ -0,0 +1,459 @@
...
+            Ok(())
+        };
+
+        // Build parts based on options (following en-US format order)
+        // Month
+        if let Some(ref month_opt) = self.options.month {
+            let month_val = match month_opt.as_str() {
+                "2-digit" => format!("{:02}", local_dt.month()),
+                "numeric" => format!("{}", local_dt.month()),
+                _ => format!("{:02}", local_dt.month()),
```
richarddavison: This is repeated heavily and can be optimized without format

modules/llrt_intl/src/lib.rs
```diff
@@ -0,0 +1,91 @@
...
+}
+
+#[cfg(test)]
+mod tests {
+    #[test]
+    fn test_system_timezone() {
+        let tz = super::get_system_timezone();
+        assert!(!tz.is_empty());
+    }
+}
```
richarddavison, line 228: Some more rust tests would be nice

richarddavison (changes requested):
Thank you for this PR! Great work some minor optimization. Extensive use for format! can be removed in favor of more optimized ways. Also moving we can avoid moving from and to JS by using rust methods called from two places. And finally some duplication that could be removed :) 

Also review lock file, I think we might have duplicate dep of perfect hash map

chessbyte:
@richarddavison I tried to address your concerns in the 4th commit.

The lockfile has two versions of `phf` because:
- phf 0.12.1 - used by `chrono-tz`
- phf 0.13.1 - used by `llrt_encoding` and `rquickjs-core`

modules/llrt_intl/README.md
```diff
@@ -0,0 +1,58 @@
...
+dayjs.extend(timezone);
+
+const date = dayjs("2022-03-02T15:45:34Z");
+console.log(date.tz("America/Denver").format()); // "2022-03-02T08:45:34-07:00"
+console.log(date.tz("Asia/Tokyo").format()); // "2022-03-03T00:45:34+09:00"
+```
+
+## Note on ES Module imports
+
+When using dayjs with LLRT, use CommonJS `require()` syntax. ES module imports of dayjs plugins may not work correctly due to UMD wrapper handling differences.
```
richarddavison: is this correct?

modules/llrt_intl/src/lib.rs
```diff
@@ -0,0 +1,102 @@
...
+    date_proto.set("toLocaleString", Func::from(date_to_locale_string))?;
+
+    Ok(())
+}
+
+/// Custom Date.prototype.toLocaleString implementation with timezone support
+fn date_to_locale_string<'js>(
+    ctx: Ctx<'js>,
+    this: This<Value<'js>>,
+    _locale: Opt<Value<'js>>,
```
richarddavison: Used? Why is it here?

richarddavison (changes requested):
Just some minor questions and we should be good to go soon :) Thanks for this!

modules/llrt_intl/README.md
chessbyte: Removed this unverified ES Module imports note from README (in 4th commit)

modules/llrt_intl/src/lib.rs
chessbyte: The `locale` parameter is accepted but not used. It's required in the function signature to match the JavaScript API `toLocaleString(locales, options)`, ensuring the options object is correctly received as the second argument.

If desired, I can remove it.

chessbyte:
@richarddavison pushed another commit to address your feedback

modules/llrt_intl/src/lib.rs
richarddavison: It should be used since it determines the formatting:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toLocaleString

chessbyte:
> It should be used since it determines the formatting: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toLocaleString

@richarddavison Full support requires implementation of the following missing features:
1. Locale-aware formatting - date order (MDY vs DMY vs YMD), separators (/, ., -)
2. dateStyle/timeStyle - "full", "long", "medium", "short" presets
3. Localized text - month names, weekday names, era, AM/PM in different languages
4. Additional options - weekday, era, dayPeriod, fractionalSecondDigits, hourCycle

Key locale differences:
| Locale | Format             | Example                    |
|--------|--------------------|----------------------------|
| en-US  | MM/DD/YYYY, 12h    | 12/20/2012, 3:00:00 AM     |
| en-GB  | DD/MM/YYYY, 24h    | 20/12/2012, 03:00:00       |
| de-DE  | DD.MM.YYYY, 24h    | 20.12.2012, 03:00:00       |
| fr-FR  | DD/MM/YYYY, 24h    | 20/12/2012 03:00:00        |
| ja-JP  | YYYY/MM/DD, 24h    | 2012/12/20 3:00:00         |
| ko-KR  | YYYY. MM. DD., 12h | 2012. 12. 20. 오전 3:00:00 |

Options for implementation:

ICU4X (icu crate)
- Pros: Full CLDR compliance, actively maintained by Unicode Consortium, modular design, no_std support
- Cons: Even with data trimming, adds non-trivial size

Baked CLDR Patterns (custom)
- Extract just the date/time patterns from CLDR JSON
- Embed patterns for ~10-15 common locales directly in Rust
- Much smaller than full ICU4X: ~50-100KB for patterns + names (vs potentially MBs for full ICU4X)

Hybrid
- Use ICU4X with heavily trimmed data (only datetime, only target locales)
- Or build-time generation of baked patterns from CLDR

Given LLRT's minimal footprint goals, which approach do you prefer?  Or maybe an alternative approach not considered above.

richarddavison:
Hi, I think "Baked CLDR Patterns (custom)" would be better. Size is important here

chessbyte:
@richarddavison added 5th commit with the locale support.  It should be straightforward to add/remove locales as per the needs of this project.

## Binary Size Impact

| Metric | Size |
|--------|------|
| Main branch (no intl) | 9.38 MB |
| Feature branch (31 locales) | 10.61 MB |
| **Total feature addition** | **1.23 MB** |

### Component Breakdown
- chrono-tz timezone data: ~1.1 MB (597 IANA timezones)
- CLDR locale data (31 locales): ~158 KB
- **Per locale average: ~5.1 KB**

## Supported Locales (31)

### Tier 1: Core (14 locales) - Essential for global web apps
| Locale | Language | Coverage | Rationale |
|--------|----------|----------|-----------|
| en-US | English (US) | 380M native, 1.5B total | Primary fallback, most common |
| en-GB | English (UK) | 60M native | Different date format (dd/MM/yyyy) |
| de | German | 100M speakers | Major EU language |
| fr | French | 300M speakers | Official in 29 countries |
| es | Spanish | 550M speakers | 2nd most native speakers globally |
| it | Italian | 65M speakers | Major EU language |
| pt | Portuguese | 260M speakers | Brazil is 6th largest economy |
| nl | Dutch | 25M speakers | Netherlands/Belgium |
| ru | Russian | 150M speakers | Major global language |
| ja | Japanese | 125M speakers | 3rd largest economy |
| ko | Korean | 80M speakers | Major tech market |
| zh | Chinese (Simplified) | 1.1B speakers | Largest by native speakers |
| zh-TW | Chinese (Traditional) | 50M speakers | Taiwan/Hong Kong |
| ar | Arabic | 400M speakers | RTL, official in 26 countries |

### Tier 2: High Priority (8 locales) - Large user bases
| Locale | Language | Coverage | Rationale |
|--------|----------|----------|-----------|
| hi | Hindi | 600M speakers | India - massive developer market |
| bn | Bengali | 270M speakers | Bangladesh + India |
| vi | Vietnamese | 85M speakers | Growing tech hub |
| th | Thai | 60M speakers | Southeast Asia |
| id | Indonesian | 200M speakers | Largest SE Asian country |
| tr | Turkish | 80M speakers | Bridge between EU/Asia |
| pl | Polish | 45M speakers | Major EU tech market |
| uk | Ukrainian | 40M speakers | Large developer community |

### Tier 3: Medium Priority (9 locales) - EU/Regional coverage
| Locale | Language | Coverage | Rationale |
|--------|----------|----------|-----------|
| sv | Swedish | 10M speakers | Nordic coverage |
| da | Danish | 6M speakers | Nordic coverage |
| nb | Norwegian Bokmål | 5M speakers | Nordic coverage |
| fi | Finnish | 5M speakers | Nordic coverage |
| cs | Czech | 10M speakers | Central Europe |
| el | Greek | 13M speakers | Southern EU |
| he | Hebrew | 9M speakers | Israel tech hub |
| hu | Hungarian | 13M speakers | Central Europe |
| ro | Romanian | 26M speakers | Large EU country |

## Size Implications

- Adding 1 locale: ~5.1 KB increase
- Removing 1 locale: ~5.1 KB decrease

The CLDR data is highly compressed in the binary since:
- Static string slices are deduplicated
- Only date/time patterns + month/day names are included
- No number formatting data (handled separately)

nabetti1720:
I'm very excited about the 5th commit!
As a further improvement, if `cldr_data.rs` is separated under libs as `llrt_cldr` and any locale can be attached using `[features]`, I think this will enable both future extensibility and footprint optimization. The default definition in this case is undisputed to be 31 locales.

chessbyte:
@nabetti1720 Thanks for the feedback! I like the idea of feature-gated locales for flexibility.

That said, if we're adding optionality for CLDR data (~158KB), we should also consider the same approach for chrono-tz timezone data (~1.1MB) where the size savings are 7x greater. Options include:

1. Tiered timezone features - tz-minimal (major cities), tz-common, tz-all
2. chrono-tz's filter-by-regex - Build-time filtering to specific regions (e.g., America|Europe|Asia)
3. Full opt-out - Make timezone support entirely optional (UTC-only fallback)

Should I implement:
- (a) Just the CLDR feature flags in this PR
- (b) Just the TZ feature flags in this PR
- (c) Both CLDR and TZ optionality in this PR
- (d) Neither for now, save for a dedicated "size optimization" PR

Happy to go whichever direction makes sense for the project.

nabetti1720:
@chessbyte Thank you for your consideration!

Personally, I don't think we need to do everything in this PR, but I'll follow @richarddavison and @Sytten advice. What they say always points in the right direction.

Well, I think an approach to narrow down the time zone data of chrono-tz is necessary, but since the locale and time zone identifiers do not match, I think you need to think carefully about how to link them.
https://en.wikipedia.org/wiki/List_of_tz_database_time_zones

If that's possible, introducing the tier concept we developed and managing both locales and time zones with these feature gates might be a good way to keep the implementation clear.

```toml
#libs/llrt_cldr - Cargo.toml
[features]
default = ["tier1", "tier2", "tier3"]
```

richarddavison:
@nabetti1720 @chessbyte agree, I think we should limit this PR to add basic support.
For later enhancements maybe we can be a bit clever how we store the data related toIntl. We can probably use some form of compact representation in memory for most of the behavior + shared strings. I don't think chrono-tz timezone data (~1.1MB) was designed for size at all.

chessbyte:
> agree, I think we should limit this PR to add basic support.

@richarddavison please let me know if you want me to make any further changes to this PR and/or rebase.

richarddavison (approved):

PR #1284, opened 2025-12-12
fix(buffer): use Latin-1 encoding for atob() per WHATWG spec

### LINKED ISSUE #966: Error: PKCS#8 ASN.1 error: ASN.1 INTEGER not canonically encoded as DER
It seems that there is something strange in the crypto implementation that prevents verifying JWT tokens manually or with popular libraries like node-jose/aws-jwt-verify/etc.

Is it possible to get this issue fixed with the current crypto implementation?  Or is there a recommended workaround?

I'll share my reimplementation of JWT verification so I could verify where the problem is.  If you run it locally (deno for example) it will verify true.  If you run it on a llrt lambda, it will fail with the `Error: PKCS#8 ASN.1 error: ASN.1 INTEGER not canonically encoded as DER` error.  Again, this happens with other libraries as well. I put about 4 days into testing what was going on deeply.

The reason why this is popular is it can help to sidestep the ~300ms cold start penalty of using a cognito authorization lambda.  

Thank you in advance!  Love this work!

```
import {
  COGNITO_CLIENT_ID,
  COGNITO_USER_POOL_ID,
  REGION,
} from "@b/configuration.ts";
// const COGNITO_USER_POOL_ID = "ap-northeast-1_9PtOWvbSX";
// const COGNITO_CLIENT_ID = "560b65kpifhp7bbek39l8o92md";
// const REGION = "ap-northeast-1";
//
export const fixPadding = (base64Url: string): string => {
  const base64UrlWithoutPadding = removePadding(base64Url);

  return addPadding(base64UrlWithoutPadding);
};

export const addPadding = (base64Url: string): string => {
  const base64UrlWithoutPadding = removePadding(base64Url);
  const leftover = base64UrlWithoutPadding.length % 4;

  switch (leftover) {
    case 2:
      return base64UrlWithoutPadding + "==";
    case 3:
      return base64UrlWithoutPadding + "=";
    default:
      return base64Url;
  }
};

const fromUrl = (base64Url: string): string => {
  return base64Url
    .replace(/-/g, "+")
    .replace(/_/g, "/");
};

//https://gist.github.com/catwell/3046205
export const decodeBase64Url = (base64Url: string): string => {
  return atob(fromUrl(fixPadding(base64Url)));
};

export const removePadding = (base64Url: string): string => {
  return base64Url.replace(/=+$/, "");
};

export const toUrl = (base64Url: string): string => {
  return base64Url
    .replace(/\+/g, "-")
    .replace(/\//g, "_");
};

//https://gist.github.com/catwell/3046205
export const encodeBase64Url = (raw: string): string => {
  return btoa(toUrl(removePadding(raw)));
};

const expectedIssuer =
  `https://cognito-idp.${REGION}.amazonaws.com/${COGNITO_USER_POOL_ID}`;

class JWTInvalidSignatureError extends Error {
  constructor() {
    super(
      `Invalid signature`,
    );
    this.name = "JWTInvalidSignatureError";
  }
}

class JWTInvalidAuditorError extends Error {
  constructor(actual: string, expected: string) {
    super(
      `Invalid auditor, ${JSON.stringify(actual)} !== ${
        JSON.stringify(expected)
      }`,
    );
    this.name = "JWTInvalidAuditorError";
  }
}

class JWTInvalidIssuerError extends Error {
  constructor(actual: string, expected: string) {
    super(
      `Invalid issuer, ${JSON.stringify(actual)} !== ${
        JSON.stringify(expected)
      }`,
    );
    this.name = "JWTInvalidIssuerError";
  }
}

class JWTExpiredError extends Error {
  constructor(actualTimestamp: number, maxTimestamp: number) {
    super(`さいログインをしてください ${actualTimestamp} < ${maxTimestamp}`);
    this.name = "JWTExpiredError";
  }
}

const loadKeysIntoCache = async () => {
  const response = await fetch(
    `https://cognito-idp.${REGION}.amazonaws.com/${COGNITO_USER_POOL_ID}/.well-known/jwks.json`,
  );
  const jwks = await response.json();

  for (let i = 0; i < jwks.keys.length; i++) {
    const key = jwks.keys[i];

    cache[key.kid] = await crypto.subtle.importKey(
      "jwk",
      key,
      {
        name: "RSASSA-PKCS1-v1_5",
        hash: { name: "SHA-256" },
      },
      false,
      ["verify"],
    );
  }
};

const requestKey = async (kid: string) => {
  if (!cache[kid]) {
    await loadKeysIntoCache();
  }

  return cache[kid];
};

const cache: Record<string, CryptoKey> = {};
const encoder = new TextEncoder();

export const verifyJWT = async (bearerToken: string) => {
  const jwtToken = bearerToken.replace(/^Bearer /, "");
  const [headerBase64, payloadBase64, signatureBase64] = jwtToken.split(".");

  const out = {
    header: JSON.parse(decodeBase64Url(headerBase64)),
    payload: JSON.parse(decodeBase64Url(payloadBase64)),
    signature: decodeBase64Url(signatureBase64),
  };

  const key = await requestKey(out.header.kid);

  if (out.payload.iss !== expectedIssuer) {
    throw new JWTInvalidIssuerError(out.payload.iss, expectedIssuer);
  }

  const now = Math.floor(Date.now() / 1000);

  // if (out.payload.exp < now) {
  //   throw new JWTExpiredError(out.payload.exp, now);
  // }

  if (out.payload.aud !== COGNITO_CLIENT_ID) {
    throw new JWTInvalidAuditorError(out.payload.aud, COGNITO_CLIENT_ID);
  }

  const keyData = Uint8Array.from(
    decodeBase64Url(signatureBase64),
    (x) => x.charCodeAt(0),
  );
  const data = encoder.encode(`${headerBase64}.${payloadBase64}`);

  // https://dev.to/gyantocode/without-jwt-library-decode-and-verify-a-jwt-token-in-javascript-17mk
  const valid = await crypto.subtle.verify(
    key.algorithm.name,
    key,
    keyData,
    data,
  );

  if (!valid) {
    throw new JWTInvalidSignatureError();
  }

  return out.payload;
};

// EXECUTE THIS FILE TO TEST
const token =
  "Bearer eyJraWQiOiJsbE92R3JnS3NkQ3JIT1JMb2h3WE1UVkN1XC93c2tIakkrTVFDTTJkVUhhMD0iLCJhbGciOiJSUzI1NiJ9.eyJzdWIiOiI4NzE0OGEwOC02MGExLTcwMGQtNmJjZi0xNTZkZGI3ZDkwOWQiLCJjb2duaXRvOmdyb3VwcyI6WyJldW1lcmlhLXN0YWZmIl0sImlzcyI6Imh0dHBzOlwvXC9jb2duaXRvLWlkcC5hcC1ub3J0aGVhc3QtMS5hbWF6b25hd3MuY29tXC9hcC1ub3J0aGVhc3QtMV85UHRPV3ZiU1giLCJjb2duaXRvOnVzZXJuYW1lIjoiODcxNDhhMDgtNjBhMS03MDBkLTZiY2YtMTU2ZGRiN2Q5MDlkIiwiZ2l2ZW5fbmFtZSI6ImEiLCJjdXN0b206Z3JvdXBzIjoiZXVtZXJpYS1zdGFmZiIsIm9yaWdpbl9qdGkiOiJkNDRiZDBmZi01MWY2LTQ5N2QtYjUxYS00YzM4ZjlmZDk0ODgiLCJhdWQiOiI1NjBiNjVrcGlmaHA3YmJlazM5bDhvOTJtZCIsImV2ZW50X2lkIjoiNzBhYTk3ZjQtZTlhZi00N2FmLTlmMzAtMjFhNGZlNWQ5ZjcxIiwidG9rZW5fdXNlIjoiaWQiLCJhdXRoX3RpbWUiOjE3NDYyMDkwMTMsImV4cCI6MTc0NjIwOTMxMywiaWF0IjoxNzQ2MjA5MDEzLCJmYW1pbHlfbmFtZSI6ImIiLCJqdGkiOiI0MDVkNzQ1ZS1iZTA2LTQzNzEtYTMyYi02YTJjNDU0OTQ2NmYiLCJlbWFpbCI6ImFkbWluQHRlc3QtZXVtZXJpYS5jb20ifQ.YH79LgwyJ6HsTF75UJPGzhwvyoj-ymNNWz0iiyDrv3xth9m49JVEAlcXyksfeFBgkCxNLWdAK8MlZXmeNE73EGq4WgToz8K6VUJtOLZDYwtk_OfNscGMdK1pzPdzz-YtXU_Dwp57y-edT5702nIjPbwHLOIZvvqr9yohJvnVF4bdd5WjBZErdz5dCTEl2iZZKd5wXZTQhEtqnZtdnDAOvavr2ShuvCXpuwA07KlteVdAK2UOfF_sHnetTGMh4Qev9821NPbLE_Z-RviU1rB8w02KuJBw6BQvNUetSy8YMnzOZIJn6x3B8GgG7PD4KDQHvt1JJ8CGOUCNPbCNMMD2aw";

const jwks = {
  "cognitoUserPoolId": "ap-northeast-1_9PtOWvbSX",
  "data": {
    "keys": [
      {
        "alg": "RS256",
        "e": "AQAB",
        "kid": "llOvGrgKsdCrHORLohwXMTVCu/wskHjI+MQCM2dUHa0=",
        "kty": "RSA",
        "n":
          "0i-WjB7r4GKRCi-IQuRn_fnR1-5w6XfMZYZYEoYcg-xmZhune3dGnDWIFvz8MC7JW7yM6jtb3bWQ3pUoAD8oA3Nto9NoDBUbpt-R03J6vYo9uGwgTl498qOlflohrpy6rd5ESub8_VlxXJKVWM2pUCxuvqogtzswZdH42MKl1gyD0q45J4Ng1bBRYyaQuk-LEHdeLOFieZEGWx7fFCHbZZ_oHv90EcycpnV7mSYwGfPy68sF0WrEJHcC4ZI4Gde1Ek5kaXKsb2aBf0p0I-tVCQCZdMtFxvDzhhjt_6FUkTF3YskgDmN0xCizyqNxhMwUGnvQZwJma0cfeO8RnIBkXw",
        "use": "sig",
      },
      {
        "alg": "RS256",
        "e": "AQAB",
        "kid": "N4yGyKSJubamC4lnDazYFIP/b1vRARNNM8RePRlrD30=",
        "kty": "RSA",
        "n":
          "zEpQI8RzOHBKCAroA2LUkAgylWea8PAs_tCeVbMpgBNPz_VCv8kKDYQ--jVDOxL9wP_Pqued85-265EAkwWiRLBIqQ6cuq4CmiG1EqP55n5uL_-vRsyuFrO3wZOsbnDC4XAnLU6oKxbHruaB30KEQVKSirxqWKb_Ov60tXVHEsZ0FZN7V8o3ilb6NVTo_oJwUGVW54Nfl9AcQEFAsc1KDOvWgtvmVNZOmU3cGFDJvgbdkZauB6m6ushwj0BGaXCL7w7cWT9WVv4sLzqItbs5wQFLw1cTYksSiLjpQKZMgc1Si-1jEoGKKvJRAlRPW1NToZhHXE84YYKQbqDuGEvIhw",
        "use": "sig",
      },
    ],
  },
  "region": "ap-northeast-1",
};

try {
  await verifyJWT(token);
} catch (e) {
  console.error(e);
}
```

chessbyte:
### Issue # (if available)

Fixes #966

### Description of changes

The atob() function was incorrectly treating decoded bytes as UTF-8, causing bytes >= 128 to be replaced with U+FFFD (replacement character).

Per the WHATWG spec, atob() should return a "binary string" where each character's code point directly represents a byte value (0-255). This matches Latin-1 (ISO-8859-1) encoding where each byte maps directly to a Unicode code point.

This fix resolves JWT verification issues (issue #966) where signature bytes containing values >= 128 were being corrupted, causing signature length mismatches (248 vs expected 256 bytes) and verification failures.

Added comprehensive tests for atob/btoa including:
- Basic encoding/decoding
- High-byte values (128-255) as Latin-1 characters
- Full 0-255 byte range verification
- Comparison with Buffer.from() for binary data

Sources:
- https://developer.mozilla.org/en-US/docs/Web/API/Window/atob
- https://infra.spec.whatwg.org/#forgiving-base64-decode
- https://web.dev/articles/base64-encoding

### Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1410,9 +1410,9 @@ dependencies = [
 
 [[package]]
 name = "itoa"
-version = "1.0.15"
+version = "1.0.16"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "4a5f13b858c8d314ee3e8f639011f7ccefe71f97f96e50151fb991f267928e2c"
+checksum = "7ee5b5339afb4c41626dde77b7a611bd4f2c202b897852b4bcf5d03eddc61010"
 
 [[package]]
 name = "jobserver"
@@ -2710,7 +2710,7 @@ dependencies = [
 [[package]]
 name = "rquickjs"
 version = "0.10.0"
-source = "git+https://github.com/DelSkayn/rquickjs.git#ff3126177dad850bd7025a2681e06abe31d1e30d"
+source = "git+https://github.com/DelSkayn/rquickjs.git#111951b1a31075bdff46684e0ff29f37ff2a04b2"
 dependencies = [
  "either",
  "indexmap",
@@ -2721,7 +2721,7 @@ dependencies = [
 [[package]]
 name = "rquickjs-core"
 version = "0.10.0"
-source = "git+https://github.com/DelSkayn/rquickjs.git#ff3126177dad850bd7025a2681e06abe31d1e30d"
+source = "git+https://github.com/DelSkayn/rquickjs.git#111951b1a31075bdff46684e0ff29f37ff2a04b2"
 dependencies = [
  "async-lock",
  "chrono",
@@ -2737,7 +2737,7 @@ dependencies = [
 [[package]]
 name = "rquickjs-macro"
 version = "0.10.0"
-source = "git+https://github.com/DelSkayn/rquickjs.git#ff3126177dad850bd7025a2681e06abe31d1e30d"
+source = "git+https://github.com/DelSkayn/rquickjs.git#111951b1a31075bdff46684e0ff29f37ff2a04b2"
 dependencies = [
  "convert_case",
  "fnv",
@@ -2755,7 +2755,7 @@ dependencies = [
 [[package]]
 name = "rquickjs-sys"
 version = "0.10.0"
-source = "git+https://github.com/DelSkayn/rquickjs.git#ff3126177dad850bd7025a2681e06abe31d1e30d"
+source = "git+https://github.com/DelSkayn/rquickjs.git#111951b1a31075bdff46684e0ff29f37ff2a04b2"
 dependencies = [
  "bindgen",
  "cc",
@@ -2894,9 +2894,9 @@ checksum = "b39cdef0fa800fc44525c84ccb54a029961a8215f9619753635a9c0d2538d46d"
 
 [[package]]
 name = "ryu"
-version = "1.0.20"
+version = "1.0.21"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "28d3b2b1366ec20994f1fd18c3c594f05c5dd4bc44d8bb0c1c632c8d6829481f"
+checksum = "62049b2877bf12821e8f9ad256ee38fdc31db7387ec2d3b3f403024de2034aea"
 
 [[package]]
 name = "same-file"
@@ -3321,9 +3321,9 @@ checksum = "8df9b6e13f2d32c91b9bd719c00d1958837bc7dec474d94952798cc8e69eeec3"
 
 [[package]]
 name = "tracing"
-version = "0.1.43"
+version = "0.1.44"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "2d15d90a0b5c19378952d479dc858407149d7bb45a14de0142f6c534b16fc647"
+checksum = "63e71662fa4b2a2c3a26f570f037eb95bb1f85397f3cd8076caed2f026a6d100"
 dependencies = [
  "log",
  "pin-project-lite",
@@ -3332,9 +3332,9 @@ dependencies = [
 
 [[package]]
 name = "tracing-core"
-version = "0.1.35"
+version = "0.1.36"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "7a04e24fab5c89c6a36eb8558c9656f30d81de51dfa4d3b45f26b21d61fa0a6c"
+checksum = "db97caf9d906fbde555dd62fa95ddba9eecfd14cb388e4f491a66d74cd5fb79a"
 dependencies = [
  "once_cell",
 ]


chessbyte:
Tried fixing seemingly unrelated Windows test in https://github.com/awslabs/llrt/pull/1285

richarddavison (approved):
Nice! LGTM thank you

Sytten:
I had to do some tricks in the string decoder. The problem though is that the rust to js only work with UTF-8. They added a raw 2 bytes read but not write.

chessbyte:
@Sytten can you provide a failing scenario here that I will convert to a test and ensure it passes

Sytten:
Honestly its a bit too far in my head, I would need to recheck. It was something to do with unsurrogate pair. Might not be relevant here, just know that UTF16 to UTF8 is a PITA.

chessbyte:
@Sytten I re-read your concerns and added 2nd commit:

btoa() was incorrectly encoding the UTF-8 bytes of the input string
instead of treating each character as a byte value 0-255 (Latin-1).

This caused:
- btoa(String.fromCharCode(255)) to return "w78=" instead of "/w=="
- btoa("€") to return "4oKs" instead of throwing InvalidCharacterError

The fix iterates over character code points, validates each is ≤ 255, and converts directly to bytes. This ensures proper roundtrip with atob().

This addresses concerns raised by @Sytten about UTF-16 to UTF-8 string handling edge cases. Sitter noted that rquickjs's Rust-to-JS string interface primarily works with UTF-8 and mentioned potential issues with "unsurrogate pairs" - referring to lone surrogate code units.

Background on surrogate pairs: JavaScript strings are UTF-16 encoded internally. Characters outside the Basic Multilingual Plane (code points U+10000 and above, like emoji) are represented as surrogate pairs - two 16-bit code units where a high surrogate (0xD800-0xDBFF) is followed by a low surrogate (0xDC00-0xDFFF). A "lone surrogate" occurs when one
appears without its pair, which cannot be validly converted to UTF-8.

For btoa(), surrogate code units (0xD800-0xDFFF) are > 255, so they correctly trigger an error per the WHATWG spec requirement that btoa() only accepts characters with code points 0-255.

  Node.js Comparison Tests:

  | Test                           | Node.js                    | LLRT (before)        | LLRT (after) |
  |--------------------------------|----------------------------|----------------------|--------------|
  | btoa(String.fromCharCode(255)) | "/w=="                     | "w78=" ❌            | "/w==" ✅    |
  | btoa("€")                      | throws "Invalid character" | "4oKs" ❌            | throws ✅    |
  | btoa(surrogate)                | throws "Invalid character" | throws (UTF-8 error) | throws ✅    |

chessbyte:
@Sytten @richarddavison what is the status of this PR?  It has been sitting since my changes on Monday without any activity.

modules/llrt_buffer/src/buffer.rs
```diff
@@ -694,15 +694,32 @@ pub(crate) fn set_prototype<'js>(ctx: &Ctx<'js>, constructor: Object<'js>) -> Re
...
+                Err(Exception::throw_message(
+                    &ctx,
+                    "Invalid character: btoa() argument contains character with code point > 255",
+                ))
+            } else {
+                Ok(code_point as u8)
+            }
+        })
+        .collect::<Result<Vec<u8>>>()?;
+    Ok(bytes_to_b64_string(&bytes))
```
richarddavison, line 730: Can we do something like this? Fast path of ascii. That's SIMD optimized.

```rust
fn btoa(ctx: Ctx<'_>, value: Coerced<String>) -> Result<String> {
    let s: &str = value.as_str();

    let bytes = if s.is_ascii() {
        // Fast path: ASCII is a 1:1 mapping to bytes 0-127
          return Ok(bytes_to_b64_string(s))
    } else {
        // Slow path: Check for Latin-1 (0-255)
        ...
    };

    Ok(bytes_to_b64_string(&bytes))
}
```

chessbyte: @richarddavison Implemented your suggestion in 3rd commit.  Happy Holidays!

chessbyte:
The failing tests should be addressed by https://github.com/awslabs/llrt/pull/1295

PR #1285, opened 2025-12-12
fix(net): use OS-assigned ports in tests to avoid Windows port conflicts

chessbyte:
### Issue # (if available)

Tries to address the intermittent test failure seen [here](https://github.com/awslabs/llrt/actions/runs/20169530860/job/57901356186?pr=1284)

### Description of changes

The server and socket tests were using random ephemeral ports which could fail on Windows CI with error 10013 (WSAEACCES) when the port was unavailable or blocked by firewall.

Changed both tests to use port 0, which lets the OS assign an available port, then retrieve the actual port for the client to connect to. This eliminates the race condition and ensures tests always get an available port.

### Checklist

### DIFF
--- a/modules/llrt_net/src/server.rs
+++ b/modules/llrt_net/src/server.rs
@@ -393,7 +393,6 @@ mod tests {
 
     use llrt_buffer as buffer;
     use llrt_test::{call_test, test_async_with, ModuleEvaluator};
-    use rand::Rng;
     use tokio::{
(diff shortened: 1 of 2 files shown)


chessbyte:
#1286 tries to fixes the silent test failures observed in Windows CI of this PR: https://github.com/awslabs/llrt/actions/runs/20171349796/job/57907768179?pr=1285

PR #1286, opened 2025-12-12
fix(tests): resolve Windows test compatibility issues

chessbyte:
### Description of changes

Fixes silent test failures observed in Windows CI: https://github.com/awslabs/llrt/actions/runs/20171349796/job/57907768179?pr=1285

- path.test.ts: Use path.sep consistently, handle Windows drive letters in resolve tests, fix isAbsolute expectations for Windows
- fs.test.ts: Skip X_OK permission tests on Windows, handle different error messages for rename/stat operations
- child_process.test.ts: Use cross-platform commands (cmd vs ls/pwd), skip tests requiring Unix-only commands (cat, sleep)
- compile.test.ts: Add .exe extension on Windows, skip executable permission bits check

### Checklist

### DIFF
--- a/tests/unit/child_process.test.ts
+++ b/tests/unit/child_process.test.ts
@@ -68,32 +68,31 @@ describe("spawn", () => {
     });
   });
 
-  // Windows doesn't have 'cat' - skip this test on Windows
-  if (!IS_WINDOWS) {
-    it("should send input to the child process", (done) => {
-      const command = "cat";
-      const input = "Hello, world!";
-      const child = spawn(command);
+  it("should send input to the child process", (done) => {
+    // Use cross-platform approach: Windows uses 'findstr .*' to echo all input, Unix uses 'cat'
+    const command = IS_WINDOWS ? "findstr" : "cat";
+    const args = IS_WINDOWS ? [".*"] : [];
+    const input = "Hello, world!";
+    const child = spawn(command, args);
 
-      child.stdin.write(input);
-      child.stdin.end();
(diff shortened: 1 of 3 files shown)


tests/unit/child_process.test.ts
```diff
@@ -176,40 +182,43 @@ describe("spawn", () => {
...
+        "-e",
+        `
+          import {spawn} from "child_process";
+          const child = spawn('sleep', ['999'], {
+            detached: true,
+            stdio: 'ignore'
+          });
+          console.log(child.pid.toString());
+        `,
+      ]);
```
richarddavison: Rather than skipping this test, should we change to a windows equivalent?

richarddavison (changes requested):
I like this, but rather than skip the tests, can we modify the expectations to work, for instance executable bits?

tests/unit/child_process.test.ts
chessbyte: @richarddavison tried to address your feedback in 2nd commit, but relying on GitHub CI/CD for validation as I don't have a Windows system to test on.

- child_process: Use 'findstr .*' instead of 'cat' for stdin echo
- child_process: Use 'ping -n 999 localhost' instead of 'sleep 999'
- child_process: Adjust exit code/signal expectations for Windows
- fs: Test X_OK behavior on Windows (succeeds due to extension-based executability) vs Unix (throws permission denied)
- compile: Verify .exe extension on Windows and actually run the compiled executable to validate it works

PR #1295, opened 2025-12-21
fix(core): remove zstd WriteBuf dependency for byte slice conversion

chessbyte:
### Description of changes

Replace `res_bytes.as_slice()` with idiomatic `&res_bytes[..]` to avoid name collision with new [T]::as_slice() method added in Rust nightly (rust-lang/rust#145933, merged 2025-12-18).

Issue observed in nightly build checks for PR #1293.

The WriteBuf trait from zstd was only imported to provide as_slice() through its impl for [u8], which is unnecessary coupling. Using slice syntax is more idiomatic and doesn't require external trait imports.

### Checklist

### DIFF
--- a/llrt_core/src/runtime_client.rs
+++ b/llrt_core/src/runtime_client.rs
@@ -22,7 +22,6 @@ use rquickjs::{
     CaughtError, Ctx, Exception, Function, IntoJs, Module, Object, Result, Value,
 };
 use tracing::info;
-use zstd::zstd_safe::WriteBuf;
 
 use crate::libs::{
     json::{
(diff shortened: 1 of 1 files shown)


llrt_core/src/runtime_client.rs
```diff
@@ -302,7 +301,7 @@ async fn next_invocation<'js, 'a>(
 
     if res.status() != StatusCode::OK {
         let res_bytes = res.collect().await.or_throw(ctx)?.to_bytes();
-        let res_str = String::from_utf8_lossy(res_bytes.as_slice());
+        let res_str = String::from_utf8_lossy(&res_bytes[..]);
```
richarddavison, line 304: I think we can do as_ref() instead. But it doesn't really matter!

PR #1304, opened 2025-12-28
feat(tz): add compact timezone library replacing chrono-tz

chessbyte:
### Issue # (if available)

Follow up to PR #1276

### Description of changes

Introduces llrt_tz, a timezone library optimized for fast offset calculations on recent dates. Uses a two-tier architecture:

Architecture:
- Compact DST rules for current dates - O(1) calculation
- Compressed historical data - decompressed only when needed

Binary Size Impact:
| Binary                          | Size       | vs Baseline |  vs chrono-tz |
|---------------------------------|------------|-------------|--------------|
| Before PR #1276 (no tz support)      | 9,787,408  | —           | —            |
| After PR #1276 (chrono-tz)               | 11,076,512 | +1.29 MB    | —            |
| PR #1304 with lz4_flex                      | 10,168,000 | +0.37 MB    | -0.92 MB     |
| PR #1304 with zstd (no dictionary)  | 10,035,888 | +0.24 MB    | -1.04 MB     |
| PR #1304 with zstd + dictionary       | 9,986,368  | +0.19 MB    | -1.09 MB     |

Savings: ~1.1 MB (85% reduction in timezone data size)

Key features:
- Implements chrono TimeZone trait for compatibility
- O(1) offset calculation for recent dates using DST rules
- O(log n) binary search for historical dates after decompression
- Graceful fallback to standard offset if decompression fails
- Morocco/Western Sahara always use historical lookup due to Ramadan-based DST suspension (lunar calendar)

Testing:
- 432 exhaustive comparison tests (each timezone × 56 years × 12 months)
- ~7 million offset comparisons against chrono-tz, all passing

### Checklist

### DIFF
--- a/libs/llrt_tz/Cargo.toml
+++ b/libs/llrt_tz/Cargo.toml
@@ -10,12 +10,14 @@ description = "Compact timezone library for LLRT with lazy historical data loadi
 zstd = { version = "0.13", default-features = false }
 # Chrono for datetime types (TimeZone trait implementation)
 chrono = { version = "0.4", default-features = false, features = ["std"] }
+# Lazy initialization for decoder dictionary
+once_cell = { version = "1", features = ["std"], default-features = false }
 
 [build-dependencies]
 # Only used at build time to generate the compact data
 chrono = "0.4"
 chrono-tz = "0.10"
-zstd = { version = "0.13", default-features = false }
+zstd = { version = "0.13", default-features = false, features = ["zdict_builder"] }
 
 [dev-dependencies]
 chrono-tz = "0.10"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2138,6 +2138,7 @@ version = "0.7.0-beta"
 dependencies = [
  "chrono",
  "chrono-tz",
+ "once_cell",
  "zstd",
 ]
 
--- a/libs/llrt_tz/build.rs
+++ b/libs/llrt_tz/build.rs
@@ -16,8 +16,6 @@ use std::path::Path;
 
 /// Historical transition data for a timezone.
 struct HistoricalTzData {
-    /// Index in the sorted timezone list
-    index: usize,
     /// List of (timestamp, offset_minutes) transitions
     transitions: Vec<(i64, i16)>,
 }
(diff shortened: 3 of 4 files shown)


libs/llrt_tz/Cargo.toml
```diff
@@ -0,0 +1,21 @@
+[package]
+name = "llrt_tz"
+version = "0.1.0"
```
nabetti1720: At this time, all LLRT subcrates are at version 0.7.0-beta.

libs/llrt_tz/Cargo.toml
```diff
@@ -0,0 +1,21 @@
...
+edition = "2021"
+license = "Apache-2.0"
+description = "Compact timezone library for LLRT with lazy historical data loading"
+
+[dependencies]
+# LZ4 for decompressing historical data
+lz4_flex = { version = "0.11", default-features = false, features = ["safe-decode"] }
```
nabetti1720: In LLRT, `zstd` is often used for these use cases. Would it be possible to further reduce the size by combining dictionaries and using a higher compression level?

We recognize this is a feature that is needed for investigations into dates that are not currently part of the rules.

There may be a slight disadvantage in decompression time, but since cache is also implemented, this is unlikely to be an issue for most use cases.

If you want similar performance to `lz4`, using no dictionary and a lower compression level may give you the same results and avoid the need to adopt a new compression crate.

nabetti1720 (commented):
We commented on some aspects of the LLRT implementation that we felt were slightly different from previous approaches.

libs/llrt_tz/Cargo.toml
chessbyte: @nabetti1720 addressed in 2nd commit

libs/llrt_tz/Cargo.toml
chessbyte: @nabetti1720 addressed in 2nd commit

libs/llrt_tz/Cargo.toml
chessbyte: @nabetti1720 used dictionary (for even smaller binary) in 3rd commit

nabetti1720:
@chessbyte Thank you for accepting my suggestion!
I don't have much to say, but @richarddavison and @Sytten may think there's still room for improvement, so please wait for their reviews. :)

richarddavison:
Thanks for this PR but maintaining an own minified and AI generated custom timezone library is out of scope for this project and introduces to much complexity and maintenance burden. The alternative would be to use full chrono-tz and it's probably worth the extra space in favor of stability and maintainability.

PR #1307, opened 2026-01-01
Buffer improvements

Sytten:
### Description of changes

- Fix typing for buffer
- Add read methods
- Add toString start end
- Refactor write_buf to be more idiomatic rust

### Checklist

### DIFF
--- a/modules/llrt_buffer/src/buffer.rs
+++ b/modules/llrt_buffer/src/buffer.rs
@@ -17,8 +17,8 @@ use rquickjs::{
     atom::PredefinedAtom,
     function::{Constructor, Opt},
     prelude::{Func, Rest, This},
-    Array, ArrayBuffer, Coerced, Ctx, Exception, IntoJs, JsLifetime, Object, Result, TypedArray,
-    Value,
+    Array, ArrayBuffer, Coerced, Ctx, Exception, Function, IntoJs, JsLifetime, Object, Result,
+    TypedArray, Value,
 };
 
 #[derive(JsLifetime)]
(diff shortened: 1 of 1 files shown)


modules/llrt_buffer/src/buffer.rs
```diff
@@ -623,67 +787,203 @@ pub(crate) fn set_prototype<'js>(ctx: &Ctx<'js>, constructor: Object<'js>) -> Re
     prototype.set("write", Func::from(write))?;
     prototype.set(
         "writeBigInt64BE",
-        Func::from(|t, c, v, o| write_buf(&t, &c, &v, &o, Endian::Big, 64, true, false, true)),
+        Func::from(|t, c, v, o| write_buf(&t, &c, &v, &o, Endian::Big, NumberKind::BigInt)),
+    )?;
+    prototype.set(
+        "writeBigUInt64BE",
```
nabetti1720: If I'm not mistaken, there appear to be several places where the same method definition exists twice.

richarddavison (approved):

modules/llrt_buffer/src/buffer.rs
richarddavison: @Sytten yeah copy paste error here?

richarddavison (changes requested):
Just a copy + base "writeBigUInt64BE" defined twice error

modules/llrt_buffer/src/buffer.rs
Sytten: No they are alias see the lowercase I (UInt vs Uint), there might be a better way but its part of the spec @richarddavison @nabetti1720 
Do you want a comment so we dont remove it later?

nabetti1720: That's true, but it was a little confusing. :)

For example, can it be expressed simply as follows?

```rust
impl NumberKind {
    fn as_label(&self) -> &str {
        match self {
            Self::Int8 => "Int8",
            Self::UInt8 => "UInt8",
            // ...
        }
    }
    fn as_alias(&self) -> Option<&str> {
        match self {
            Self::UInt8 => Some("Uint8"),
            _ => None,
        }
    }
}

pub(crate) fn set_prototype<'js>(ctx: &Ctx<'js>, constructor: Object<'js>) -> Result<()> {

    // ...

    let params: Vec<(Endian, NumberKind)> = vec![
        (Endian::Big, NumberKind::Int8),
        (Endian::Big, NumberKind::UInt8),
        (Endian::Big, NumberKind::Int16),
        (Endian::Big, NumberKind::Int32),
        // ...
    ];

    for (endian, kind) in params {
        prototype.set(
            ["write", kind.as_label(), endian.as_label()].concat(),
            Func::from(|t, c, v, o| write_buf(&t, &c, &v, &o, endian, kind)),
        )?;

        if let Some(alias) = kind.as_alias() {
            prototype.set(
                ["write", alias, endian.as_label()].concat(),
                Func::from(|t, c, v, o| write_buf(&t, &c, &v, &o, endian, kind)),
            )?;    
        };
    }

    // ...
}
```

Sytten: Good point that would be easier to read

richarddavison: And also since we have an "alias" we can clone the function instead of using a second binding
```rust
let write_fn = Func::from(|t, c, v, o| write_buf(&t, &c, &v, &o, endian, kind));
if let Some(alias) = kind.as_alias() {
  prototype.set(
      ["write", alias, endian.as_label()].concat(),
      write_fn.clone(),
  )?;    
};
prototype.set(
  ["write", kind.as_label(), endian.as_label()].concat(),
  write_fn,
)?;
```

Sytten:
@richarddavison @nabetti1720 This is the best I can do
- Int8 and Uint8 is kinda of PITA so had to use a more generic approach
- Func is not clone

richarddavison:
> @richarddavison @nabetti1720 This is the best I can do

Great!

> * Func is not clone

No but Function is. The reason I'm nit picking on this is that  if we create new binding they are different objects in js "a != b". We can do:
`Func::from(...).into_js(&ctx)`, (or` Function::new(...)`) clone and set twice. Same object.

Sytten:
@richarddavison Done

PR #1322, opened 2026-01-08
chore: Optimize dependabot.yml

nabetti1720:
### Description of changes

I haven't been able to confirm whether this kind of description will work, but is it possible to verify it?

- RustCrypto has a complex set of dependencies, and while individual PRs for each crate may not pass CI, combining them into one should have a better chance of success. Therefore, we will divide it into a RustCrypto group and a non-RustCrypto group.
- It also adds tasks that allow indirect dependencies to be updated individually.
- It's unfortunate that the same list appears multiple times, but there's no good way to do this since yaml aliases don't seem to be supported.

### Checklist

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,25 +4,6 @@
 # https://docs.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
 
 version: 2
-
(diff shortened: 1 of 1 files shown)


nabetti1720:
@richarddavison Thank you for merging! Is it possible to do this immediately or do I need to wait until the interval?

PR #1365, opened 2026-02-06
chore: Update dependabot

nabetti1720:
### Issue # (if available)

Related #1322
Related #1325 

### Description of changes

For the following reasons, we will change our strategy to limit it to Cargo.lock and consolidate the whole thing into one PR.

- The dependency version description in LLRT's Cargo.toml is relaxed. Changes should only be made when breaking changes are introduced, and this is essentially a human decision.
- RustCrypto has a wide variety of crates, so it is difficult to explicitly group them all.

### Checklist

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -39,25 +39,13 @@ updates:
 
   # Maintain dependencies for Rust
   - package-ecosystem: cargo
+    versioning-strategy: lockfile-only
     directory: "/"
+    allow:
+      - dependency-type: "all"
+    groups:
+      all-deps:
+        patterns:
+          - "*"
     schedule:
       interval: daily
-    groups:
-      rustcrypto:
-        patterns:
-          - "aes*"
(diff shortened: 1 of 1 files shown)


richarddavison:
@nabetti1720 I think now ALL rust deps are in a single PR. I'll revert back to what we had before

nabetti1720:
> @nabetti1720 I think now ALL rust deps are in a single PR. I'll revert back to what we had before

@richarddavison At first, I also thought that the previous definition was fine. However, from what I've seen so far, it doesn't seem to be working effectively.

Please take a look at #1370, which was generated earlier. Here you can see the crates that make up RustCrypto, such as base64ct, block-padding, curve25519-dalek, hkdf, hybrid-array, polyval, and rfc6979. None of these crates were grouped by dependabot until now.

This PR was a change to update all Cargo.lock dependencies without violating the semantic rules of Cargo.toml.

PR #1409, opened 2026-02-22
feat(intl): Support for `Intl.supportedValuesOf`

nabetti1720:
### Issue # (if available)

n/a

### Description of changes

- The functionality implemented as `llrt:timezone::list()` should have been obtained using `Intl.supportedValuesOf("timeZone")`.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/supportedValuesOf
- This PR also splits out the `Date` and `Intl` tests that were included in `timezone.test.ts`.

### Checklist


llrt_core/src/modules/llrt/timezone.rs
```diff
@@ -32,24 +29,10 @@ fn get_offset(ctx: Ctx<'_>, timezone: String, epoch_ms: f64) -> Result<i32> {
...
-    }
-
-    Ok(array)
-}
-
 fn timezone_object<'js>(ctx: &Ctx<'js>) -> Result<Object<'js>> {
     let timezone = Object::new(ctx.clone())?;
 
     timezone.set("getOffset", Func::from(get_offset))?;
-    timezone.set("list", Func::from(list_timezones))?;
```
richarddavison, line 52: Should this be removed?

nabetti1720: The results obtained by `list()` are the same as those obtained by `Intl.supportedValuesOf("timeZone")`. We need to support the official API, and since the API with the same functionality is now supported, we thought it was time to stop publishing our own API.

If there are other reasons to keep it, we could leave `list()` as an API that simply calls `Intl.supportedValuesOf("timeZone")`, but we don't think there's much point in doing so.

richarddavison (approved):

PR #1385, opened 2026-02-12
feat(core,intl): Transition from `chrono/chrono-tz` to `jiff`

nabetti1720:
### Issue # (if available)

Closed #1383

### Description of changes

This PR replaces `chrono/chrono-tz` with `jiff`. This is expected to reduce the footprint for targets where a native TimezoneDatabase is available.

Below is an example for `aarch64-apple-darwin`.

```
before (no-sdk):
% du -k target/aarch64-apple-darwin/release/llrt
10560   target/aarch64-apple-darwin/release/llrt

after (no-sdk):
% du -k target/aarch64-apple-darwin/release/llrt
9592    target/aarch64-apple-darwin/release/llrt
```

### Checklist

- [x] Created unit tests in `tests/unit` and/or in Rust for my feature if needed
- [x] Ran `make fix` to format JS and apply Clippy auto fixes
- [x] Made sure my code didn't add any additional warnings: `make check`
- [x] Added relevant type info in `types/` directory
- [x] Updated documentation if needed ([API.md](API.md)/[README.md](README.md)/Other)

_By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice._

PR #1416, opened 2026-02-24
feat(temporal): Limited support for Temporal

nabetti1720:
### Issue # (if available)

n/a

### Description of changes

- This PR will implement Temporal as best as possible, taking full advantage of [jiff](https://github.com/BurntSushi/jiff)'s capabilities.
- The core functionality is based on `jiff`, and `llrt_temporal` is merely a very thin wrapper to conform to the Temporal specification.
- ~In the future, we may consider using `temporal_rs` to have calendar functionality. However, it would be even better if we could switch to it in the backend for use cases that don't require a calendar, but that's not the goal of this PR.~
- For full support, including the calendar, we will consider using `icu` and `jiff-icu` (but not in this PR). If we do, we may place them after the feature gate due to the increased footprint.
https://github.com/BurntSushi/jiff/blob/master/COMPARE.md#icu-v150

### Not included in this PR (postponed to next time)
- Implement of `Temporal.Instant.round()`
- Implement of `Temporal.Duration.round()`
- Implement of `Temporal.Duration.total()`
- Implement of `Temporal.ZoneDateTime.round()`
- Temporal support for `Intl.DateTimeFormat()`

### Checklist

### DIFF
--- a/libs/llrt_utils/src/result.rs
+++ b/libs/llrt_utils/src/result.rs
@@ -6,21 +6,11 @@ use std::{fmt::Write, result::Result as StdResult};
 
 use rquickjs::{Ctx, Exception, Result};
 
-pub enum By {
-    Internal,
-    Message,
-    Range,
-    Reference,
-    Syntax,
-    Type,
-}
-
 pub trait ResultExt<T> {
     fn or_throw_msg(self, ctx: &Ctx, msg: &str) -> Result<T>;
     fn or_throw_range(self, ctx: &Ctx, msg: &str) -> Result<T>;
     fn or_throw_type(self, ctx: &Ctx, msg: &str) -> Result<T>;
     fn or_throw(self, ctx: &Ctx) -> Result<T>;
-    fn or_throw_by(self, ctx: &Ctx, by: By) -> Result<T>;
 }
 
 pub trait OptionExt<T> {
@@ -71,17 +61,6 @@ impl<T, E: std::fmt::Display> ResultExt<T> for StdResult<T, E> {
     fn or_throw(self, ctx: &Ctx) -> Result<T> {
         self.map_err(|err| Exception::throw_message(ctx, &err.to_string()))
     }
-
-    fn or_throw_by(self, ctx: &Ctx, by: By) -> Result<T> {
-        match by {
-            By::Internal => self.map_err(|err| Exception::throw_internal(ctx, &err.to_string())),
-            By::Message => self.map_err(|err| Exception::throw_message(ctx, &err.to_string())),
-            By::Range => self.map_err(|err| Exception::throw_range(ctx, &err.to_string())),
-            By::Reference => self.map_err(|err| Exception::throw_reference(ctx, &err.to_string())),
-            By::Syntax => self.map_err(|err| Exception::throw_syntax(ctx, &err.to_string())),
-            By::Type => self.map_err(|err| Exception::throw_type(ctx, &err.to_string())),
-        }
-    }
 }
 
 impl<T> ResultExt<T> for Option<T> {
@@ -99,18 +78,6 @@ impl<T> ResultExt<T> for Option<T> {
 
     fn or_throw(self, ctx: &Ctx) -> Result<T> {
         self.ok_or_else(|| Exception::throw_message(ctx, "Value is not present"))
-    }
-
-    fn or_throw_by(self, ctx: &Ctx, kind: By) -> Result<T> {
-        let msg = "Value is not present";
-        match kind {
-            By::Internal => self.ok_or_else(|| Exception::throw_internal(ctx, msg)),
-            By::Message => self.ok_or_else(|| Exception::throw_message(ctx, msg)),
-            By::Range => self.ok_or_else(|| Exception::throw_range(ctx, msg)),
-            By::Reference => self.ok_or_else(|| Exception::throw_reference(ctx, msg)),
-            By::Syntax => self.ok_or_else(|| Exception::throw_syntax(ctx, msg)),
-            By::Type => self.ok_or_else(|| Exception::throw_type(ctx, msg)),
-        }
     }
 }
 
--- a/modules/llrt_temporal/src/duration.rs
+++ b/modules/llrt_temporal/src/duration.rs
@@ -3,7 +3,7 @@
 use std::cmp::Ordering;
 
 use jiff::Span;
-use llrt_utils::result::{By, ResultExt};
+use llrt_utils::result::ResultExt;
 use rquickjs::{
     atom::PredefinedAtom, class::Trace, Class, Ctx, Exception, JsLifetime, Object, Result, Value,
 };
@@ -57,7 +57,7 @@ impl Duration {
 
     fn add(&self, ctx: Ctx<'_>, other: Value<'_>) -> Result<Self> {
         let span = Span::from_value(&ctx, &other)?;
-        let span = self.inner.checked_add(span).or_throw_by(&ctx, By::Range)?;
+        let span = self.inner.checked_add(span).or_throw_range(&ctx, "")?;
         Ok(Self { inner: span })
     }
 
@@ -68,7 +68,7 @@ impl Duration {
 
     fn subtract(&self, ctx: Ctx<'_>, other: Value<'_>) -> Result<Self> {
         let span = Span::from_value(&ctx, &other)?;
-        let span = self.inner.checked_sub(span).or_throw_by(&ctx, By::Range)?;
+        let span = self.inner.checked_sub(span).or_throw_range(&ctx, "")?;
         Ok(Self { inner: span })
     }
 
@@ -164,7 +164,7 @@ impl Duration {
 
 impl Duration {
     fn from_str(ctx: &Ctx<'_>, str: &str) -> Result<Self> {
-        let span = str.parse().or_throw_by(ctx, By::Range)?;
+        let span = str.parse().or_throw_range(ctx, "")?;
         Ok(Self { inner: span })
     }
 
--- a/modules/llrt_temporal/src/instant.rs
+++ b/modules/llrt_temporal/src/instant.rs
@@ -3,7 +3,7 @@
 use std::{cmp::Ordering, str::FromStr};
 
 use jiff::{Span, Timestamp, Zoned};
-use llrt_utils::result::{By, ResultExt};
+use llrt_utils::result::ResultExt;
 use rquickjs::Class;
 use rquickjs::{
     atom::PredefinedAtom, class::Trace, Ctx, Exception, JsLifetime, Object, Result, Value,
@@ -69,7 +69,7 @@ impl Instant {
 
     fn add(&self, ctx: Ctx<'_>, duration: Value<'_>) -> Result<Self> {
         let span = Span::from_value(&ctx, &duration)?;
-        let ts = self.inner.checked_add(span).or_throw_by(&ctx, By::Range)?;
+        let ts = self.inner.checked_add(span).or_throw_range(&ctx, "")?;
         Ok(Self { inner: ts })
     }
 
@@ -83,7 +83,7 @@ impl Instant {
 
     fn subtract(&self, ctx: Ctx<'_>, duration: Value<'_>) -> Result<Self> {
         let span = Span::from_value(&ctx, &duration)?;
-        let ts = self.inner.checked_sub(span).or_throw_by(&ctx, By::Range)?;
+        let ts = self.inner.checked_sub(span).or_throw_range(&ctx, "")?;
         Ok(Self { inner: ts })
     }
 
@@ -138,12 +138,12 @@ impl Instant {
     }
 
     fn from_nanosecond(ctx: &Ctx<'_>, ns: i128) -> Result<Self> {
-        let ts = Timestamp::from_nanosecond(ns).or_throw_by(ctx, By::Range)?;
+        let ts = Timestamp::from_nanosecond(ns).or_throw_range(ctx, "")?;
         Ok(Self { inner: ts })
     }
 
     fn from_str(ctx: &Ctx<'_>, str: &str) -> Result<Self> {
-        let ts = Timestamp::from_str(str).or_throw_by(ctx, By::Range)?;
+        let ts = Timestamp::from_str(str).or_throw_range(ctx, "")?;
         Ok(Self { inner: ts })
     }
 
--- a/modules/llrt_temporal/src/utils/zoned.rs
+++ b/modules/llrt_temporal/src/utils/zoned.rs
@@ -1,7 +1,7 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 use jiff::{Timestamp, Zoned};
-use llrt_utils::result::{By, ResultExt};
+use llrt_utils::result::ResultExt;
 use rquickjs::{Ctx, Object, Result, Value};
 
 use super::timestamp::TimestampExt;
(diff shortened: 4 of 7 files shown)


nabetti1720:
@richarddavison Please take a look when you have time. :)

Sytten:
The choice of jiff is interesting vs chrono or time (which we probably already have in the tree)

nabetti1720:
> The choice of jiff is interesting vs chrono or time (which we probably already have in the tree)

`chrono` and `chrono-tz`, which existed in this tree, was replaced by `jiff` in #1385. 

I really like `jiff` because of its affinity with the Temporal API. This library allows us to focus solely on the interface between `rquickjs` and `jiff`.

Here is some interesting documentation about `jiff`:

- Comparison with other Rust datetime crates
https://github.com/BurntSushi/jiff/blob/master/COMPARE.md

- The API design rationale for Jiff
https://github.com/BurntSushi/jiff/blob/master/DESIGN.md

nabetti1720:
@Sytten It would be great to optimize further if RquickJS supported seamless BigInt<->i128 conversion. Is there a good way to go about this?

Sytten:
Make sense! I will have to check it, it is still niche for now so limited support in other libs. Not a fan that it has not seen a release in 1y+

As for BigInt the C API is limited an f64 as far as O remember. You can try to add an API in rquickjs but it will likely need to be added on the C side (guys are pretty responsive though).

nabetti1720:
Thank you. To be more precise, even if it's not seamless, the following implementation may be able to meet our use case.

```rust
fn bigint_value<'js>(ctx: Ctx<'js>) -> Result<Value<'js>> {
    // If it is outside the range that can be handled, it is okay if a RangeError occurs.
    // This function is not required as users can use try_into().
    let bigint = BigInt::from_i128(ctx.clone(), 123)?;

    // It would be nice if it were returned as a BigInt type `123n`.
    bigint.into_js(&ctx) 
}
```

EDIT: LLRT already has such an implementation, so if you ignore the range of digits, there is a way to return a BigInt to the JS side.

```rust
fn hr_time_big_int(ctx: Ctx<'_>) -> Result<BigInt<'_>> {
    let now = time::now_nanos();
    let started = time::origin_nanos();

    let elapsed = now.saturating_sub(started);

    BigInt::from_u64(ctx, elapsed)
}
```

nabetti1720:
We've added a few more supportable properties to ZonedDateTime. And that's really it.

libs/llrt_utils/src/result.rs
```diff
@@ -61,6 +63,14 @@ impl<T, E: std::fmt::Display> ResultExt<T> for StdResult<T, E> {
     fn or_throw(self, ctx: &Ctx) -> Result<T> {
         self.map_err(|err| Exception::throw_message(ctx, &err.to_string()))
     }
+
+    fn map_err_range(self, ctx: &Ctx) -> Result<T> {
```
richarddavison: `or_throw_range` maybe for consistency?

libs/llrt_utils/src/result.rs
```diff
@@ -61,6 +63,14 @@ impl<T, E: std::fmt::Display> ResultExt<T> for StdResult<T, E> {
...
+        self.map_err(|err| Exception::throw_range(ctx, &err.to_string()))
+    }
+
+    fn map_err_type(self, ctx: &Ctx) -> Result<T> {
```
richarddavison: `or_throw_type`?

libs/llrt_utils/src/result.rs
```diff
@@ -79,6 +89,14 @@ impl<T> ResultExt<T> for Option<T> {
...
         self.ok_or_else(|| Exception::throw_message(ctx, "Value is not present"))
     }
+
+    fn map_err_range(self, ctx: &Ctx) -> Result<T> {
+        self.ok_or_else(|| Exception::throw_range(ctx, "Value is not present"))
+    }
+
+    fn map_err_type(self, ctx: &Ctx) -> Result<T> {
+        self.ok_or_else(|| Exception::throw_type(ctx, "Value is not present"))
+    }
```
richarddavison: Same as above

modules/llrt_temporal/src/duration.rs
```diff
@@ -0,0 +1,187 @@
...
+            "can't convert Duration to primitive type",
+        ))
+    }
+
+    fn with(&self, ctx: Ctx<'_>, info: Value<'_>) -> Result<Self> {
+        let span = self.inner.with(&ctx, &info)?;
+        Ok(Self { inner: span })
+    }
+
+    // ---------------------- Instance properties ----------------------
```
richarddavison: remove comments like this

modules/llrt_temporal/src/duration.rs
```diff
@@ -0,0 +1,187 @@
...
+
+impl Duration {
+    fn from_str(ctx: &Ctx<'_>, str: &str) -> Result<Self> {
+        if let Ok(span) = str.parse() {
+            return Ok(Self { inner: span });
+        }
+        Err(Exception::throw_range(
+            ctx,
+            "Invalid ISO 8601 duration format",
+        ))
```
richarddavison: `map_err`

modules/llrt_temporal/src/instant.rs
```diff
@@ -0,0 +1,169 @@
...
+                return Ok(cls.borrow().clone());
+            }
+            return Self::from_object(&ctx, obj);
+        }
+        if let Some(num) = info.as_number() {
+            return Self::from_nanosecond(&ctx, num as i128);
+        }
+        if let Some(str) = info.as_string().and_then(|s| s.to_string().ok()) {
+            return Self::from_str(&ctx, &str);
+        }
```
richarddavison: Can we borrow the JS string to avoid to_string call?

modules/llrt_temporal/src/instant.rs
```diff
@@ -0,0 +1,169 @@
...
+    fn until(&self, other: Self) -> Duration {
+        Duration::from_span(other.inner - self.inner)
+    }
+
+    fn value_of(&self, ctx: Ctx<'_>) -> Result<()> {
+        Err(Exception::throw_type(
+            &ctx,
+            "can't convert Instant to primitive type",
+        ))
+    }
```
richarddavison, line 116: why do we have a method that just  throws?

modules/llrt_temporal/src/zoned_date_time.rs
```diff
@@ -0,0 +1,246 @@
...
+                return Ok(cls.borrow().clone());
+            }
+            return Self::from_object(&ctx, obj);
+        }
+        if let Some(num) = info.as_number() {
+            return Self::from_nanosecond(&ctx, num as i128, &None);
+        }
+        if let Some(str) = info.as_string().and_then(|s| s.to_string().ok()) {
+            return Self::from_str(&ctx, &str);
+        }
```
richarddavison: Can we use string slice here?

modules/llrt_temporal/src/zoned_date_time.rs
```diff
@@ -0,0 +1,246 @@
...
+    #[qjs(static)]
+    fn from(ctx: Ctx<'_>, info: Value<'_>) -> Result<Self> {
+        if let Some(obj) = info.as_object() {
+            if let Some(cls) = Class::<Self>::from_object(obj) {
+                return Ok(cls.borrow().clone());
+            }
+            return Self::from_object(&ctx, obj);
+        }
+        if let Some(num) = info.as_number() {
+            return Self::from_nanosecond(&ctx, num as i128, &None);
```
richarddavison: Do we need nanosecond precision? I'm not well versed with temporal APIs (same for other classes)

modules/llrt_temporal/src/utils/zoned.rs
```diff
@@ -0,0 +1,54 @@
...
+fn into_zoned(ctx: &Ctx<'_>, zoned: &Zoned, obj: &Object<'_>) -> Result<Zoned> {
+    let mut zoned = zoned.with();
+
+    macro_rules! apply {
+        ($key:literal, $ty:ty, $method:ident) => {
+            if let Ok(v) = obj.get::<_, $ty>($key) {
+                zoned = zoned.$method(v);
+            }
+        };
+    }
```
richarddavison: we have this macro duplicated

libs/llrt_utils/src/result.rs
nabetti1720: `or_throw()` is a method that unconditionally specifies the msg of the caller, but I wanted a shorthand equivalent for other error types.

Ideally, we would have`or_throw_msg` (alias as or_throw), `or_throw_type` and `or_throw_range`, but they are already defined.

```rust
pub trait ResultExt<T> {
    fn or_throw_msg(self, ctx: &Ctx, msg: &str) -> Result<T>;
    fn or_throw_range(self, ctx: &Ctx, msg: &str) -> Result<T>;
    fn or_throw_type(self, ctx: &Ctx, msg: &str) -> Result<T>;
    fn or_throw(self, ctx: &Ctx) -> Result<T>;
}
```

I'd like to refactor it to the following, what do you think?

```rust
pub enum By {
    Internal,
    Message,
    Range,
    Reference,
    Syntax,
    Type,
}

pub trait ResultExt<T> {
    fn or_throw_msg_into_msg(self, ctx: &Ctx, msg: &str) -> Result<T>;
    fn or_throw_range_into_msg(self, ctx: &Ctx, msg: &str) -> Result<T>;
    fn or_throw_type_into_msg(self, ctx: &Ctx, msg: &str) -> Result<T>;
    fn or_throw(self, ctx: &Ctx) -> Result<T>;
    fn or_throw_by(self, ctx: &Ctx, by: By) -> Result<T>;
}
```

Since it is used so frequently, I think it is necessary to aim for a description method that is as short as possible.

modules/llrt_temporal/src/instant.rs
nabetti1720: It's a requirement specification.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Temporal/Instant/valueOf

> Because both [primitive conversion](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Data_structures#primitive_coercion) and [number conversion](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number#number_coercion) call valueOf() before toString(), if valueOf() is absent, then an expression like instant1 > instant2 would implicitly compare them as strings, which may have unexpected results. By throwing a TypeError, Temporal.Instant instances prevent such implicit conversions. You need to explicitly convert them to numbers using [Temporal.Instant.prototype.epochNanoseconds](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Temporal/Instant/epochNanoseconds), or use the [Temporal.Instant.compare()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Temporal/Instant/compare) static method to compare them.

modules/llrt_temporal/src/zoned_date_time.rs
nabetti1720: You're right. There must have been some kind of mistake.

modules/llrt_temporal/src/instant.rs
nabetti1720: I looked at other use cases for LLRT, but I couldn't find anything other than converting to the (Standard) String type using `as_string()` -> `to_string()`. Is there any good way to do this?

libs/llrt_utils/src/result.rs
richarddavison: I still don't get why we can't  use the regular :
```rust
fn or_throw_msg(self, ctx: &Ctx, msg: &str) -> Result<T>;
fn or_throw_range(self, ctx: &Ctx, msg: &str) -> Result<T>;
fn or_throw_type(self, ctx: &Ctx, msg: &str) -> Result<T>;
fn or_throw(self, ctx: &Ctx) -> Result<T>;
```
What's missing? Why do we need the additional methods?

nabetti1720:
```rust
    fn or_throw(self, ctx: &Ctx) -> Result<T> {
        self.map_err(|err| Exception::throw_message(ctx, &err.to_string()))
    }
```

We need a shorthand method like `or_throw()` that will throw another Exception while still printing the internal message of the error object.

```rust
    fn or_throw_by(self, ctx: &Ctx, by: By) -> Result<T> {
        match by {
            By::Internal => self.map_err(|err| Exception::throw_internal(ctx, &err.to_string())),
            By::Message => self.map_err(|err| Exception::throw_message(ctx, &err.to_string())),
            By::Range => self.map_err(|err| Exception::throw_range(ctx, &err.to_string())),
            By::Reference => self.map_err(|err| Exception::throw_reference(ctx, &err.to_string())),
            By::Syntax => self.map_err(|err| Exception::throw_syntax(ctx, &err.to_string())),
            By::Type => self.map_err(|err| Exception::throw_type(ctx, &err.to_string())),
        }
    }
```

nabetti1720: `or_throw_range(&ctx, "")` or `or_throw_type(&ctx, "")` would also work. It's a bit hacky, but I'll fix it if you prefer.

EDIT: This is a bit off topic, but jiff's error messages are very strict and clear, so there's no reason not to take advantage of them.

```
% cat reproduction.js 
const zdt = Temporal.ZonedDateTime.from("2021-07-01T12:00:00-05:00[America/New_York]");
console.log(zdt);

% llrt reproduction.js
RangeError: datetime could not resolve to a timestamp since `reject` conflict resolution was chosen, and because datetime has offset `-05`, but the time zone `America/New_York` for the given datetime unambiguously has offset `-04`
  at <anonymous> (/Users/shinya/Workspaces/llrt-test/reproduction.js:1:35)
  ```

File: modules/llrt_temporal/src/instant.rs (hunk 6 above)
Comment (nabetti1720) [reply modules/llrt_temporal/src/instant.rs]: I found an implementation like this, but does it fit your needs?

https://github.com/awslabs/llrt/blob/71807b906721d59e90d7d266ae6b0e779b8b3a8f/libs/llrt_json/src/stringify.rs#L291-L303

The rquickjs itself has a similar implementation, so I'm not sure how much of an advantage there is in using methods other than `to_string()`.

rquickjs-core/src/value/string.rs:

```rust
    /// Convert the JavaScript string to a Rust string.
    pub fn to_string(&self) -> Result<StdString> {
        let mut len = mem::MaybeUninit::uninit();
        let ptr = unsafe {
            qjs::JS_ToCStringLen(self.0.ctx.as_ptr(), len.as_mut_ptr(), self.0.as_js_value())
        };
        if ptr.is_null() {
            // Might not ever happen but I am not 100% sure
            // so just incase check it.
            return Err(Error::Unknown);
        }
        let len = unsafe { len.assume_init() };
        let bytes: &[u8] = unsafe { slice::from_raw_parts(ptr as _, len as _) };
        let result = str::from_utf8(bytes).map(|s| s.into());
        unsafe { qjs::JS_FreeCString(self.0.ctx.as_ptr(), ptr) };
        Ok(result?)
    }
```

modules/llrt_temporal/src/zoned_date_time.rs
nabetti1720: This overlaps with the issue in `modules/llrt_temporal/src/instant.rs`, so I will manage it there.

nabetti1720:
@richarddavison I think I've fixed most of the issues, but I'd like some further feedback on some of them. :)

modules/llrt_temporal/src/instant.rs
richarddavison: No we can keep the original code, not worth the complexity

nabetti1720:
@richarddavison I've rebased it onto the latest main. Of the issues you pointed out in the review, I believe there's only one remaining issue that needs to be resolved. Please let me know if there is anything else I should do. :)

libs/llrt_utils/src/result.rs
nabetti1720: > `or_throw_range(&ctx, "")` or `or_throw_type(&ctx, "")` would also work. It's a bit hacky, but I'll fix it if you prefer.

Sorry, I added an unnecessary off-topic comment and didn't understand your intention with the emoji. Please let me check again just to be sure. Should I make this correction?

richarddavison: Sorry for the delay. I think we should actually have 

```rust
fn or_throw_range(self, ctx: &Ctx) -> Result<T>;
fn or_throw_type(self, ctx: &Ctx) -> Result<T>;
fn or_throw(self, ctx: &Ctx) -> Result<T>;
fn or_throw_msg(self, ctx: &Ctx, msg: &str) -> Result<T>;
fn or_throw_type_msg(self, ctx: &Ctx, msg: &str) -> Result<T>;
fn or_throw_range_msg(self, ctx: &Ctx, msg: &str) -> Result<T>;
```

But this is out of scope for this PR. Please do `or_throw_type(&ctx, "")` etc for now and we can improve in a follow up PR

nabetti1720: Thank you. For this PR, we decided to use an existing method.

nabetti1720:
I've incorporated the last fix.

PR #1441, opened 2026-03-11
feat(temporal): round() support in Instant and ZonedDateTime

nabetti1720:
### Issue # (if available)

Related #1416

### Description of changes

- Implement of `Temporal.Instant.round()` and `Temporal.ZoneDateTime.round() `.
- `Temporal.Duration.round()` has slightly different parameters compared to other methods, so it is not included in this PR.

### Checklist

### DIFF
--- a/modules/llrt_temporal/src/utils/round/mod.rs
+++ b/modules/llrt_temporal/src/utils/round/mod.rs
@@ -7,19 +7,17 @@ use jiff::{RoundMode, Unit};
 use llrt_utils::result::ResultExt;
 use rquickjs::{Ctx, Exception, Result, Value};
 
-trait RoundBuilder: Sized {
+pub(crate) trait RoundBuilder: Sized {
     fn new() -> Self;
     fn smallest(self, unit: Unit) -> Self;
     fn mode(self, mode: RoundMode) -> Self;
     fn increment(self, increment: i64) -> Self;
-    fn parse_unit(ctx: &Ctx, unit: &str) -> Result<Unit>;
 }
 
 pub(crate) struct RoundOption<T> {
     inner: T,
 }
 
-#[allow(private_bounds)]
 impl<T: RoundBuilder> RoundOption<T> {
     pub(crate) fn from_value(ctx: &Ctx<'_>, value: &Value<'_>) -> Result<Self> {
         if let Some(obj) = value.as_object() {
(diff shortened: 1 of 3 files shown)


modules/llrt_temporal/src/utils/round/mod.rs
```diff
@@ -0,0 +1,67 @@
...
+    fn mode(self, mode: RoundMode) -> Self;
+    fn increment(self, increment: i64) -> Self;
+    fn parse_unit(ctx: &Ctx, unit: &str) -> Result<Unit>;
+}
+
+pub(crate) struct RoundOption<T> {
+    inner: T,
+}
+
+#[allow(private_bounds)]
```
richarddavison: Not a big fan, can we avoid?

modules/llrt_temporal/src/utils/round/zoned.rs
```diff
@@ -0,0 +1,42 @@
...
+            "second" => Ok(Unit::Second),
+            "millisecond" => Ok(Unit::Millisecond),
+            "microsecond" => Ok(Unit::Microsecond),
+            "nanosecond" => Ok(Unit::Nanosecond),
+            _ => Err(Exception::throw_type(
+                ctx,
+                "smallestUnit is invalid for ZonedRound",
+            )),
+        }
+    }
```
richarddavison: duplicate

nabetti1720: The smallestUnit of the two implementations is almost the same, but the difference is whether it includes "day" or not. 
However, with a little more effort, we may be able to eliminate the duplication. Let me think about this for a moment.

### Temporal.Instant.prototype.round() - smallestUnit
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Temporal/Instant/round#smallestunit_2
> A string representing the smallest unit to include in the output. The value must be one of the following: "hour", "minute", "second", "millisecond", "microsecond", "nanosecond", or their plural forms. For units larger than "nanosecond", fractional parts of the smallestUnit will be rounded according to the roundingIncrement and roundingMode settings.

### Temporal.ZonedDateTime.prototype.round() - smallestUnit
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Temporal/ZonedDateTime/round#smallestunit_2
> A string representing the smallest unit to include in the output. The value must be one of the following: "day", "hour", "minute", "second", "millisecond", "microsecond", "nanosecond", or their plural forms. For units larger than "nanosecond", fractional parts of the smallestUnit will be rounded according to the roundingIncrement and roundingMode settings.

nabetti1720: I think there is a way to delegate error handling to jiff. Even if you pass `day` as the `smallestUnit` of `Instant.prototype.round()`, jiff will probably return an appropriate error.

richarddavison (approved):
NICE! LGTM!

PR #1552, opened 2026-05-11
fix(lambda): drop error type prefix from errorMessage to match Node.js shape (#1519)

### LINKED ISSUE #1519: Bug when handling lambda error
When throwing an error from a Lambda handler, the error object is different in **Node.js** and **LLRT**:
```
export const handler = async () => {
  throw new TypeError("something went wrong");
};
```

When running in **Node24** we get an error object with:
```
{
  "errorType": "TypeError",
  "errorMessage": "something went wrong"
}
```

But on **LLRT** is:
```
{
  "errorType": "TypeError",
  "errorMessage": "TypeError: something went wrong"
}
```

I'm not sure if this is a bug or if it's intended, we just found out because we needed to parse that error message.

SAY-5:
### Issue # (if available)

Closes #1519.

### Description of changes

`post_error` in `llrt_core/src/runtime_client.rs` was constructing the Lambda `errorMessage` as `"{className}: {message}"`. Node's Lambda runtime exposes the class as `errorType` separately and sends only the bare message in `errorMessage`, so parsers expecting the Node shape broke under LLRT.

Drops the prefix from the message: `errorType` continues to receive the class name and `errorMessage` is now `ex.message()`. The `Error`/`Value` arms already prefix `"Error: "` for value-style throws, left those untouched as they don't have a class name to redundantly include.

### Checklist

No unit tests in this file today; the bug repro is the exact two snippets in the issue (one Node, one LLRT). Locally rust-syntax-checks cleanly; full `cargo check` is blocked by the compression-dictionary build script which is unrelated to this diff.

### DIFF
--- a/llrt_core/src/runtime_client.rs
+++ b/llrt_core/src/runtime_client.rs
@@ -495,10 +495,12 @@ async fn post_error<'js>(
                 .unwrap_or(None)
                 .unwrap_or(String::from("Error"));
 
-            let mut str = String::with_capacity(100);
-            str.push_str(&error_name);
-            str.push_str(": ");
-            str.push_str(&ex.message().unwrap_or_default());
+            // Node's Lambda runtime exposes the error class as `errorType`
+            // and the bare message as `errorMessage`. Previously the message
+            // was prefixed with "{name}: " which broke parsers expecting the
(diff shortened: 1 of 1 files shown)


richarddavison (approved):
Thank you!

PR #1661, opened 2026-07-11
feat(buffer): Improving Web standards for Blobs

nabetti1720:
### Issue # (if available)

n/a

### Description of changes

We re-enabled several tests within the WPT suite that had previously been excluded, and improved them so that the previously failing test cases now pass.

#### FileAPI/blob > should pass Blob-constructor.any.js tests:
```
Error: [Blob constructor with no arguments, without 'new'] assert_throws_js: function "function() { var blob = Blob(); }" did not throw
Error: [Blob constructor with undefined as first argument] Failed to construct 'Blob': The provided value cannot be converted to a sequence.
Error: [Getters and value conversions should happen in order until an exception is thrown.] assert_array_equals: expected property 1 to be "length getter" but got "Symbol.iterator" (expected array ["Symbol.iterator", "length getter", "length valueOf", "0 getter", "0 toString", "length getter", "length valueOf", "1 getter"] got ["Symbol.iterator", "Symbol.iterator", "length getter", "length valueOf", "0 getter", "length getter", "length valueOf", "1 getter"])
Error: [Changes to the blobParts array should be reflected in the returned Blob (pop).] assert_equals: expected 4 but got 13
```

The following error occurred because MessageChannel is not implemented. While the test suite will continue if you skip this test, there is room for improvement in the subsequent test cases.

```
Error: [Passing a FrozenArray as the blobParts array should work (FrozenArray<MessagePort>).] MessageChannel is not defined
```

#### FileAPI/blob > should pass Blob-slice.any.js tests: 
```
Error: [no-argument Blob slice] assert_equals: expected 4 but got 0
Error: [undefined start/end Blob slice] promise_test: Unhandled rejection with value: object "TypeError: Error converting from js 'undefined' into type 'f64'"
Error: [null type Blob slice] promise_test: Unhandled rejection with value: object "TypeError: Error converting from js 'null' into type 'string'"
Error: [Slicing test: slice (1,1).] assert_equals: expected 2 but got 4
```

#### FileAPI/blob > should pass Blob-constructor-endings.any.js tests:
```
Error: [Invalid "endings" value: null] assert_throws_js: Blob constructor should throw function "() => new Blob([], {endings: value})" did not throw
Error: [The "endings" options property is used] expected 'endings' to be either 'transparent' or 'native'
```

### Checklist

### DIFF
--- a/libs/llrt_utils/src/class.rs
+++ b/libs/llrt_utils/src/class.rs
@@ -61,9 +61,9 @@ where
         let iterator_proto = &BasePrimordials::get(ctx)?.prototype_iterator;
         if proto.get_prototype().as_ref() != Some(iterator_proto) {
             proto.set_prototype(Some(iterator_proto))?;
-            let next_fn: Function = proto.get("next")?;
+            let next_fn: Function = proto.get(PredefinedAtom::Next)?;
             proto.prop(
-                "next",
+                PredefinedAtom::Next,
                 Property::from(next_fn)
                     .writable()
                     .enumerable()
--- a/libs/llrt_utils/src/string.rs
+++ b/libs/llrt_utils/src/string.rs
@@ -15,3 +15,12 @@ pub fn get_coerced_string(value: &Value<'_>) -> Option<String> {
     };
     None
 }
+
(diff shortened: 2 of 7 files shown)


richarddavison (approved):
LGTM!

PR #1689, opened 2026-07-28
fix: cookie handling not working with Hono framework

### LINKED ISSUE #1638: getCookie in hono not working with latest commits
The `getCookie` method in the hono framework no longer works using the latest commits in main.  It worked with 8.1-beta. It stopped working with a commit sometime after 2026-02-26.

I've pushed a repro here:

https://github.com/perpil/llrt-hono-repro

Main code in: https://github.com/perpil/llrt-hono-repro/blob/main/src/index.js

perpil:
### Issue # (if available)

Resolves https://github.com/awslabs/llrt/issues/1638

### Description of changes

Exempt cookie headers from being stripped, they are ok for server-side runtimes.

### Checklist

### DIFF
--- a/tests/wpt/fetch.api.request.test.ts
+++ b/tests/wpt/fetch.api.request.test.ts
@@ -9,4 +9,12 @@ runSuite(import.meta.url, runTestDynamic, [
   "request-cache-no-store.any.js", // ReferenceError: promise_test is not defined
   "request-cache-only-if-cached.any.js", // ReferenceError: promise_test is not defined
   "request-cache-reload.any.js", // ReferenceError: promise_test is not defined
+  [
+    "request-headers.any.js",
+    [
+      '[Adding invalid request header "Cookie: KO"] assert_equals: expected (object) null but got (string) "KO"',
+      '[Adding invalid request header "Cookie2: KO"] assert_equals: expected (object) null but got (string) "KO"',
+      '[Check that request constructor is filtering headers provided as init parameter] assert_equals: expected (object) null but got (string) "potato"',
+    ],
+  ],
 ]);
--- a/modules/llrt_fetch/src/headers.rs
+++ b/modules/llrt_fetch/src/headers.rs
@@ -611,8 +611,6 @@ fn is_forbidden_request_header(name: &str) -> bool {
         || name == ACCESS_CONTROL_REQUEST_METHOD
         || name == CONNECTION
         || name == CONTENT_LENGTH
-        || name == COOKIE
-        || name == "cookie2"
         || name == DATE
         || name == "dnt"
         || name == EXPECT
--- a/tests/unit/fetch.request.test.ts
+++ b/tests/unit/fetch.request.test.ts
@@ -437,4 +437,13 @@ describe("Request class", () => {
       });
     }).toThrow();
   });
+
+  it("should preserve Cookie headers when initializing a Request and setting on request.headers", () => {
+    const req = new Request("https://example.com", {
+      headers: { cookie: "foo=bar" },
+    });
+    expect(req.headers.get("cookie")).toEqual("foo=bar");
+    req.headers.append("cookie", "baz=qux");
+    expect(req.headers.get("cookie")).toEqual("foo=bar; baz=qux");
+  });
 });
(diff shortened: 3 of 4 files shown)


richarddavison:
@perpil thanks for the PR but this fails WPT tests, can you PTAL?

perpil:
This test is expected to fail now that cookies are enabled:

> /home/runner/work/llrt/llrt/bundle/js/__tests__/wpt/fetch.api.request.test.js 
fetch/api/request > should pass request-headers.any.js tests
Error: [Adding invalid request header "Cookie: KO"] assert_equals: expected (object) null but got (string) "KO"

I believe the other failures are also expected, but don't have enough familiarity with the WPT tests to know for sure.

Is the correct fix to add `header-setcookie.any.js` to this [file](https://github.com/awslabs/llrt/blob/main/tests/wpt/fetch.api.headers.test.ts)?

nabetti1720:
Personally, I am reluctant to exclude the entire file containing the tests that are currently failing. That is because there are other tests I want to pass.

It would be great if the test runner could be improved to allow specifying files and test cases for exclusion.

``` javascript
runSuite(import.meta.url, runTestDynamic, [
  ["request-headers.any.js", 'Adding invalid request header "Cookie: KO"'], // Exclude specific test cases.
  ["exclude-all-tests.any.js"], // Exclude all tests contained in that file.
]);
```

perpil:
Ok, that makes sense @nabetti1720 I had the filename wrong and didn't realize how many tests were in there. I'll see what I can do to have it skip certain errors.

perpil:
I've added the ability to skip individual tests in a file based on error string or regex and modified it to skip the 3 errors that are expected. Wpt tests are passing, should be good to review now.

nabetti1720:
@richarddavison The modifications to the harness conflict with my PR, so please prioritize this one. :)

tests/wpt/fetch.api.request.test.ts
```diff
@@ -9,4 +9,12 @@ runSuite(import.meta.url, runTestDynamic, [
...
   "request-cache-reload.any.js", // ReferenceError: promise_test is not defined
+  [
+    "request-headers.any.js",
+    [
+      '[Adding invalid request header "Cookie: KO"] assert_equals: expected (object) null but got (string) "KO"',
+      '[Adding invalid request header "Cookie2: KO"] assert_equals: expected (object) null but got (string) "KO"',
+      '[Check that request constructor is filtering headers provided as init parameter] assert_equals: expected (object) null but got (string) "potato"',
+    ],
+  ],
 ]);
```
richarddavison, line 20: Can we do a more controlled way of skipping tests? I wonder if this is an XY problem. Why doesn't WPT allow cookie or cookie2 as request header? Is not none spec compliant?
https://fetch.spec.whatwg.org/#cookie-header
https://fetch.spec.whatwg.org/#set-cookie-header

richarddavison (changes requested):

tests/wpt/fetch.api.request.test.ts
perpil: As I understand WPT is meant for browser based tests. Cookie headers in the browser aren't allowed on Request objects. Since one of the use cases for Node and llrt is  server-side, some of the browser cookie related tests are expected to fail.  The WPT tests in node seem to have a similar mechanism in the test harness to allow skipping tests: https://github.com/nodejs/node/blob/main/test/wpt/README.md

There are a couple of ways of solving this including commenting out the tests. Since the tests aren't named, I couldn't find a way of omitting tests by name, but let me know if you see a better way. There are multiple tests cases that emit `'[Check that request constructor is filtering headers provided as init parameter] assert_equals: expected (object) null but got (string) "potato"'` which is unfortunate because it isn't limited to a single test. I believe the way I've implemented it is the closest to how [Node does it](https://github.com/nodejs/node/blob/main/test/wpt/README.md).

richarddavison: Right I don't think there is a cleaner way either since we don't have names. Maybe we should suggest something upstream but that's certainly out of scope for this PR. Nice work

richarddavison (approved):

Issue #968, opened 2025-05-03
WPTs for WebCryptoAPI


panva:
From my [cursory testing](https://github.com/panva/jose/issues/802#issuecomment-2848516533) the WebCryptoAPI interface and algorithm implementations are not even remotely correct.

There are too many issues to enumerate and open individual issues for. To that end, please consider enabling the WebCryptoAPI [Web Platform Tests](https://github.com/web-platform-tests/wpt/tree/ccaadde016187b16e607c6c6ae65e37089c452c5/WebCryptoAPI) to bring the llrt SubtleCrypto implementation up to par with [existing implementations](https://wpt.fyi/results/WebCryptoAPI?label=experimental&label=master&product=node.js&product=deno&product=chrome&product=edge&product=firefox&product=safari&product=flow&product=ladybird&view=subtest&q=node.js%3A%21missing)

richarddavison:
Thanks for the report! There are inconsistencies. Some are due to limitations in Rust Crypto libraries, other are due pure incompatibilities. These should all be fixed and validated through WPT tests. I saw a lot of tests are failing due to `Received an instance of Object` which is probably a good first step.

panva:
> Thanks for the report! There are inconsistencies. Some are due to limitations in Rust Crypto libraries, other are due pure incompatibilities. These should all be fixed and validated through WPT tests. I saw a lot of tests are failing due to `Received an instance of Object` which is probably a good first step.

That's due to https://github.com/awslabs/llrt/issues/967

panva:
I do not have any WPTs running in my project. My project merely uses Web APIs and is runtime-agnostic with a bunch of test vectors and a test suite I can execute in any js runtime, i got it running in llrt and saw a ton of failures which can only be because the APIs are not implementing the standards correctly.

As I did with workerd, deno, and bun, my contribution to llrt can mostly only be time spent reporting bugs and interop issues as its Web API support matures.

richarddavison:
Thank you for that clarification. We'll add WebCrypto WPTs to the harness and fix compatibility. Your time spend reporting and providing feedback is highly appreciated!

PR #1704, opened 2026-08-02
fix(crypto): Add WPT tests and improve deriveBits and deriveKey compatibility

nabetti1720:
### Issue # (if available)

Related #968

### Description of changes

Regarding `deriveBits` and `deriveKey`, we re-enabled the WPTs (which had been disabled) and resolved the issues causing the tests to fail.

- Updated to return the appropriate DOMException and enhanced parameter normalization and validation
- Extract the raw private key bytes from PKCS#8 instead of storing the entire PKCS#8 encoding, ensuring a consistent internal key representation across import formats
- Other minor corrections

These changes resolved the following errors.

#### WebCryptoAPI/derive_bits_keys > should pass cfrg_curves_bits_curve25519.https.any.js tests:
```
Error: [X25519 key derivation checks for all-zero value result with a key of order 0] assert_equals: Should throw correct error, not Error: Invalid key. expected "OperationError" but got "Error"
Error: [X25519 key derivation checks for all-zero value result with a key of order 0] assert_equals: Should throw correct error, not DataError: Invalid key. expected "OperationError" but got "DataError"
Error: [X25519 mixed case parameters] assert_unreached: deriveBits failed with error NotSupportedError: Algorithm not supported Reached unreachable code
Error: [X25519 mismatched algorithms] assert_equals: Should throw correct error, not TypeMismatchError: Key algorithm must be X25519 expected "InvalidAccessError" but got "TypeMismatchError"
Error: [X25519 no deriveBits usage for base key] assert_equals: Should throw correct error, not Error: CryptoKey with 'X25519', doesn't support 'deriveBits' expected "InvalidAccessError" but got "Error"
Error: [X25519 public property value is a private key] assert_unreached: deriveBits succeeded but should have failed with InvalidAccessError Reached unreachable code
Error: [X25519 asking for too many bits] assert_unreached: deriveBits succeeded but should have failed with OperationError Reached unreachable code
```

#### WebCryptoAPI/derive_bits_keys > should pass cfrg_curves_keys_curve25519.https.any.js tests:
```
Error: [X25519 no deriveKey usage for base key] assert_unreached: deriveKey succeeded but should have failed with InvalidAccessError Reached unreachable code
```

#### WebCryptoAPI/derive_bits_keys > should pass derived_bits_length.https.any.js tests:
```
Error: [HKDF derivation with 230 as 'length' parameter] assert_unreached: deriveBits should have thrown an OperationError exception. Reached unreachable code
Error: [HKDF derivation with null as 'length' parameter] assert_equals: deriveBits correctly threw OperationError: Error converting from js 'null' into type 'f64' expected "OperationError" but got "TypeError"
Error: [HKDF derivation with omitted as 'length' parameter] assert_equals: deriveBits correctly threw OperationError: Error calling function with 2 argument(s) while 3 where expected expected "OperationError" but got "TypeError"
Error: [PBKDF2 derivation with 230 as 'length' parameter] assert_unreached: deriveBits should have thrown an OperationError exception. Reached unreachable code
Error: [ECDH derivation with 384 as 'length' parameter] assert_unreached: deriveBits should have thrown an OperationError exception. Reached unreachable code
Error: [ECDH derivation with 230 as 'length' parameter] assert_array_equals: Derived bits do not match the expected result. expected property 28 to be 56 but got 59 (expected array object "90,0,220,87,31,26,232,151,28,227,35,250,17,131,137,203,95,65,196,56" got object "90,0,220,87,31,26,232,151,28,227,35,250,17,131,137,203,95,65,196,59")
```

#### WebCryptoAPI/derive_bits_keys > should pass ecdh_bits.https.any.js tests:
```
Error: [P-521 mismatched curves] assert_equals: Should throw correct error, not Error: ECDH curve must be same as baseKey expected "InvalidAccessError" but got "Error"
Error: [P-521 public property of algorithm is not an ECDSA public key] assert_unreached: deriveBits succeeded but should have failed with InvalidAccessError Reached unreachable code
Error: [P-521 public property value is a private key] assert_equals: Should throw correct error, not DataError: Invalid key expected "InvalidAccessError" but got "DataError"
```

#### WebCryptoAPI/derive_bits_keys > should pass hkdf.https.any.js tests:
```
Error: [Derived key of type name: AES-CBC length: 128  using short derivedKey, normal salt, SHA-384, with normal info with bad hash name SHA384] assert_unreached: bad hash name should have thrown an NotSupportedError Reached unreachable code
Error: [Derived key of type name: AES-CBC length: 128  using short derivedKey, normal salt, SHA-384, with normal info with wrong (ECDH) key] assert_equals: deriveKey with wrong (ECDH) key correctly threw InvalidAccessError: Key algorithm must be HKDF expected "InvalidAccessError" but got "TypeMismatchError"
```

### Checklist


nabetti1720:
It is just an intermittent error unrelated to this PR.

```
--- wpt_errors.baseline	2026-08-03 10:24:18.158787097 +0000
+++ wpt_errors.txt	2026-08-03 10:24:56.102375781 +0000
@@ -8,6 +8,7 @@
 fetch/api/basic > should pass referrer.any.js tests
 fetch/api/basic > should pass request-headers-case.any.js tests
 fetch/api/basic > should pass request-referrer.any.js tests
+fetch/api/basic > should pass response-null-body.any.js tests
 streams/piping > should pass throwing-options.any.js tests
 streams/transferable > should pass transform-stream-members.any.js tests
 streams/transform-streams > should pass strategies.any.js tests

WPT results differ from the committed baseline (diff above).
If the change is intended, commit the regenerated wpt_errors.txt.
```

modules/llrt_crypto/src/subtle/derive_bits.rs
```diff
@@ -0,0 +1,189 @@
...
+) -> impl Future<Output = Result<ArrayBuffer<'js>>> + 'js {
+    let prepared = prepare_derive_bits(&ctx, algorithm);
+
+    async move {
+        let algorithm = prepared?;
+
+        let base_key = base_key.borrow();
+        base_key
+            .check_validity("deriveBits")
+            .or_throw_dom(&ctx, "")?;
```
richarddavison: Maybe empty string here should be changed so that `or_throw_dom` takes an optional string instead of this if we use this in more places

modules/llrt_crypto/src/subtle/derive_bits.rs
```diff
@@ -0,0 +1,189 @@
...
+        let length = parse_derive_bits_length(&ctx, length)?;
+        let bytes = derive_bits(&ctx, &algorithm, &base_key, length)?;
+
+        ArrayBuffer::new(ctx, bytes)
+    }
+}
+
+fn prepare_derive_bits<'js>(ctx: &Ctx<'js>, algorithm: Value<'js>) -> Result<DeriveAlgorithm> {
+    DeriveAlgorithm::from_js(ctx, algorithm)
+}
```
richarddavison: why not just call  `DeriveAlgorithm::from_js` at callsite? This function serves no purpose

modules/llrt_crypto/src/subtle/key_algorithm.rs
```diff
@@ -1418,7 +1422,14 @@ pub fn extract_sha_hash<'js>(ctx: &Ctx<'js>, obj: &Object<'js>) -> Result<HashAl
...
+    let hash = hash.as_str();
+    if !matches!(hash, "SHA-1" | "SHA-256" | "SHA-384" | "SHA-512") {
+        return Err(DOMException::not_supported_error(
+            ctx,
+            "Unsupported hash algorithm",
+        ));
+    };
+    HashAlgorithm::try_from(hash)
         .map_err(NotSupportedError)
         .or_throw_dom(ctx, "")
```
richarddavison: don't we duplicate here? we already check in  HashAlgorithm::try_from(hash) so no need to do pre-match

richarddavison (changes requested):
Thanks a lot! Some minor changes!

modules/llrt_crypto/src/subtle/key_algorithm.rs
nabetti1720: Due to specification requirements, a preliminary check with conditions different from those of `try_from()` is necessary. Specifically, the algorithm must be one of SHA-1, SHA-128, SHA-256, or SHA-512, and the hyphen cannot be omitted.

Therefore, I decided to create a new method, `HashAlgorithm::from_strict_str()`.

modules/llrt_crypto/src/subtle/derive_bits.rs
nabetti1720: The intention was to align the structure with `deriveKey`, but you're right—it's unnecessary. I'll remove it.

modules/llrt_crypto/src/subtle/derive_bits.rs
nabetti1720: Even if we were to accept an `Option<String>`, we would ultimately end up with calls like `or_throw_dom(&ctx, None)` and `or_throw_dom(&ctx, Some("message"))`, which I don't think feels quite right.

It might be better to use something like `or_throw_dom(&ctx)` and `or_throw_dom_with_msg(&ctx, "message")`.

I have made revisions based on this approach, so please take a look.

nabetti1720:
The revisions are complete. Please review them, as there may be some points that do not fully align with your preferences.

richarddd (approved):

richarddavison (approved):

PR #1714, opened 2026-08-10
fix(stream): preserve TransformStream backpressure

dills122:
### Issue # (if available)

No issue filed. This focused conformance bug was identified through the committed TransformStream WPT failure baseline.

### Description of changes

Preserve `TransformStream` backpressure when the readable queue reaches its high water mark. Before this change, one additional queued write could transform before the readable side was consumed.

The controller resolved its existing backpressure-change promise when pressure became active but kept that settled promise installed. The next queued write therefore waited on an already-fulfilled promise and transformed immediately, overfilling the readable queue.

This change follows the Streams Standard's `TransformStreamSetBackpressure` operation by resolving the previous promise, installing a fresh pending promise, and then updating the backpressure state. Readable-side pull transitions now use the same shared operation.

Streams Standard: https://streams.spec.whatwg.org/#transform-stream-set-backpressure

### Validation

- Added `readable_high_water_mark_applies_backpressure`; it fails before the fix with `0,1,2,3` transformed and passes after the fix with only `0,1,2` transformed until a read occurs
- `env RUSTUP_TOOLCHAIN=1.95.0 make fix`
- `env RUSTUP_TOOLCHAIN=1.95.0 make check`
- `cargo +1.95.0 test -p llrt_stream_web` — 13 passed
- `cargo +1.95.0 fmt --all -- --check`
- `cargo +nightly build`
- Runtime reproduction matches Node 22: nine chunks transform at readable HWM 9, and the tenth resumes after one read
- Fork CI: all 32 jobs pass, including WPT and the complete platform/crypto matrix

### Scope notes

- Does not address strategy value conversion
- Does not address the separate `controller.terminate()` error-propagation panic

### Checklist

- [x] Added a focused Rust regression test
- [x] Made sure the code adds no warnings with `make check`
- [x] Type changes are not required
- [x] Documentation changes are not required

### DIFF
--- a/modules/llrt_stream_web/src/transform/controller.rs
+++ b/modules/llrt_stream_web/src/transform/controller.rs
@@ -156,7 +156,7 @@ pub(super) fn transform_stream_default_controller_enqueue<'js>(
 
     let current_bp = stream_class.borrow().backpressure;
     if has_backpressure != current_bp {
-        transform_stream_set_backpressure(stream_class, true)?;
+        transform_stream_set_backpressure(&ctx, stream_class, true)?;
     }
 
     Ok(())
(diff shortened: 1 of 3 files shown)


nabetti1720:
It's a very minor detail, but you might need to commit the updated wpt_errors.txt.

richarddavison (approved):
Thank you. LGTM!

PR #1715, opened 2026-08-11
feat: expose bundled AWS SDK version

### LINKED ISSUE #1446: Determining the AWS SDK version at runtime?
Is there a way to determine the aws sdk version the llrt binary was built with at runtime? Doing something like this doesn't work:

```javascript
import packageJson from '@aws-sts/client/package.json' with { type: 'json' };
const sdkVersion = packageJson.version;
```

Also, super minor, but worth mentioning, the latest release on GitHub `v0.8.1-beta` (arm64 zip) reports `v0.8.0-beta` when you use `process.version`. If you build locally with the latest code it reports `v0.8.1-beta` so something might have been missed that sets the version number during the release process. Strangely, the [commit that changes it to 8.1](https://github.com/awslabs/llrt/commit/6d460c5cec39380ebbd5d76cc34af633ea7015b3) seems to be part of the 8.1 release.

dills122:
## Summary

- expose the bundled AWS SDK version as `process.versions["@aws-sdk"]`
- generate the value from the installed `@aws-sdk/client-sts` package during SDK bundling
- set the property directly while `llrt_process` constructs `process.versions`
- omit the property from `NONE` bundles and Cargo `no-sdk` builds
- document the optional property and cover presence, absence, and build integration in tests

## Why

LLRT does not bundle AWS SDK package metadata, so applications cannot reliably import a client package's `package.json` to determine the bundled SDK version. `process.versions` already reports LLRT and Node compatibility versions and provides a runtime-visible location for this metadata.

Closes #1446.

## Implementation

The JavaScript build writes `bundle/js/.aws-sdk-version` only after a standard or full SDK bundle succeeds. A build script owned by `llrt_process` reads that marker and embeds it as `LLRT_AWS_SDK_VERSION` when compiling the process module. `llrt_process::init` then adds the value while constructing `process.versions`.

The existing `no-sdk` feature is forwarded through `llrt_core` and `llrt_modules` to `llrt_process`. This makes Cargo `no-sdk` authoritative even if a marker from an earlier SDK bundle remains present.

## Review Feedback

- moved AWS SDK version handling out of `llrt_core` embedded-runtime initialization
- colocated version injection, `process.versions` construction, and tests in `modules/llrt_process`, as requested
- read `LLRT_AWS_SDK_VERSION` directly while constructing `process.versions`
- removed the fixed SDK-version assertion so future SDK upgrades do not require test changes
- retained the SDK/no-SDK safeguards while changing ownership

## Validation

- `cargo check --workspace`
- `cargo test -p llrt_process` — 3 passed, 0 failed
- `cargo test -p llrt_process --features no-sdk` — 3 passed, 0 failed
- `cargo clippy -p llrt_process --tests -- -D warnings`
- focused LLRT process JavaScript suite — 43 passed, 0 failed
- Prettier check for changed JavaScript and TypeScript files
- `git diff --check`

Runtime proof:

| Build case | Generated marker | Runtime result |
| --- | --- | --- |
| `SDK_BUNDLE_MODE=STD` | `3.1057.0` | `{"hasAwsSdk":true,"version":"3.1057.0"}` |
| `SDK_BUNDLE_MODE=NONE` | absent | `{"hasAwsSdk":false,"version":null}` |
| STD marker + Cargo `--features no-sdk` | `3.1057.0` | `{"hasAwsSdk":false,"version":null}` |

The workspace was restored to a normal STD build after the matrix, which reports `process.versions["@aws-sdk"] === "3.1057.0"`.

### DIFF
--- a/modules/llrt_process/src/lib.rs
+++ b/modules/llrt_process/src/lib.rs
@@ -99,6 +99,22 @@ fn env_proxy_setter<'js>(
     Ok(true)
 }
 
+fn create_process_versions<'js>(
+    ctx: &Ctx<'js>,
+    aws_sdk_version: Option<&str>,
+) -> Result<Object<'js>> {
+    let versions = Object::new(ctx.clone())?;
+    versions.set("llrt", VERSION)?;
+    // Node.js version - Set for compatibility with some Node.js packages (e.g. cls-hooked).
+    versions.set("node", "0.0.0")?;
+
+    if let Some(version) = aws_sdk_version {
+        versions.set("@aws-sdk", version)?;
+    }
+
+    Ok(versions)
+}
+
 #[cfg(unix)]
 fn getuid() -> u32 {
     unsafe { libc::getuid() }
@@ -143,10 +159,7 @@ pub fn init(ctx: &Ctx<'_>) -> Result<()> {
     let globals = ctx.globals();
     BasePrimordials::init(ctx)?;
     let process = Object::new(ctx.clone())?;
-    let process_versions = Object::new(ctx.clone())?;
-    process_versions.set("llrt", VERSION)?;
-    // Node.js version - Set for compatibility with some Node.js packages (e.g. cls-hooked).
-    process_versions.set("node", "0.0.0")?;
+    let process_versions = create_process_versions(ctx, option_env!("LLRT_AWS_SDK_VERSION"))?;
 
     let hr_time = Function::new(ctx.clone(), hr_time)?;
     hr_time.set("bigint", Func::from(hr_time_big_int))?;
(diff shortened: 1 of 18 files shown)


richarddavison (changes requested):
We could instead set here:
https://github.com/awslabs/llrt/blob/ff690644d0f4b341b3ec76b4e837a1f2beb5291e/modules/llrt_process/src/lib.rs#L142

modules/llrt_process/src/lib.rs
```diff
@@ -99,6 +99,22 @@ fn env_proxy_setter<'js>(
...
+    // Node.js version - Set for compatibility with some Node.js packages (e.g. cls-hooked).
+    versions.set("node", "0.0.0")?;
+
+    if let Some(version) = aws_sdk_version {
+        versions.set("@aws-sdk", version)?;
+    }
+
+    Ok(versions)
+}
+
```
richarddavison: Instead of passing option_env!("LLRT_AWS_SDK_VERSION") we should just use it in here directly:

`if let Some(version) = option_env!("LLRT_AWS_SDK_VERSION") {`

richarddavison (changes requested):
Thanks! Just a minor suggestion and we're good to go!

modules/llrt_process/src/lib.rs
```diff
@@ -296,6 +309,51 @@ mod tests {
...
 
+    #[tokio::test]
+    async fn test_process_versions_with_aws_sdk() {
+        test_async_with(|ctx| {
+            Box::pin(async move {
+                let versions = create_process_versions(&ctx, Some("3.1057.0")).unwrap();
+
+                assert_eq!(versions.get::<_, String>("llrt").unwrap(), VERSION);
+                assert_eq!(versions.get::<_, String>("node").unwrap(), "0.0.0");
+                assert_eq!(versions.get::<_, String>("@aws-sdk").unwrap(), "3.1057.0");
```
nabetti1720: Can the SDK version be specified dynamically? Otherwise, I think the tests will fail during future SDK version upgrades.

richarddavison: The version is actually injected to make the code testable. I think we can skip this test entirely and just check contains a value.

richarddavison (approved):

PR #1716, opened 2026-08-11
fix(crypto): allow RSA JWK imports without alg

dills122:
## Summary

- Allow WebCrypto RSA JWK imports when the optional `alg` member is absent.
- Restore compatibility with OIDC/Cognito-style JWKS verification, including the failure reported in panva/jose#802.
- Related to awslabs/llrt#968.

## What Changed

- Treat RSA JWK `alg` as optional during import.
- Preserve strict algorithm-family and hash validation whenever `alg` is present.
- Add a Cognito/JWKS-shaped RSA public-key regression that verifies a real signature and rejects mismatched `PS256` and `RS384` values.
- Enable the upstream WebCrypto `rsa_importKey.https.any.js` WPT.

## Root Cause

RSA JWK import required `keyData.alg` before decoding the key, although WebCrypto specifies that the member may be absent. This caused standards-compliant public JWKS keys to fail with `TypeError: keyData 'alg' property required`.

## Validation

- `cargo fmt --check`
- `cargo +nightly clippy -p llrt_crypto --all-targets --no-deps -- -D warnings`
- `cargo test -p llrt_crypto` — 23 passed
- Focused `crypto.subtle` runtime suite — 16 passed
- Upstream `rsa_importKey.https.any.js` WPT — 1 wrapper passed
- panva/jose compatibility bundle — 185 passed; the local JWKS and RSA JWK import paths pass, with 36 unrelated compatibility failures remaining

## Scope Notes

- String AlgorithmIdentifier support for parameterless AES-KW and RSA-OAEP operations remains a separate follow-up.
- Ed25519/X25519 compatibility is intentionally out of scope.
- No dependency, provider, type, documentation, lockfile, or generated-file changes are included.

## Checklist

- [x] Created unit and WPT regression coverage
- [x] Verified formatting and module-scoped Clippy checks
- [x] Confirmed no type or documentation update is required for this behavior fix

### DIFF
--- a/modules/llrt_crypto/src/subtle/key_algorithm.rs
+++ b/modules/llrt_crypto/src/subtle/key_algorithm.rs
@@ -1061,34 +1061,23 @@ fn import_rsa_key<'js>(
     let (modulus_length, public_exponent) = match format {
         KeyFormatData::Jwk(object) => {
             let kty: String = object.get_required("kty", "keyData")?;
-            let alg: String = object.get_required("alg", "keyData")?;
             if kty != "RSA" {
                 return algorithm_mismatch_error(ctx, algorithm_name);
             }
-            let prefix = &alg[..2];
-            let numeric_hash_str = match prefix {
-                "RS" => {
-                    if algorithm_name == "RSA-OAEP" {
-                        if !alg.starts_with(algorithm_name) {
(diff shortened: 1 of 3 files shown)


richarddavison (approved):
LGTM! Thanks!

PR #1717, opened 2026-08-12
fix(crypto): support string encryption algorithm identifiers

dills122:
### Issue # (if available)

Related to #968.

### Description of changes

Accept case-insensitive DOMString algorithm identifiers for parameterless AES-KW wrap/unwrap and RSA-OAEP encrypt/decrypt operations.

WebCrypto `AlgorithmIdentifier` accepts either a DOMString or an algorithm dictionary. The encryption parser previously required an object unconditionally, so valid string identifiers failed before AES-KW or RSA-OAEP could be selected. The parser now reuses the existing algorithm-name normalization path and permits omitted parameter objects only for those operations.

AES-CBC, AES-CTR, and AES-GCM continue to reject string-only inputs with `TypeError` because their required parameter dictionaries are absent. Object-based RSA-OAEP labels are also covered to ensure the parser change preserves non-empty label handling.

Relevant standard:

- [Web Cryptography Level 2 — algorithm normalization](https://www.w3.org/TR/webcrypto-2/#algorithm-normalization)

Validation:

- `cargo test -p llrt_crypto` — 23 passed
- focused LLRT `crypto.subtle` runtime suite — 20 passed
- `cargo clippy -p llrt_crypto --all-targets -- -D warnings`
- `cargo fmt --all -- --check`
- Prettier check for `tests/unit/crypto.subtle.test.ts`
- `git diff --check`
- complete fork CI matrix passed, including the successful rerun of a transient Homebrew setup failure

### Checklist

- [x] Created focused unit coverage for AES-KW, RSA-OAEP, required AES dictionaries, and RSA-OAEP labels
- [x] Ran targeted Rust and TypeScript formatting checks
- [x] Made sure the code adds no warnings with warning-denied Clippy and the CI `check` job
- [x] Confirmed type changes are not applicable
- [x] Confirmed documentation changes are not required

### DIFF
--- a/modules/llrt_crypto/src/subtle/encryption_algorithm.rs
+++ b/modules/llrt_crypto/src/subtle/encryption_algorithm.rs
@@ -33,8 +33,7 @@ impl<'js> FromJs<'js> for EncryptionAlgorithm {
 
         match name.as_str() {
             "AES-CBC" => {
-                let obj =
-                    obj.map_err(|_| Exception::throw_type(ctx, "algorithm is not an object"))?;
+                let obj = obj?;
                 let iv = obj
                     .get_required::<_, ObjectBytes>("iv", "algorithm")?
                     .into_bytes(ctx)?
@@ -50,8 +49,7 @@ impl<'js> FromJs<'js> for EncryptionAlgorithm {
                 Ok(EncryptionAlgorithm::AesCbc { iv })
             },
             "AES-CTR" => {
-                let obj =
-                    obj.map_err(|_| Exception::throw_type(ctx, "algorithm is not an object"))?;
+                let obj = obj?;
                 let counter = obj
                     .get_required::<_, ObjectBytes>("counter", "algorithm")?
                     .into_bytes(ctx)?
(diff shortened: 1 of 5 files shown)


modules/llrt_crypto/src/subtle/encryption_algorithm.rs
```diff
@@ -28,12 +28,13 @@ pub enum EncryptionAlgorithm {
...
-        let obj = value.into_object_or_throw(ctx, "algorithm")?;
-
-        let name: String = obj.get_required("name", "algorithm")?;
+        let (name, obj) = to_name_and_maybe_object(ctx, value)?;
+        let name = normalize_algorithm_name(&name);
 
         match name.as_str() {
             "AES-CBC" => {
+                let obj =
+                    obj.map_err(|_| Exception::throw_type(ctx, "algorithm is not an object"))?;
```
richarddavison: We can do this in `to_name_and_maybe_object` and just return the result out from that method.
Then we can just do `let obj = obj?;` and save duplicated code.

richarddavison (changes requested):

richarddd (approved):

richarddd (approved):

richarddavison (approved):

PR #1722, opened 2026-08-14
fix(buffer): Improved Blob compatibility

nabetti1720:
### Issue # (if available)

n/a

### Description of changes

We skipped the tests utilizing `MessageChannel` that were acting as blockers in the Blob WPT suite. As a result, we detected a new incompatibility and implemented improvements to address it.

The specific improvements are as follows.

- The options had to be read in lexicographic order.
- In the constructor sequence, the `endings` are first retrieved from the options, and the parts are processed according to the retrieved `endings`. Subsequently, it was necessary to retrieve `type` from the options.
- The option needed to allow for an `object(Property bag)`, `null`, or `undefined`.

Ultimately, the following error was resolved.

#### FileAPI/blob > should pass Blob-constructor.any.js tests:

```
Error: [options properties should be accessed in lexicographic order.] assert_array_equals: expected property 0 to be "endings" but got "type" (expected array ["endings", "type"] got ["type", "endings"])
Error: [Arguments should be evaluated from left to right.] assert_unreached: type getter should not be called. Reached unreachable code
Error: [Passing 123 for options should throw] assert_throws_js: Blob constructor should throw with invalid property bag function "() => new Blob([], arg)" did not throw
Error: [Passing null (index 0) for options should use the defaults.] promise_test: Unhandled rejection with value: object "TypeError: Failed to construct 'Blob': options is not an object"
Error: [Passing undefined (index 1) for options should use the defaults.] promise_test: Unhandled rejection with value: object "TypeError: Failed to construct 'Blob': options is not an object"
```

### Checklist

### DIFF
--- a/modules/llrt_buffer/src/blob.rs
+++ b/modules/llrt_buffer/src/blob.rs
@@ -10,6 +10,7 @@ use llrt_stream_web::{
 use llrt_utils::{
     array_buffer::shared_array_buffer_view,
     bytes::{get_lossy_string, ObjectBytes},
+    object::not_a_object_error,
     primordials::Primordial,
     result::ResultExt,
     string::get_coerced_defined_string,
@@ -227,10 +228,7 @@ impl<'js> Blob<'js> {
     ) -> Result<Self> {
         if let Some(options) = options.0.as_ref() {
             if !options.is_null() && !options.is_undefined() && options.as_object().is_none() {
-                return Err(Exception::throw_type(
-                    &ctx,
-                    "Failed to construct 'Blob': options is not an object",
-                ));
+                return Err(not_a_object_error(&ctx, "options"));
             }
         }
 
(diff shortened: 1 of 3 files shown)


modules/llrt_buffer/src/blob.rs
```diff
@@ -225,15 +225,18 @@ impl<'js> Blob<'js> {
...
-        let mut endings = EndingType::Transparent;
-        let mut mime_type = String::new();
+        if let Some(options) = options.0.as_ref() {
+            if !options.is_null() && !options.is_undefined() && options.as_object().is_none() {
+                return Err(Exception::throw_type(
+                    &ctx,
+                    "Failed to construct 'Blob': options is not an object",
+                ));
+            }
+        }
```
richarddd, line 233: Use utility for not-an-object error
https://github.com/awslabs/llrt/blob/main/libs/llrt_utils/src/object.rs#L82C8-L82C26

richarddd (changes requested):
Minor adjustment

richarddd (approved):

Issue #1738, opened 2026-08-24
Proposal: Implement "Modern Algorithms in the Web Cryptography API" (WICG specification)


panva:
The WICG is incubating a specification that adds post-quantum secure and modern cryptographic algorithms to the Web Cryptography API, much like X25519 was incubated before being merged into [Web Cryptography](https://w3c.github.io/webcrypto/): https://wicg.github.io/webcrypto-modern-algos/

Adoption is well underway, see [Implementation status](https://github.com/WICG/webcrypto-modern-algos/issues/25).

The specification's [Adoption Guidance](https://wicg.github.io/webcrypto-modern-algos/#adoption-guidance) section recommends the following implementation priorities:

1. **ML-KEM** (FIPS 203) — post-quantum key encapsulation mechanism. Combined with traditional key agreement it enables hybrid PQ/T key establishment as used in protocols like HPKE and its [PQ and PQ/T algorithms](https://datatracker.ietf.org/doc/draft-ietf-hpke-pq/). ML-KEM requires new `SubtleCrypto` methods: `encapsulateKey`, `encapsulateBits`, `decapsulateKey`, and `decapsulateBits`.
2. **ML-DSA** (FIPS 204) — post-quantum digital signature algorithm.
3. **ChaCha20-Poly1305** (RFC 8439) — widely deployed AEAD cipher used across TLS, SSH, WireGuard, HPKE.
4. **SHA-3** (FIPS 202) — `SHA3-256`, `SHA3-384`, `SHA3-512` hash functions.
5. **cSHAKE** (NIST SP 800-185) — `cSHAKE128`, `cSHAKE256` extendable-output functions used in post-quantum constructions including ML-KEM key derivation and hybrid KEM combiners. When invoked without customization parameters, cSHAKE produces output identical to SHAKE.
6. **TurboSHAKE** (RFC 9861) — `TurboSHAKE128`, `TurboSHAKE256` XOFs based on round-reduced Keccak, roughly twice as fast as SHAKE.

The specification also introduces two important new `SubtleCrypto` methods that benefit all algorithms (including those already in Web Crypto):

- **`SubtleCrypto.supports(operation, algorithm)`** — synchronous feature detection for algorithm/operation support. This is critical for enabling progressive enhancement and graceful fallback as runtimes adopt these algorithms incrementally.
[...]

PR #1743, opened 2026-08-26
fix(webcrypto): align validation and key derivation behaviour

panva:
- Align Web Crypto integer conversion, key-usage normalization, and validation ordering with the specifications.
- Validate RSA-PSS, ECDH, HKDF, PBKDF2, HMAC, RSA-OAEP, and AES-GCM parameters with the required errors and ordering.
- Compare HMAC signatures in constant time.
- Apply key-format conversion asynchronously and align import/export validation.
- Import ECDH and X25519 shared secrets as proper HKDF or PBKDF2 base keys during `deriveKey()`.

Cleanup and fixes prior to submitting new algorithms and methods (#1738)

cc @nabetti1720 @richarddd


tests/wpt/WebCryptoAPI.wrapKey_unwrapKey.test.ts
```diff
@@ -3,6 +3,9 @@ import { runSuite, runTestDynamic } from "./WebCryptoAPI.harness.js";
 runSuite(import.meta.url, runTestDynamic, [
   [
     "wrapKey_unwrapKey.https.any.js",
-    ["[setup] Key import/export format must be 'jwk','raw','spki' or 'pkcs8'"], // 'raw-secret' is not supported.
+    [
+      "[setup] Key import/export format must be 'jwk','raw','spki' or 'pkcs8'", // 'raw-secret' is not supported.
+      /(?:using .* and AES-GCM|AES-GCM.*non-extractable)/, // Upstream vectors use a 128-bit IV; only 96-bit IVs are supported.
```
nabetti1720, line 8: First of all, I would like to acknowledge the excellent work you are doing. Please let me know if you have any information regarding this point.

The comments in the RustCrypto AES-GCM crate contain a description like the following.
https://github.com/RustCrypto/AEADs/blob/138da3734f22a9f34eaa464c84a6e5f53e798f56/aes-gcm/src/lib.rs#L169-L171
```rust
/// The `NonceSize` generic parameter can be used to instantiate AES-GCM with other
/// nonce sizes, however it's recommended to use it with `typenum::U12`,
/// the default of 96-bits.
```

Regarding the upstream, while IVs other than 96 bits are not recommended, it appears that support for them is possible. To what extent does the specification require compliance? Also, how are they handled in other server-side runtimes such as Node.js, Deno, and Bun?

Please rest assured that I am not asking for this to be implemented in this PR.

panva: The 96-bit restriction here comes from LLRT's concrete backends and predates this commit, rather than from the RustCrypto crate's theoretical capabilities.

The RustCrypto backend instantiates AES-GCM with `U12` and converts the IV to `Nonce<U12>`. The Graviola backend similarly converts it to `[u8; 12]`. OpenSSL can support other IV lengths, but exposing those would make `SubtleCrypto` behavior depend on the selected LLRT provider.

Before this commit, algorithm normalization already rejected every IV whose length was not 12 bytes. This commit does not narrow the supported API, it moves the existing check after key validation so the operation has the correct validation order and `OperationError`, and records the corresponding upstream WPT exclusions.

Supporting other IV lengths would be a separate enhancement requiring the LLRT backends to be generalized consistently. I had that built but the churn was too large for me to justify.

richarddd (approved):
NICE!!! Thank you!

panva:
I don't believe the changes are responsible for the nightly `std` compilation failures.

richarddd:
> I don't believe the changes are responsible for the nightly `std` compilation failures.

It's a nightly regression

nabetti1720:
> > I don't believe the changes are responsible for the nightly `std` compilation failures.
> 
> It's a nightly regression

@richarddd Are there any remaining reasons to keep using the nightly version?

richarddd:
> @richarddd Are there any remaining reasons to keep using the nightly version?

@nabetti1720 we have to in order to use `build-std`

PR #1747, opened 2026-08-30
Modern Algorithms for the Web Cryptography API

panva:
Followup to #1743 this implements a subset of https://wicg.github.io/webcrypto-modern-algos/ + https://github.com/WICG/webcrypto-modern-algos/pull/67

- `ML-KEM` + encap/decap methods
- Hybrid KEMs
- `ML-DSA`
- `ChaCha20-Poly1305`
- `SHA-3` as digest only
- `cSHAKE` as digest only
- `TurboSHAKE` as digest only
- `SubtleCrypto.supports(operation, algorithm)`
- `SubtleCrypto.prototype.getPublicKey(key, keyUsages)`
- `raw-{secret,public,private,seed}` key formats and `raw` to `raw-secret/raw-public` aliases for existing algorithms

Refs: #1738

### DIFF
--- a/modules/llrt_crypto/Cargo.toml
+++ b/modules/llrt_crypto/Cargo.toml
@@ -32,8 +32,11 @@ _modern-webcrypto = [
   "keccak",
   "ml-dsa",
   "ml-kem",
+  "p256",
+  "p384",
   "sha3",
   "sponge-cursor",
+  "x25519-dalek",
 ]
 
 # Internal feature: enables RustCrypto dependencies for key import/export
@@ -135,11 +138,13 @@ rsa = { version = "0.10.0-rc.18", features = [
   "encoding",
 ], default-features = false, optional = true }
 p256 = { version = "0.14", features = [
+  "alloc",
   "ecdh",
   "ecdsa",
   "pkcs8",
 ], default-features = false, optional = true }
 p384 = { version = "0.14.0-rc.15", features = [
+  "alloc",
   "ecdh",
   "ecdsa",
   "pkcs8",
--- a/modules/llrt_crypto/Cargo.toml
+++ b/modules/llrt_crypto/Cargo.toml
@@ -25,6 +25,9 @@ _subtle-full = [
   "x25519-dalek",
 ]
 
+# Internal feature: enables modern Web Cryptography algorithms
+_modern-webcrypto = ["sha3"]
+
 # Internal feature: enables RustCrypto dependencies for key import/export
 _rustcrypto = [
   "_subtle-full",
@@ -45,12 +48,12 @@ _rustcrypto = [
   "sha1",
 ]
 
-crypto-rust = ["_rustcrypto"]
-crypto-ring = ["ring"]
-crypto-ring-rust = ["ring", "_rustcrypto"]
-crypto-graviola = ["graviola"]
-crypto-graviola-rust = ["graviola", "_rustcrypto"]
-crypto-openssl = ["openssl-sys", "openssl", "_subtle-full"]
+crypto-rust = ["_rustcrypto", "_modern-webcrypto"]
+crypto-ring = ["ring", "_modern-webcrypto"]
+crypto-ring-rust = ["ring", "_rustcrypto", "_modern-webcrypto"]
+crypto-graviola = ["graviola", "_modern-webcrypto"]
+crypto-graviola-rust = ["graviola", "_rustcrypto", "_modern-webcrypto"]
+crypto-openssl = ["openssl-sys", "openssl", "_subtle-full", "_modern-webcrypto"]
 
 [dependencies]
 crc32c = { version = "0.6", default-features = false }
@@ -127,6 +130,7 @@ pkcs8 = { version = "0.11", default-features = false, features = [
   "alloc",
 ], optional = true }
 sha1 = { version = "0.11", features = ["oid"], default-features = false, optional = true }
+sha3 = { version = "0.11.0-rc.8", default-features = false, optional = true }
 spki = { version = "0.8", features = [
   "alloc",
 ], default-features = false, optional = true }
--- a/modules/llrt_crypto/Cargo.toml
+++ b/modules/llrt_crypto/Cargo.toml
@@ -123,12 +123,12 @@ elliptic-curve = { version = "0.14", features = [
 hkdf = { version = "0.13", default-features = false, optional = true }
 hmac = { version = "0.13", default-features = false, optional = true }
 keccak = { version = "0.2", default-features = false, optional = true }
-ml-dsa = { version = "0.1.1", features = [
+ml-dsa = { version = "0.1", features = [
   "alloc",
   "pkcs8",
   "rand_core",
 ], default-features = false, optional = true }
-ml-kem = { version = "0.3.2", features = [
+ml-kem = { version = "0.3", features = [
   "alloc",
   "pkcs8",
 ], default-features = false, optional = true }
@@ -160,7 +160,7 @@ pkcs8 = { version = "0.11", default-features = false, features = [
 ], optional = true }
 sponge-cursor = { version = "0.1", default-features = false, optional = true }
 sha1 = { version = "0.11", features = ["oid"], default-features = false, optional = true }
-sha3 = { version = "0.11.0-rc.8", default-features = false, optional = true }
+sha3 = { version = "0.11", default-features = false, optional = true }
 spki = { version = "0.8", features = [
   "alloc",
 ], default-features = false, optional = true }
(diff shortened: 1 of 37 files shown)


modules/llrt_crypto/src/subtle/encapsulation.rs
```diff
@@ -0,0 +1,230 @@
...
+}
+
+fn check_encapsulation_key(
+    ctx: &Ctx<'_>,
+    key: &CryptoKey<'_>,
+    algorithm: EncapsulationAlgorithm,
+    usage: &str,
+    kind: KeyKind,
+) -> Result<()> {
+    let matches_algorithm = matches!(
```
richarddd:
```rust
let matches_algorithm = match (algorithm, &key.algorithm) {
    (EncapsulationAlgorithm::MlKem(a), KeyAlgorithm::MlKem(b)) => a == *b,
    (EncapsulationAlgorithm::HybridKem(a), KeyAlgorithm::HybridKem(b)) => a == *b,
    _ => false,
};
```

richarddd (changes requested):
General feedback (at first glance):
1. There seems to be a lot of duplication, mainly from utility metods that exists in other crypto/provider files (required_object etc)
2.  Rather than relying on string matching in place, can we move and extend/reuse modules in isolation such as  HashAlgorithm::try_from... which enforces separation of concern.
3. Avoid specififing full package names ` let result = rquickjs::Object::new(ctx.clone())?; `could be  `let result = Object::new(ctx.clone())?;`

panva:
@richarddd thank you, let me get it in better shape

modules/llrt_crypto/src/subtle/key_algorithm.rs
```diff
@@ -968,7 +1452,41 @@ impl KeyAlgorithm {
                 &mut private_usages,
                 &mut public_usages,
             )?,
-            _ => return algorithm_not_supported_error(ctx),
+            _ => {
+                if let Ok(variant) = MlDsaVariant::try_from(algorithm_name) {
```
nabetti1720: If you write it this way, I think you can maintain the flow established so far.
```rust
 "ML-DSA-44" | "ML-DSA-65" | "ML-DSA-87" => {
    let variant =  MlDsaVariant::try_from(algorithm_name).or_throw_dom(&ctx)?;
    from_ml_dsa(
        ctx,
        mode,
        algorithm_name,
        variant
        &usages,
        &mut private_usages,
        &mut public_usages,
    )?
},
```

modules/llrt_crypto/Cargo.toml
```diff
@@ -102,17 +122,29 @@ elliptic-curve = { version = "0.14", features = [
 ], default-features = false, optional = true }
 hkdf = { version = "0.13", default-features = false, optional = true }
 hmac = { version = "0.13", default-features = false, optional = true }
+keccak = { version = "0.2", default-features = false, optional = true }
+ml-dsa = { version = "0.1.1", features = [
```
nabetti1720: LLRT uses a more relaxed version of the specification.

modules/llrt_crypto/Cargo.toml
```diff
@@ -126,7 +158,9 @@ pbkdf2 = { version = "0.13", default-features = false, optional = true }
 pkcs8 = { version = "0.11", default-features = false, features = [
   "alloc",
 ], optional = true }
+sponge-cursor = { version = "0.1", default-features = false, optional = true }
 sha1 = { version = "0.11", features = ["oid"], default-features = false, optional = true }
+sha3 = { version = "0.11.0-rc.8", default-features = false, optional = true }
```
nabetti1720: A stable version of `sha3` has already been released.

modules/llrt_crypto/src/subtle/digest.rs
```diff
@@ -32,27 +104,187 @@ fn prepare_digest<'js>(
...
-    } else if let Some(name) = algorithm.get_optional::<_, String>("name")? {
-        name
-    } else {
-        return Err(rquickjs::Exception::throw_type(
-            ctx,
-            "Algorithm 'name' property required",
-        ));
+) -> Result<(DigestAlgorithm, Vec<u8>)> {
+    let (name, object) = to_name_and_maybe_object(ctx, algorithm)?;
+    let name = match DigestAlgorithmName::try_from(name.as_str()) {
```
nabetti1720: Can't it be written like this?
```rust
let Some(name) = DigestAlgorithmName::try_from(name.as_str()) else {
     return algorithm_not_supported_error(ctx);
};

File: modules/llrt_crypto/src/subtle/key_algorithm.rs [hunk 6]
```diff
@@ -661,6 +733,390 @@ fn from_hmac<'js>(
...
+                    .or_throw_dom(ctx)?;
+                *kind = KeyKind::Private;
+            },
+            KeyFormatData::Pkcs8(bytes) => {
+                let bytes = bytes.as_bytes(ctx)?;
+                validate_ml_private_key_info(
+                    ctx,
+                    bytes,
+                    match variant {
+                        MlDsaVariant::MlDsa44 => "2.16.840.1.101.3.4.3.17",
```
nabetti1720: Hasn't it been defined in `const_oid` yet?

nabetti1720 (commented):
I would like to offer some comments on a few points that caught my attention.

modules/llrt_crypto/src/subtle/key_algorithm.rs
```diff
@@ -661,6 +733,390 @@ fn from_hmac<'js>(
...
+                *data = modern::import_ml_dsa_private_key(variant, bytes.as_bytes(ctx)?, false)
+                    .or_throw_dom(ctx)?;
+                *kind = KeyKind::Private;
+            },
+            KeyFormatData::Pkcs8(bytes) => {
+                let bytes = bytes.as_bytes(ctx)?;
+                validate_ml_private_key_info(
+                    ctx,
+                    bytes,
+                    match variant {
```
nabetti1720, line 838: It might be worth considering exposing these values (like `oid` and length) as an `impl` for `VariantEnum`.

modules/llrt_crypto/src/subtle/key_algorithm.rs
```diff
@@ -661,6 +733,390 @@ fn from_hmac<'js>(
...
+    ) -> Result<Option<KeyKind>> {
+        use crate::provider::modern;
+
+        let KeyAlgorithmMode::Import { format, kind, data } = mode else {
+            return Ok(None);
+        };
+
+        match format {
+            KeyFormatData::RawPublic(bytes) => {
+                *data = modern::import_ml_dsa_public_key(variant, bytes.as_bytes(ctx)?, false)
```
nabetti1720, line 819: Could you please let me know if there was a specific reason for not exposing these as part of the `CryptoProvider` functionality? (I am not asking you to do so; I just want to understand the intention behind the design.)

nabetti1720 (commented):
I have a few more concerns.

modules/llrt_crypto/src/subtle/key_algorithm.rs
panva: These algorithms only have one RustCrypto implementation regardless of the selected provider. Adding them to CryptoProvider would just mean adding the same forwarding methods to every provider.

Not a strong reason. Open to change.

nabetti1720: It seems that `glaviora` is starting to support modern algorithms, but it doesn't appear to be as broad as the scope of this PR.
https://github.com/ctz/graviola/#algorithms

Therefore, I don't think we need to rush to integrate it into `CryptoProvider` and make it swappable in this PR.
I'm sorry for the confusion.

modules/llrt_crypto/src/subtle/key_algorithm.rs
```diff
@@ -661,6 +733,390 @@ fn from_hmac<'js>(
...
+        if data.len() != 32 {
+            return Err(DOMException::data_error(
+                ctx,
+                "ChaCha20-Poly1305 keys must be 256 bits",
+            ));
+        }
+        Ok(Some(*kind))
+    }
+
+    let key_kind = import(ctx, mode, algorithm_name)?;
```
richarddd: Not a big fan of inline fn like this, can we move otu?

richarddd (changes requested):

modules/llrt_crypto/src/subtle/key_algorithm.rs
panva: 8de06b9c

richarddd (approved):

PR #1759, opened 2026-09-05
chore: Remove aws-sdk sample and clean up dependencies

### LINKED ISSUE #1748: Possible fix(deps): uuid 8.0.0 → 11.1.1, 12.0.1, 13.0.1 (CVE-2026-41907) in yarn.lock
I think there may be a problem in `yarn.lock` around line 3578.

Vulnerability: CVE-2026-41907 (MEDIUM) in `uuid@8.0.0`, pinned in yarn.lock (lines 3578-3581). The v3, v5, and v6 UUID generators accept caller-supplied output buffers (`buf`/`offset` options) but, in 8.x, do not validate buffer size or offset bounds. When a buffer is too small or the offset is out of range, the function performs a silent partial write into caller-owned memory instead of throwing, which can corrupt adjacent data and cause unpredictable behavior or data-integrity failures. Impact: exploitability is limited to code paths that explicitly pass a buffer/offset to uuid.v3(), uuid.v5(), or uuid.v6(); the common usage (no buffer, returns a string) is unaffected. However, because the failure is silent, any affected path should be treated as a correctness/integrity risk. Risk level: MEDIUM (context-dependent; HIGH only if untrusted input controls buffer sizing). Fix: upgrade to a patched release (>=11.1.1, >=12.0.1, >=13.0.1, or 14.0.0). Note when upgrading from 8.x: use named imports (no default export since v10), v3/v5 API uses an options object, and patched versions now throw a RangeError on out-of-range writes — update any code that relied on the old silent behavior.

Something like this might fix it:

````diff
```diff
--- a/package.json
+++ b/package.json
@@
   "dependencies": {
-    "uuid": "^8.0.0"
+    "uuid": "^13.0.1"
   }
--- a/yarn.lock
+++ b/yarn.lock
@@ -3578,7 +3578,7 @@
-uuid@^8.0.0:
-  version "8.0.0"
-  resolved "https://registry.yarnpkg.com/uuid/-/uuid-8.0.0.tgz#<old-hash>"
-  integrity sha512-<old-integrity>
+uuid@^13.0.1:
+  version "13.0.1"
+  resolved "https://registry.yarnpkg.com/uuid/-/uuid-13.0.1.tgz#<new-hash>"
+  integrity sha512-<new-integrity>
```

Recommended steps:
1. Regenerate the lockfile rather than hand-editing: `yarn upgrade uuid@^13.0.1` (or install 14.0.0 for the latest major).
2. Update imports if needed: `const { v3, v5, v6 } = require('uuid');` (named imports; default export was removed after v9).
3. Audit call sites that pass `buf`/`offset` to v3/v5/v6 and wrap them in try/catch, since patched versions now throw `RangeError` on out-of-range writes instead of writing silently:

```js
const bytes = new Uint8Array(16);
try {
  v5('example', v5.DNS, bytes, 0); // now throws RangeError if buf too small or offset invalid
} catch (e) {
  // handle invalid buffer/offset explicitly
}
```
````

For reference: rule `CVE-2026-41907`. Rated high.

I do not maintain this project, so I may well be missing context — if this is intentional or already handled elsewhere, please just close it.

---
*Found with automated scanning ([RedGem](https://code.redgem.net)) and reviewed before opening. If it is not useful, closing it is completely fine.*

nabetti1720:
### Issue # (if available)

Closed #1748 

### Description of changes

Keep dependencies clean by removing outdated aws-sdk v2 samples.

### Checklist


richarddd (approved):
Might as well, this is is bad practise anyway

PR #1771, opened 2026-09-10
chore: reduce release binary size (~10%) with zero perf/functional regressions

richarddd:
### Issue # (if available)

N/A

### Description of changes

Cuts release binary size ~10% (11.5MB -> 10.4MB on arm64 musl): a few extra codegen/linker flags for the musl targets, makes Intl/Temporal (`intl402`) opt-in instead of default, and de-genericizes the stream promise helpers so they stop getting duplicated per call site. No perf or test regressions.

### Checklist


Makefile
nabetti1720, line 1: You can run all implemented tests by adding `--features pqc` when executing `make test` and `make test-wpt`. This is a rational approach because it allows for a clear distinction between the functionality being tested and the functionality included in the provided artifacts.

panva:
Is cutting post-quantum algorithms out of default builds the right move? Wouldn't opt-outs be more appropriate for those chasing smaller footprints over compatibility? i.e. opt out of entire SubtleCrypto instead.

nabetti1720:
> Is cutting post-quantum algorithms out of default builds the right move? Wouldn't opt-outs be more appropriate for those chasing smaller footprints over compatibility? i.e. opt out of entire SubtleCrypto instead.

I agree with this view as well. Furthermore, given that SubtleCrypto is already widely known and relied upon by numerous downstream consumers, I do not think it is particularly worthwhile to prepare an opt-out feature for it.

As another proposal that would contribute somewhat to reducing the footprint, how about placing `intl` and `temporal` behind an `intl402` feature and making them opt-in?

Support for `Intl` and `Temporal` is not yet sufficient, and there are concerns that future extensions will lead to a further increase in the footprint. We are making incremental improvements to `Temporal`, but we want to avoid increasing the size of the artifacts officially provided by LLRT as a result.

richarddd:
Thanks both for the feedback.

On pqc: keeping it opt-in as-is. SubtleCrypto as a whole is too broad/foundational to gate behind a flag over a chunk that's still tentative/spec-draft territory — the size cost here is specifically the ML-DSA/ML-KEM parameter-set monomorphization, not crypto in general.

@nabetti1720 good call on intl402 — added it. Intl/Temporal are now opt-in the same way pqc is (off by default, ~550 KiB smaller, `--features intl402` brings it back).

richarddd:
Reconsidered the pqc point, you're both right. Reverted: pqc is back to default-on, the flag itself just stays around so it's still excludable explicitly if someone wants to. intl402 stays opt-in per the earlier suggestion.

panva:
ML-KEM, ML-DSA, and the Hybrids all go into Chrome 155 unflagged. I don't consider them tentative.

PR #1776, opened 2026-09-13
feat(buffer,fetch): Implement Blob, Request, and Response textStream()

nabetti1720:
### Issue # (if available)

n/a

### Description of changes

Implement the `textStream()` API for `Blob`, `Request`, and `Response`.
- Add native Rust implementation of `Blob.prototype.textStream()`
- Add `Request.prototype.textStream()`
- Add `Response.prototype.textStream()`
- Decode body data as UTF-8 and return `ReadableStream<string>`
- Handle empty bodies by returning an empty stream
- Throw when the body has already been consumed or locked

### Checklist


modules/llrt_fetch/src/body_helpers.rs
```diff
@@ -5,15 +5,66 @@ use std::{borrow::Cow, sync::RwLock};
...
+        .map(|body| Class::<ReadableStream>::from_value(&body))
+        .transpose()?;
+    let pull = PullAlgorithm::from_fn_once(move |ctx, controller| {
+        let source = source.clone();
+        let future_ctx = ctx.clone();
+        Promise::wrap_future(&ctx, async move {
+            let text = if let Some(source) = source {
+                let bytes = collect_readable_stream(&source).await?;
+                let stripped = strip_bom(bytes);
+                Some(String::from_utf8_lossy(&stripped).into_owned())
```
richarddd: can we avoid copy here if there is no bom?

richarddd (changes requested):

modules/llrt_fetch/src/body_helpers.rs
nabetti1720: Corrections complete.

nabetti1720:
@richarddd I have addressed the review comments. A different CI error has appeared, but I believe re-running the job will resolve it.

richarddd:
Rerunning!

modules/llrt_fetch/src/body_helpers.rs
```diff
@@ -163,6 +214,16 @@ pub(crate) fn strip_bom<'a>(bytes: impl Into<Cow<'a, [u8]>>) -> Cow<'a, [u8]> {
...
 
+pub fn bytes_to_utf8_string_lossy(bytes: Cow<'_, [u8]>) -> String {
+    match bytes {
+        Cow::Owned(vec) => match String::from_utf8(vec) {
+            Ok(s) => s,
+            Err(e) => String::from_utf8_lossy(e.as_bytes()).into_owned(),
+        },
+        Cow::Borrowed(b) => String::from_utf8_lossy(b).into_owned(),
+    }
+}
```
richarddd, line 230: I think we should do:

```rust
use simdutf8::basic::from_utf8;

pub fn bytes_to_utf8_simd_lossy(bytes: Cow<'_, [u8]>) -> String {
    match bytes {
        Cow::Owned(vec) => {
            if from_utf8(&vec).is_ok() {
                // SAFTEY: valid UFT8
                unsafe { String::from_utf8_unchecked(vec) }
            } else {
                String::from_utf8_lossy(&vec).into_owned()
            }
        }
        Cow::Borrowed(b) => String::from_utf8_lossy(b).into_owned(),
    }
}
```

richarddd (changes requested):
Smal perf tweak

nabetti1720:
The corrections have been completed.

richarddd (approved):

PR #1787, opened 2026-09-18
fix(wpt): improve harness compatibility

nabetti1720:
### Issue # (if available)

n/a

### Description of changes
Improve WPT harness compatibility with LLRT's server-side runtime.

- Provide browser-like `self`, `console`, and `DOMException` globals.
- Add runtime global fallback for WPT contexts.
- Support `self.location` and relative URL resolution in Fetch tests.

### Checklist


richarddd:
@nabetti1720 this is a bit out of scope for this PR but the hole point of wpt-errors is to keep track of what's supported and whats not. Now that we discard unsupported behaviour via filters in wpt-harness we no longer see that. I suggest we remove the filter and just maintain the `wpt-errors.txt` file.

nabetti1720:
> @nabetti1720 this is a bit out of scope for this PR but the hole point of wpt-errors is to keep track of what's supported and whats not. Now that we discard unsupported behaviour via filters in wpt-harness we no longer see that. I suggest we remove the filter and just maintain the `wpt-errors.txt` file.

I excluded that test, deeming it meaningless in a server-side runtime environment. However, there is value in keeping it as an error precisely to make that point clear.

The remaining fix concerns the harness mismatch. Resolving this cleared up the strange errors and revealed new errors specific to the test cases themselves.

richarddd (approved):

PR #1788, opened 2026-09-18
fix(encoding,util): Fix encoding streams behavior and share WTF-8 encoding logic

nabetti1720:
### Issue # (if available)

n/a

#### Summary

- Fix encoding streams WPT failures caused by incorrect input conversion and unresolved stream errors
- Share WTF-8 and surrogate-pair handling between `TextEncoder` and `TextEncoderStream`
- Add the encoding streams WPT suite configuration

#### Changes

- Accept JavaScript boolean coercion for `TextDecoder` options
- Reject empty encoding labels instead of silently defaulting to UTF-8
- Validate `TextDecoderStream` chunks as BufferSource values
- Propagate transform rejections to both readable and writable sides
- Handle lone and split UTF-16 surrogates in `TextEncoderStream`
- Extract common WTF-8 conversion helpers into `llrt_utils::bytes`
- Add exclusions for unsupported encoding test prerequisites

### Checklist

- [x] Created **wpt** tests in `tests/wpt` and/or in Rust for my feature if needed


libs/llrt_utils/src/bytes.rs
```diff
@@ -673,6 +667,85 @@ pub fn get_string_bytes(
     Ok(None)
 }
 
+pub fn get_wtf8_string_bytes(value: Value<'_>) -> Result<Vec<u8>> {
```
richarddd, line 675: Rework this function. Error is verbose comment to long :)

richarddd: I think we already have a similar function

modules/llrt_util/src/text_encoder_stream.rs
```diff
@@ -49,8 +60,43 @@ impl<'js> TextEncoderStream<'js> {
...
+fn transform_chunk<'js>(
+    ctx: Ctx<'js>,
+    pending: &Mutex<Option<u16>>,
+    chunk: Value<'js>,
+    controller: Object<'js>,
+) -> Result<()> {
+    let string = if chunk.is_string() {
+        chunk
+    } else {
+        let string: Function = ctx.globals().get(PredefinedAtom::String)?;
```
richarddd: Primordials

modules/llrt_util/src/text_encoder_stream.rs
```diff
@@ -49,8 +60,43 @@ impl<'js> TextEncoderStream<'js> {
     }
 }
 
-fn transform<'js>(ctx: Ctx<'js>, chunk: Coerced<String>, controller: Object<'js>) -> Result<()> {
-    let bytes = TypedArray::new(ctx.clone(), chunk.0.as_bytes())?;
-    let enqueue: Function = controller.get("enqueue")?;
-    enqueue.call::<_, ()>((This(controller), bytes))
+fn transform_chunk<'js>(
+    ctx: Ctx<'js>,
+    pending: &Mutex<Option<u16>>,
```
richarddd: Atomics instead?

modules/llrt_util/src/text_encoder_stream.rs
```diff
@@ -49,8 +60,43 @@ impl<'js> TextEncoderStream<'js> {
...
+        let encoded = TypedArray::new(ctx.clone(), output)?;
+        let enqueue: Function = controller.get("enqueue")?;
+        enqueue.call::<_, ()>((This(controller), encoded))?;
+    }
+    Ok(())
+}
+
+fn flush_stream<'js>(
+    ctx: Ctx<'js>,
+    pending: &Mutex<Option<u16>>,
```
richarddd: Attomics?

richarddd (changes requested):
I think this can be cleaned up a bit. Check also so we're not duplicating string handling. We already have a lot of string handling else where.

libs/llrt_utils/src/bytes.rs
nabetti1720: `get_lossy_string` and `get_wtf8_string_bytes` have different semantics. The former replaces lone surrogates according to USVString, while the latter preserves WTF-8 bytes for `TextEncoder`. The shared Value to WTF-8 byte conversion is now implemented by `get_wtf8_string_bytes`, which is reused by `get_lossy_string`.

richarddd (approved):

